PR #120, opened 2024-04-30
Fix date timer in the top header

Chewbaka69:
Only line 130 and  536 are modified to update the date 
refer #3

### DIFF
--- a/resources/views/ingame/layouts/main.blade.php
+++ b/resources/views/ingame/layouts/main.blade.php
@@ -120,15 +120,16 @@
                     </li>
                     <li><a href="{{ route('options.index') }}" accesskey="">@lang('Options')</a></li>
                     <li><a href="#">@lang('Support')</a></li>
-                    <li><a href="{{ route('logout') }}" onclick="event.preventDefault();
+                    <li>
+                        <a href="{{ route('logout') }}" onclick="event.preventDefault();
                                                      document.getElementById('logout-form').submit();">@lang('Log out')</a>
 
-                            <form id="logout-form" action="{{ route('logout') }}" method="POST"
-                                style="display: none;">
-                                {{ csrf_field() }}
-                            </form>
-                        </li>
-                        <li class="OGameClock"></li>
+                        <form id="logout-form" action="{{ route('logout') }}" method="POST"
+                              style="display: none;">
+                            {{ csrf_field() }}
+                        </form>
+                    </li>
+                    <li class="OGameClock">{{ \Carbon\Carbon::now()->format('d.m.Y H:i:s') }}</li>
                     </ul>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


lanedirt:
Hi @Chewbaka69, thank you very much for investing your time to contribute. Much appreciated!!

I checked and tested your changes. See my comments inline.

I'm happy to merge this change if you can fix the date display issue. Thanks :)

resources/views/ingame/layouts/main.blade.php
```diff
@@ -123,16 +123,17 @@ class="overlay" data-overlay-title="My notes"
...
-                        </form>
-                    </li>
-                    <li class="OGameClock">16.12.2017 <span>12:18:12</span></li>
-                </ul>
+                            <form id="logout-form" action="{{ route('logout') }}" method="POST"
+                                style="display: none;">
+                                {{ csrf_field() }}
+                            </form>
+                        </li>
+                        <li class="OGameClock"></li>
```
lanedirt: This animation shows the top menu switching positions because of this empty initial date:
fde1748e-2301-4fd4-bbd4-a29c8500fec4

I think this could be fixed by adding the initial time to the `<li class="OGameClock"></li>` element by using `date()`.

resources/views/ingame/layouts/main.blade.php
```diff
@@ -1032,6 +1040,14 @@ function initChatAsync() {
...
                                     </a>
                                     <!--
+                                        @if (!empty($build_active) && $planet->getPlanetId() == $currentPlanet->getPlanetId())
+                                            <a class="constructionIcon tooltip js_hideTipOnMobile tpd-hideOnClickOutside"
+                                                data-link="{{ $urlToCurrentWithUpdatedParam }}"
+                                                href="{{ $urlToCurrentWithUpdatedParam }}" title=""
+                                                previewlistener="true">
+                                                <span class="icon12px icon_wrench"></span>
+                                            </a>
+                                        @endif
```
lanedirt: I see that you added a change for the constructionIcon but it's commented out now. In order to make this work properly I think the best way is to add a new method to the PlanetService class which checks if something is under construction for that planet. Then this new method can be called during the PlanetList render.

Chewbaka69:
Hey sorry,
But i did not see any comments

lanedirt (changes requested):

lanedirt:
Ah I see, my bad. Forgot to submit the comments. Now they should be visible. 😄

resources/views/ingame/layouts/main.blade.php
Chewbaka69: Yes i have remove it, because i work on it, maybe in an other PR :)

Chewbaka69:
I have add the server time in the header, to fix the header to move right to left with the empty clock

lanedirt (approved):
Looking good, thanks for the fix!

PR #140, opened 2024-05-05
Add plasma technology table

### LINKED ISSUE #127: Add plasma technology table to "?" popup
When clicking on the "?" button for Plasma Technology the popup should contain a table with dynamic level and bonus values.

TODO:
- Add plasma technology table to the "?" popup. 
- Code for this popup can be found in `/app/Http/Controllers/TechtreeController.php`. 
- Some game objects such as Metal Mine, Crystal Mine, Solar Plant and ship units (rapidfire) already have logic to show a table. This logic can be reused.

See this as an example:
Screenshot 2024-04-30 at 17 41 01

Also see #119 which is related to this behavior.

Danielopes7:
resolve #127.

I've essentially followed the established standard. Please let me know if there are any issues 👍

### DIFF
--- a/app/Http/Controllers/TechtreeController.php
+++ b/app/Http/Controllers/TechtreeController.php
@@ -311,9 +311,8 @@ class TechtreeController extends OGameController
         if ($object->type != 'research') {
             return view('empty');
         }
-        
-        $planet = $player->planets->current();
-        $current_level = $player->planets->current()->getObjectLevel($object->machine_name);
+
+        $current_level = $player->getResearchLevel($object->machine_name);
         $plasma_table = [];
         $min_level = (($current_level - 2) > 1) ? $current_level - 2 : 1;
         for ($i = $min_level; $i < $min_level + 15; $i++) {
@@ -327,7 +326,6 @@ class TechtreeController extends OGameController
         }
         return view('ingame.techtree.info.plasma')->with([
             'object' => $object,
-            'planet' => $planet,
             'plasma_table' => $plasma_table,
             'current_level' => $current_level,
         ]);
--- a/app/Http/Controllers/TechtreeController.php
+++ b/app/Http/Controllers/TechtreeController.php
@@ -45,6 +45,7 @@ class TechtreeController extends OGameController
                 'storage_table' => $this->getStorageTable($object, $player, $objects),
                 'rapidfire_table' => $this->getRapidfireTable($object, $objects),
                 'properties_table' => $this->getPropertiesTable($object, $player, $objects),
+                'plasma_table' => $this->getPlasmaTable($object, $player),
             ]);
         } elseif ($tab == 3) {
             return view('ingame.techtree.technology')->with([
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/TechtreeController.php
```diff
@@ -297,4 +298,38 @@ public function getPropertyTooltip(string $name, array $breakdown, int $value):
...
+     * @throws Exception
+     */
+    public function getPlasmaTable(GameObject $object, PlayerService $player): View
+    {
+        if ($object->type != 'research') {
+            return view('empty');
+        }
+        
+        $planet = $player->planets->current();
+        $current_level = $player->planets->current()->getObjectLevel($object->machine_name);
```
lanedirt: ⚠️ I tested the change on my local environment and I noticed that with a plasma technology level of "7" it still shows the level 1-15 table.

I think the issue is probably in this line. The method that is currently being used is to determine a object level that is attached to a planet. Because research level are global for the player across all planets, they are not attached to the planet but to the current user.

You should be able to use the `$player->getResearchLevel()` method here. Then I think it should work.

app/Http/Controllers/TechtreeController.php
```diff
@@ -297,4 +298,38 @@ public function getPropertyTooltip(string $name, array $breakdown, int $value):
...
+        $plasma_table = [];
+        $min_level = (($current_level - 2) > 1) ? $current_level - 2 : 1;
+        for ($i = $min_level; $i < $min_level + 15; $i++) {
+
+            $plasma_table[] = [
+                'level' => $i,
+                'metal_bonus' => $i,
+                'crystal_bonus' => $i * 0.66,
+                'deuterium_bonus' => $i * 0.33
```
lanedirt, line 324: 😻  Looks good! 

ℹ️ However for documentation purposes: later when implementing these bonuses into the actual resource production of the planet these values will probably be need to moved to a central location.

Currently objects like "metal mine" have the following formula defined in the BuildingObjects.php file to calculate the production value:
```
        $metalMine->production = new GameObjectProduction();
        $metalMine->production->metal = 'return (30 * $building_level * pow((1.1), $building_level)) * (0.1 * $building_percentage);';
        $metalMine->production->energy = 'return - (10 * $building_level * pow((1.1), $building_level)) * (0.1 * $building_percentage);';
```

These formulas are ultimately evaluated by the `PlanetService::getBuildingProduction()` method.

This production values system currently only works for buildings on a planet.  So we'll have to define something similarly for these tech bonuses so the bonuses on top of the regular production can be calculated.

If you want to feel free to take a look at this, but IMO it's not a blocker for this issue and I'm happy to merge as-is. We can add a new issue to the backlog for the future. 😄 .

lanedirt (changes requested):
Hi @Danielopes7,

Welcome to this repo and thank you very much for your contribution! 🚀 Greatly appreciated. 

Two comments:
- The automatic tests have just ran, they indicate a small newline indentation issue.
- I've also tested the feature functionally and I have some suggestions, I've placed the comments inline to the code. Could you take a look at those?

If you have any questions, feel free to let me know. Thanks again for your help!

app/Http/Controllers/TechtreeController.php
Danielopes7: I really didn't notice that detail, I sent a new commit to fix it.

app/Http/Controllers/TechtreeController.php
Danielopes7: I understood it slightly 😆, as I've never played this game, but I can get it later from the backlog when I know better 👍.

lanedirt:
Changes are looking good, awesome work. 👍 I have merged the PR. Thanks again for your contribution!

PR #178, opened 2024-05-19
best practices, condition and other stuff

Chewbaka69:
Hi,

I would to open a PR, to fix some little issue and linter.

- remove intval to use cast better solution since PHP8
- use === instead of ==, in PHP we need to use triple equal to compare type and value
- in_array function, the third parameter is to use a strict comparison
- replace ?string by string|null in function parameter, because it is more readable for other
- replace `protected` properties by `private` because we use protected when we want that a field is use on child by heritage
- replace `app()->getLocale()` by `App::getLocale()` because in other file, you use it like that, so to be more consistent
- replace `'OGame\Models\User'` by `User::class`
- remove array_merge when not needed
- replace `$model_name = 'OGame\Models\UnitQueue'; $model = new $model_name();` by `$model = new UnitQueue();` better for PhpStorm/VS Code 
- `for ($i = 1, $iMax = count($this->queue); $i < $iMax; $i++)` do not use `count($this->queue)` in the middle parameters because the count function is call in every loop, so not optimize

Some issue on my side to run your test and I don't know why.

Feel free to add comment, and to not agree with me, all comment are welcome :)

### DIFF
--- a/app/Factories/PlanetServiceFactory.php
+++ b/app/Factories/PlanetServiceFactory.php
@@ -46,9 +46,9 @@ class PlanetServiceFactory
      * it is advised to use makeForPlayer() method if playerService is already available.
      *
      * @param int $planetId
-     * @return PlanetService
+     * @return PlanetService|null
      */
-    public function make(int $planetId): PlanetService
+    public function make(int $planetId): PlanetService|null
     {
         if (!isset($this->instancesById[$planetId])) {
             try {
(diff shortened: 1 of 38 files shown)


lanedirt:
Hi @Chewbaka69,

Thanks for your suggestions and changes! I reviewed your suggestions and they all sound good to me. 👍 Makes the code cleaner and more readable.

Thanks for these fixes. Will merge the PR!

PR #185, opened 2024-05-21
Fix issue on building techinfo

Chewbaka69:
When the ressource factor is reduce or put to 0, the techinfo table of the building production is devide by the percentage

image

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -1156,9 +1156,12 @@ class PlanetService
      * @param string $machine_name
      *  The machine name of the building to calculate the production for.
      *
-     * @param int $building_level
+     * @param int|null $building_level
      *  Optional parameter to calculate the production for a specific level
      *  of a building. Defaults to the current level.
+     *
+     * @param bool $force_factor
+     * Optional parameter use to force/simulate the production at 100%
      *
      * @return Resources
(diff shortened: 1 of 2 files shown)


app/Services/PlanetService.php
```diff
@@ -1163,7 +1163,7 @@ private function updateResourceProductionStatsInner(Resources $production_total,
      * @return Resources
      * @throws Exception
      */
-    public function getBuildingProduction(string $machine_name, int $building_level = 0): Resources
+    public function getBuildingProduction(string $machine_name, int|null $building_level = null, bool $force_factor = false): Resources
     {
```
lanedirt, line 1170: Could you please update the phpdoc comment to include the new parameter.

lanedirt (changes requested):

lanedirt:
Nice find, thanks for the fix. I have one small remark regarding missing phpdoc argument, other than that looks good!

app/Services/PlanetService.php
Chewbaka69: Oh yes sure my bad

lanedirt: Super, thx!

lanedirt (approved):

PR #186, opened 2024-05-21
Implement the application tab in the ? pop-up

### LINKED ISSUE #151: Implement "?" popup Applications tab
Implement the logic so the applications tab for the technology popup works. 

See `resources/views/ingame/techtree/applications.blade.php`.

It should look like this:

Screenshot 2024-05-14 at 08 09 44

Screenshot 2024-05-14 at 08 11 27

Chewbaka69:
resolve #151

### DIFF
--- a/app/GameObjects/BuildingObjects.php
+++ b/app/GameObjects/BuildingObjects.php
@@ -120,7 +120,7 @@ class BuildingObjects
         $fusionReactor->price = new GameObjectPrice(900, 360, 180, 0, 1.8);
         $fusionReactor->requirements = [
             new GameObjectRequirement('deuterium_synthesizer', 5),
-            new GameObjectRequirement('research_lab', 5),
+            new GameObjectRequirement('energy_technology', 3),
         ];
         $fusionReactor->production = new GameObjectProduction();
         $fusionReactor->production->deuterium = 'return - (10 * $building_level * pow(1.1, $building_level));';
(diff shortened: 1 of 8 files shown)


lanedirt:
Thanks for the contribution!

I tested this branch and two things that stand out to me:

1) In the official game it shows a red border around a GameObject if the requirements are currently not met. However it doesn't seem to work correctly for me. Below an example, I cannot build Terraformer as shown by the facilities page, but it shows as a green border in the popup. I would expect the terraformer to be greyed out and have a red border.
Screenshot 2024-05-21 at 18 59 02

2) Shipyard "?" popup does not show interplanetary missiles as requirement. In the official game the shipyard includes the interplanetary missiles as a requirement. But it's missing from the list for me. I'm not sure if this is a bug in the code or if a requirement is missing somewhere. I already noticed I made a few mistakes with the initial entry of the game objects, so might be my mistake as well. 😸 
Screenshot 2024-05-21 at 19 05 29

Chewbaka69:
Oooh good catch, i fix the way to have the green border, but maybe i break something else, i will check again :)

Chewbaka69:
some defense and station had the wrong requirement so i fix it, and i rework the method to check the requirement.

lanedirt:
Nice, looks good. Just tested it and as far as I can see it works beautifully now. Awesome! 🏆

PR #193, opened 2024-05-23
Optimize the function to check if requirement is met

Chewbaka69:


lanedirt:
Nice! I forgot during review that this method existed. +1 for simplifying! 👍

Issue #141, opened 2024-05-05
Show correct planet image based on planet type


lanedirt:
**TODO: The correct image for the planet needs to be shown in the following locations:**
- [X] Galaxy page
- [x] Planet list in right sidebar
- [x] Planet abandon/rename popup (button on overview page)
- [x] Empire page (does not exist yet)
- [x] Note: PR #194 added more images which resulted in some duplicates. Check which version is the best to use and delete any duplicates.

---

A planet is always one of these types:

1. dry
2. desert
3. normal
4. jungle
5. water
6. ice
7. gas

See `PlanetService::getPlanetType()` for how the planet type is determined.

The [planet type] + [planet position in galaxy:system] makes up the unique "image". E.g. planet 1:1:15 with type "dry" would be "dry_15".

Images are stored in `/public/img/planets`. Note: not all images have been added to this folder yet. Any missing images will need to be retrieved and stored in the correct folder.

Chewbaka69:
matching system of ogame to display planet by type:
![planets](https://www.zupimages.net/up/19/37/3v9n.png)

PR #194, opened 2024-05-23
Update image with all planet type and laste ogame picture

Chewbaka69:


lanedirt:
Thanks for adding additional images! I do see that we have some duplicates now, e.g. in 

- `public/images/planets/small/desert_1.gif`
- `public/images/planets/small/desert_1.png`

But any duplicates can be deleted later when working on issue #141. Additionally because of extension changes we might need to update the referencing code. But again, that's something we can tackle later.

For now I prefer more vs less 😄 .

PR #210, opened 2024-05-27
Add minimum time constraint of 1 second to all object build time calculation

lanedirt:


Chewbaka69:
@lanedirt you fix the wrong part of the code because the issue is on this part of the code:
image

lanedirt:
@Chewbaka69 

You are correct, but with the change I made the $time_per_unit should now always be at least 1. So there cannot be a division by zero anymore with new queue items. 

Having said that, older queue items created before this fix might still throw the error in your screenshot though. But if you delete all old items in the "unit_queue" database table it should work.

Chewbaka69:
Locally i try to build a colony ship, i saw this error after rebase your last commit :/

lanedirt:
That is strange. What does it say as production time when you look in the shipyard? It should be a minimum of 1 second now.

Screenshot 2024-05-27 at 22 47 32

PR #219, opened 2024-06-04
Add colonisation

Chewbaka69:


app/Http/Controllers/GalaxyController.php
```diff
@@ -149,7 +155,12 @@ public function getGalaxyArray(int $galaxy, int $system, PlayerService $player,
                             'planetMovePossible' => true,
                             'moveAction' => 'prepareMove',
                             'title' => 'Relocate'
-                        ]
+                        ],
+                        $user_planet->getObjectAmount('colony_ship') > 0 ? [
+                            'missionType' => 7,
+                            'link' => "/fleet?galaxy={$galaxy}&system={$system}&position={$i}&type=1&mission=7",
+                            'description' => "Coloniser<br>Cette position est normalement occupée par des planètes équilibrées, avec assez de deutérium et suffisamment d`énergie solaire, qui offrent assez de place pour le développement.<br><div style='display: flex;align-items: center;'><img src='/cdn/img/galaxy/activity.gif' style=''/>Sans vaisseau de colonisation, aucune colonisation de planète n`est possible!</div>"
```
lanedirt: Can you add the english translation here as default? I just checked the tooltip in the official game and the descriptions seem to be dynamic based on the position of the target planet. Examples:

Screenshot 2024-06-04 at 22 04 39

Screenshot 2024-06-04 at 22 04 47

Screenshot 2024-06-04 at 22 04 53

resources/js/ingame/e7c74974620fa35b197315ebdbb8c2.js
```diff
@@ -49092,6 +49092,8 @@ function getEmptySlotActions(galaxyContentObject, systemData) {
   let coloniseMission = galaxyContentObject.availableMissions.find(availMission => availMission.missionType === 7);
   let colonisationLink = "";
 
+  console.log(systemData.canColonize, coloniseMission, coloniseMission.link);
```
lanedirt: Can you remove the console log on this line?

lanedirt (changes requested):

lanedirt:
Thanks for the PR, you're on fire 🔥 !

I just checked the code and have added two comments. Also I tested the change on my local environment and see that overall it works well. I can send a colony ship to an empty planet and it works just like the official game. 👍  However after testing I did find two issues/bugs. Can you take a look at these?

1) **When I have a planet with 0 colony ships and go to the galaxy page, I get an error:**
Screenshot 2024-06-04 at 22 07 18

---

2) **When I do have a colony ship on my planet the galaxy page works. However the tooltip shows a missing icon:**
Screenshot 2024-06-04 at 22 09 09

app/Http/Controllers/GalaxyController.php
Chewbaka69: Oh shit!
Yep i will add it, thanks :)

resources/js/ingame/e7c74974620fa35b197315ebdbb8c2.js
Chewbaka69: Oh yes good catch!

lanedirt:
Hi @Chewbaka69, thanks for the fixes.

I tested the PR again locally and have two comments:

1) After checking out this new version I get this error:
Screenshot 2024-06-28 at 09 33 52

I'm not really sure why this only seems to happen on my local environment while the Github tests do run successfully. Anyway, I guess the reason is because the lang() in the layout/main.blade.php calls a straight string translation like so:

`<span class="textlabel">@lang('Galaxy')</span>`.

And because in this PR you added a new `resources/lang/en/galaxy.php` file it probably tries to load this translation based on the now existing array file instead of doing a straight string to language translation by doing a lookup in the optional `resources/lang/<langcode>.json` file.

My request: could you rename the `resources/lang/en/galaxy.php` to `resources/lang/en/t_galaxy.php` and change the corresponding translations you added to also use the "_t" prefix as well? Adding this prefix prevents collisions between direct string translations and array style translations.

BTW: the reason I decided to use both translation methods in this project is to not have to define all simple 1 word english strings such as "Galaxy" explicitly in the translation files but instead rely on the optional JSON files for translation lookup.

2) The "no colony ship" error message seems to be reversed. When a planet has > 0 colony ship the error shows up, while it should be the other way around. I added a comment to the code.
Screenshot 2024-06-28 at 09 45 35

Can you take a look at these two comments? Thanks for your help!

app/Http/Controllers/GalaxyController.php
```diff
@@ -140,17 +147,30 @@ public function getGalaxyArray(int $galaxy, int $system, PlayerService $player,
...
+                       'moveAction' => 'prepareMove',
+                       'title' => 'Relocate'
+                   ]
+                ];
+
+                $missions_available[] = [
+                    'missionType' => 7,
+                    'link'        => $user_planet->getObjectAmount('colony_ship') > 0 ? "/fleet?galaxy={$galaxy}&system={$system}&position={$i}&type=1&mission=7" : '#',
+                    'description' => __('galaxy.mission.colonize.name')."<br>{$planet_description}<br><div>{$has_colonize_ship}</div>"
+                ];
```
lanedirt: Currently the error message shows up when the planet has > 0 colony ships. I think this should be reversed: when the planet has exactly 0 the message should show up, when it has more than 0 no message should be displayed.

Chewbaka69:
Thanks for your review and i understand better your choice that look good for me too, so i will fix it asap and push it ;)

lanedirt (approved):

lanedirt:
Looking good, works perfectly now. Will merge now. Thanks again!

PR #240, opened 2024-06-20
fix - max_build_amount return 0 when requirements_met is false

### LINKED ISSUE #125: Change max. build amount to 0 if requirements for ship/defense are not met
Currently the ship / defense popup calculates the max amount of units that can be built with the current resources on the planet.

As seen here on the right side of the screen, the max amount of units that can be built is 12:
Image

TODO:
- The max. build amount should show "0" if the technology requirements for this object are not met (i.e. the object cannot be built). Change the template file for this ajax popup: `/resources/views/ingame/ajax/object.blade.php`.

Danielopes7:
Fixes #125 

I had to reload the player data in `UnitQueueTest.php` because the getResearchLevel in ` PlayerService.php ` was returning 0. This occurred because $this->user_tech hadn't been loaded with the researches yet.

Please let me know if there are any issues 👍

### DIFF
--- a/tests/Feature/UnitQueueTest.php
+++ b/tests/Feature/UnitQueueTest.php
@@ -8,7 +8,6 @@ use Illuminate\Support\Carbon;
 use OGame\Models\Resources;
 use OGame\Services\SettingsService;
 use Tests\AccountTestCase;
-use OGame\Services\PlayerService;
 
 /**
  * Test that the unit queue works as expected.
@@ -32,10 +31,6 @@ class UnitQueueTest extends AccountTestCase
         $this->planetSetObjectLevel('research_lab', 1);
         $this->playerSetResearchLevel('energy_technology', 1);
         $this->playerSetResearchLevel('combustion_drive', 1);
-
-        //refresh player
-        $playerService = app()->make(PlayerService::class, ['player_id' => $this->currentUserId]);
-        $playerService->load($this->currentUserId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/Feature/UnitQueueTest.php
```diff
@@ -31,6 +32,10 @@ private function basicSetup(): void
         $this->planetSetObjectLevel('research_lab', 1);
         $this->playerSetResearchLevel('energy_technology', 1);
         $this->playerSetResearchLevel('combustion_drive', 1);
+
+        //refresh player
+        $playerService = app()->make(PlayerService::class, ['player_id' => $this->currentUserId]);
+        $playerService->load($this->currentUserId);
```
lanedirt: I tried running the tests without this player refresh on my local machine and they still all pass. Could you check again if the tests still run when you remove this code?

lanedirt (changes requested):
Hi @Danielopes7, thanks for the PR. I just tested it on my local machine and it works flawlessly. Good job!

I have one comment regarding the player refresh logic in the test. See below in the review. Could you take a look at this? Thanks for your contribution and help 🚀 !

tests/Feature/UnitQueueTest.php
Danielopes7: I did the same thing now and it works. I have no idea why it wasn't working before, perhaps the branch was outdate. So I removed this part.

lanedirt: 👍 Thanks for double-checking!

lanedirt:
Looking good, merged! Thanks for this PR, much appreciated!

Issue #291, opened 2024-08-19
[BUG] Hide defense properties that are not relevant in "?" popup


lanedirt:
**Describe the bug**
Certain properties do not apply to defense objects, these are: speed, cargo capacity and fuel usage.

**To Reproduce**
1. Go to Defense.
2. Click on rocket launcher then click on the "?" button.

**Expected behavior**
Expecting the above mentioned properties to not show up for defense objects.

**Screenshots**
<img width="812" alt="Screenshot 2024-08-19 at 13 38 39" src="https://github.com/user-attachments/assets/078dd91b-e429-4f50-8c66-84823bc2358c">

**Additional context**
The template for this popup can be found in `resources/views/ingame/techtree/info/properties.blade.php`. The additional logic can most likely be added, perhaps in combination with a change in the controller which is located in `app/Http/Controllers/TechtreeController.php`.

jackbayliss:
@lanedirt  This can now be closed as it's resolved. :)

PR #366, opened 2024-09-29
Hide specific defence properties

jackbayliss:
Handles #291  

Also includes tests to ensure properties are hidden / shown.

### DIFF
--- a/app/Http/Controllers/TechtreeController.php
+++ b/app/Http/Controllers/TechtreeController.php
@@ -270,7 +270,7 @@ class TechtreeController extends OGameController
         $properties_array['shield'] = $shield->totalValue;
         $properties_array['attack'] = $attack->totalValue;
 
-        if($object->type !== GameObjectType::Defense){
+        if ($object->type !== GameObjectType::Defense) {
             $properties_array['speed'] = $speed->totalValue;
             $properties_array['capacity'] = $capacity->totalValue;
             $properties_array['fuel'] = $fuel->totalValue;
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/TechtreeController.php
```diff
@@ -269,9 +269,12 @@ public function getPropertiesTable(GameObject $object, PlayerService $player, Ob
...
         $properties_array['attack'] = $attack->totalValue;
-        $properties_array['speed'] = $speed->totalValue;
-        $properties_array['capacity'] = $capacity->totalValue;
-        $properties_array['fuel'] = $fuel->totalValue;
+
+        if ($object->type !== GameObjectType::Defense) {
+            $properties_array['speed'] = $speed->totalValue;
+            $properties_array['capacity'] = $capacity->totalValue;
+            $properties_array['fuel'] = $fuel->totalValue;
+        }
```
lanedirt, line 277: 💡 Could you add this same check to the $tooltips_array lines below?

Because right now tooltips are still being generated for properties that are no longer being shown, resulting in unnecessary calls. Preventing these unnecessary calls should make the code (a tiny bit) more efficient.

Also feel free to refactor the code a bit if you see improvements in readability or otherwise.

lanedirt (changes requested):
Hi @jackbayliss,

Thanks for your contribution and welcome  to the OGameX project ! I just tested your change on my end and it's working great. Also thanks for adding the test, that's much appreciated!

I do have a tiny suggestion on the code change, see my comments in the PR below. Could you take a look at this?

Thanks again for taking the time to help out! 👍

app/Http/Controllers/TechtreeController.php
jackbayliss: I've now applied the same check :)

lanedirt (approved):
LGTM, thanks for the quick update! And thanks again for your contribution 🚀 !

Issue #208, opened 2024-05-26
[BUG] Energy requirement check does not work when upgrading Space Dock Building


alican22:
Normally, I need to have at least 1.6K Energy to upgrade Space Dock.
![Screenshot 2567-05-26 at 12 50 12](https://github.com/lanedirt/OGameX/assets/7634657/0c0dba61-5be4-41ef-a91b-506977addbf2)

But I can upgrade successfully with -59 Energy (minus)
![Screenshot 2567-05-26 at 12 50 25](https://github.com/lanedirt/OGameX/assets/7634657/31052e59-03a2-43be-8143-25ad4f84ad98)

lanedirt:
Hi @alican22,

Thanks for the bug report. I checked the current code and it seems we are indeed missing a check for the energy cost.

The existing checks look like this now:

app/Http/Controllers/Abstracts/AbstractBuildingsController line 92:
```php
  // Get object
  $object = $objects->getObjectByMachineName($object_machine_name);

  // Get current level of building
  $current_level = $this->planet->getObjectLevel($object_machine_name);

  // Check requirements of this building
  $requirements_met = $objects->objectRequirementsMet($object_machine_name, $this->planet, $player);

  // Check if the current planet has enough resources to build this building.
  $enough_resources = $this->planet->hasResources($objects->getObjectPrice($object_machine_name, $this->planet));
```

The planet->hasResources() method which checks if the current planet has enough resources seems to be missing an energy requirement check:
```php
    /**
     * Checks if this planet has equal or more than the requested resources.
     *
     * @param Resources $resources
     * Array with resources to check.
     *
     * @return bool
     */
    public function hasResources(Resources $resources): bool
    {
        if (!empty($resources->metal->get()) && $this->metal()->get() < $resources->metal->get()) {
            return false;
        }
        if (!empty($resources->crystal->get()) && $this->crystal()->get() < $resources->crystal->get()) {
            return false;
        }
        if (!empty($resources->deuterium->get()) && $this->deuterium()->get() < $resources->deuterium->get()) {
            return false;
        }

        return true;
    }
```

I'm not fully sure, but as far as I know the energy requirements works a little different than the normal resources. Normal resources (metal,crystal,deuterium) are deducted from the planet. But energy is more of a "check" if the total production energy production of the planet is at least X. So in your screenshot where it requires 1,6K energy it should not look at your current balance "-59" but rather it should look at the actual "current production" which is visible in the tooltip:

[...]

Issue #174, opened 2024-05-19
[BUG] Graviton technology shows as possible to research when not enough energy


lanedirt:
**Describe the bug**
The graviton technology does not show a "disabled" state on the research page while there is not enough energy available on the planet.

**To Reproduce**
1. Have research lab level 12.
2. Go to research page.
3. See that graviton technology looks like its possible to research, while the planet does not have 300,000 energy.

**Expected behavior**
When < 300k energy, the graviton technology should show as disabled with stripes through the image.

**Screenshots**
<img width="1004" alt="Screenshot 2024-05-19 at 11 48 05" src="https://github.com/lanedirt/OGameX/assets/6917405/b92afbcb-35b0-48a7-ba1f-c6977306e3b2">

lanedirt:
Fixed by #370.

PR #370, opened 2024-10-02
fix energy requirement

jackbayliss:
The Energy requirement for research now checks against energy production now, rather than the current balance.

Which, hopefully should resolve https://github.com/lanedirt/OGameX/issues/208 https://github.com/lanedirt/OGameX/issues/174

I don't / haven't ever played this game, so hopefully I've got the logic / fix right. 

If not let me know 👍🏻

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -473,6 +473,9 @@ class PlanetService
         if (!empty($resources->deuterium->get()) && $this->deuterium()->get() < $resources->deuterium->get()) {
             return false;
         }
+        if (!empty($resources->energy->get()) && $this->energyProduction()->getFormattedLong() < $resources->energy->get()) {
+            return false;
+        }
 
         return true;
     }
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -473,7 +473,7 @@ class PlanetService
         if (!empty($resources->deuterium->get()) && $this->deuterium()->get() < $resources->deuterium->get()) {
             return false;
         }
-        if (!empty($resources->energy->get()) && $this->energyProduction()->getFormattedLong() < $resources->energy->get()) {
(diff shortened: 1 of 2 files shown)


app/Services/PlanetService.php
```diff
@@ -473,6 +473,9 @@ public function hasResources(Resources $resources): bool
         if (!empty($resources->deuterium->get()) && $this->deuterium()->get() < $resources->deuterium->get()) {
             return false;
         }
+        if (!empty($resources->energy->get()) && $this->energyProduction()->getFormattedLong() < $resources->energy->get()) {
```
lanedirt: I'm not sure if this is a typo but I think the `getFormattedLong()` here should be `get()` instead.

Currently the formatted string (e.g. "5.33M") is being compared with the raw number (e.g. 5330000).

Other than this the change itself looks good to me. 👍

lanedirt (changes requested):
Hi @jackbayliss,

Thanks for the PR! I have one small comment, see below. Could you take a look at this? Thanks again!

resources/views/ingame/ajax/object.blade.php
```diff
@@ -96,7 +96,7 @@
                     @endif
                     @if (!empty($price->energy->get()))
                             <li class="resource energy icon sufficient tooltip js_hideTipOnMobile
-                        @if ($planet->energy()->get() < $price->energy->get())
+                        @if ($planet->energyProduction()->getFormattedLong() < $price->energy->get())
```
lanedirt: The comment above applies here as well.

app/Services/PlanetService.php
jackbayliss: Thanks! Will adjust, I copied the logic that was showing in the pop up. Leave it with me 👍

lanedirt:
I just tested it on my local environment and looks like its working properly now for both mentioned tickets. Awesome, thanks a lot! 🚀  

Merging!

Issue #346, opened 2024-09-13
Remove social links from outgame template


lanedirt:
The outgame template contains icons with links to social media (Google+, Facebook) that should be removed. The icons can stay but the links should removed or set to `#`.
<img width="524" alt="Screenshot 2024-09-13 at 21 17 51" src="https://github.com/user-attachments/assets/fe77609e-c0cf-48ce-b3bd-5d609bedbd70">

alican22:
#351 its a minor commit but needed :)

lanedirt:
Fixed by #351.

PR #371, opened 2024-10-02
Remove facebook / google links

jackbayliss:
https://github.com/lanedirt/OGameX/issues/346

Removes the links, think the previous commit got wiped by accident?


lanedirt:
I also think the previous commit got deleted by an accidental force push. Thanks for making the change again, looks good. Merging!

Issue #340, opened 2024-09-06
[BUG] When build queue item is finished building the screen is not immediately reloaded


lanedirt:
**Describe the bug**
When you add a building, research or unit to the build queue and after it is finished the screen takes a few seconds to reload.

**To Reproduce**
1. Go to buildings
2. Try to build anything
3. Wait for the timer in the build queue to get to 0 and show "done".
4. The page reloads but it takes a few seconds.

**Expected behavior**
After the timer reaches "done", it should reload the page immediately (within 1 second).

**Screenshots**
See a video here which shows the delay of approx 3 seconds before the page reloads:

![ScreenRecording2024-09-06at09 32 57-ezgif com-video-to-gif-converter](https://github.com/user-attachments/assets/2b312d3c-99f2-4edb-974f-a323684ca782)

--

The problem is probably somewhere in the javascript that gets called when the timer reaches zero.

The building queue and countdown javascript are defined in:
- `resources/views/ingame/shared/buildqueue/building-active.blade.php`

Example snippet:

```
   <script type="text/javascript">
        var cancelBuildListEntryUrl = '{{ route('resources.cancelbuildrequest') }}';
        var questionbuilding = 'Do\u0020you\u0020want\u0020to\u0020reduce\u0020the\u0020construction\u0020time\u0020of\u0020the\u0020current\u0020construction\u0020project\u0020by\u002050\u0025\u0020of\u0020the\u0020total\u0020construction\u0020time\u0020\u00287m\u002010s\u0029\u0020for\u0020\u003Cspan\u0020style\u003D\u0022font\u002Dweight\u003A\u0020bold\u003B\u0022\u003E750\u0020Dark\u0020Matter\u003C\/span\u003E\u003F';
        var pricebuilding = 750;
        var referrerPage = $.deparam.querystring().page;

        new CountdownTimer('buildingCountdown', {{ $build_active->time_countdown }}, '{{ url()->current() }}', null, true, 3)

        function cancelbuilding(id, listId, question) {
            errorBoxDecision('Caution', "" + question + "", 'yes', 'No', function () {
                buildListActionCancel(id, listId)
            });
        }
    </script>
```

PR #372, opened 2024-10-03
queue building instant reload

jackbayliss:
Should fix https://github.com/lanedirt/OGameX/issues/340

**Changes:**
- switch default to document.location.href - and added the others as fall backs, also added a console error log here to track it easier. (It was causing errors right away in Chrome, so not sure if this had some delay effect)

- remove if statement based on `timeLeftInSeconds` , as once it's 0 or below it should trigger - this seemed to delay the reload. Again, perhaps this was done for a particular reason? 

You can see a loom here that it now loads instantly : https://www.loom.com/share/3e90629b29f849fa8dba88ba7f35a2fd (obviously some delay due to local env)

### DIFF
--- a/public/js/ingame.js
+++ b/public/js/ingame.js
@@ -62101,13 +62101,11 @@ class SimpleCountdownTimer {
         this.countdownDoneFunction();
       }
 
-      if (timeLeftInSeconds <= -1 && timeLeftInSeconds > -12 && Math.abs(timeLeftInSeconds % 3) === 0 || timeLeftInSeconds <= -12 && timeLeftInSeconds > -180 && Math.abs(timeLeftInSeconds % 10) === 0) {
         if (this.reloadPage != null) {
           reload_page(this.reloadPage);
         }
 
         timerHandler.removeCallback(this.timer);
-      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


lanedirt:
Hi @jackbayliss,

Thanks for taking the time to look into this. 

I just tested this PR on my local machine but unfortunately I don't see the correct expected behavior (yet). For me after the build finishes and the timer goes from "1 seconds" to "done", it stays on the same page for like 2-3 more seconds before it actually starts reloading. I am expecting it to start reloading max. 1 second after the timer switches to "done".

I'm not sure if I'm testing it correctly but from what I could tell based on the git diff:

I notice you made a change to the `public/js/ingame.js` file directly. This file is actually the output of the the Laravel webmix compiler. The source files are located in the `resources/js/ingame` folder. Could you try and make the same change in the appropriate source file(s)? A quick search also shows me that there are more than 1 occurence of this code:

```
if (timeLeftInSeconds <= -1 && timeLeftInSeconds > -12 && ...[snipped]
```

So perhaps the removal of this code needs to happen in more than one place. I'm not sure if it was deliberate to only remove it from one location, but just sharing what I noticed.

After you make the change to the source file(s) you can run this command in the ogame-app docker container:

```
$ npm run dev watch
```

This will recompile the JS/CSS files.

If/when I can test again feel free to let me know. Thank you again for your time!

jackbayliss:
@lanedirt  Doh! No problem I'll give it a go.

jackbayliss:
@lanedirt  I've now changed the correct resources file, I've also included the public files in the commit as it seems they're also commited to the repo. 

Here's a loom of the page refreshing as soon as it hits done - https://www.loom.com/share/ddf0bbf47c114773bb2de13678809742 

You should be able to see it's reloading in the tab as soon as the timer is done, Is the same happening for you, but perhaps it's just taking awhile to load? 💭

lanedirt:
Hi @jackbayliss,

Just tested the new version and looks like its working great now! As soon as the timer hits "done" it immediately refreshes my screen. This will make the new player experience a lot smoother.

Thanks for your research and figuring it out. 💯 👍  Merging.

Issue #377, opened 2024-10-04
Add developer shortcut option to add resources to current planet


lanedirt:
## Request
It would be nice if we could add an option to the "Developer shortcuts" screen to add resources to the current planet for certain tests. For example in order to properly reproduce #376 you need a certain amount of resources on your planet.

Note: the "Developer shortcuts" page is only accessible when you are logged in as an admin (on a local OGameX docker instance).

## Todo

- [ ] Add option to "Developer shortcuts" to add resources to the current planet. An option to "set metal/crystal/deut to X amount" would be sufficient, but being able to add (append) and remove (deduct) a certain amount of resources would be a nice addition.

## Tips

The controller of this page is located in:
```
app/Http/Controllers/Admin/DeveloperShortcutsController.php
```

The associated view that this controller uses can be found in:
```
resources/views/ingame/admin/developershortcuts.blade.php
```

### Screenshots
Example screenshot of the current developer shortcut options:
<img width="716" alt="Screenshot 2024-10-04 at 09 59 30" src="https://github.com/user-attachments/assets/85f8ccc8-cca9-4b5e-aaad-3a9b0eae7801">

lanedirt:
Implemented in #380.

PR #380, opened 2024-10-06
adds resource adjust ability on dev page

jackbayliss:
Handles https://github.com/lanedirt/OGameX/issues/377

Image below : 
image

I've added a new `ResourceType` enum to handle the different types of resources, rather than just using an array of strings etc as made sense to have some kind of strict type. This may already be somewhere so let me know If I missed it!

I've added the logic into a separate form, route and controller logic rather than dumping it into the else statement, as it's then separate and more performant than it would be when doing unit additions.  (ie [here](https://github.com/lanedirt/OGameX/blob/35c42e3b832f02f649d17d151adbfec9a6a506ae/app/Http/Controllers/Admin/DeveloperShortcutsController.php#L85) )

Also added a new `AddResource` method to the planet service, rather than passing the resources object and changing which param to pass it, again this uses the enum too.

Included tests also :)  

Let me know if you'd like any adjustments,

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -8,6 +8,7 @@ use Illuminate\Support\Facades\DB;
 use OGame\Factories\PlayerServiceFactory;
 use OGame\GameObjects\Models\Enums\GameObjectType;
 use OGame\GameObjects\Models\Units\UnitCollection;
+use OGame\Models\Enums\ResourceType;
 use OGame\Models\FleetMission;
 use OGame\Models\Planet;
 use OGame\Models\Planet\Coordinate;
@@ -988,6 +989,27 @@ class PlanetService
         if ($save_planet) {
             $this->save();
         }
+    }
+
+    /**
+     * @param string $resourceName
+     * @param int|float $amount
+     * @param bool $save_planet
+     * @return void
+     * @throws Exception
+     */
+    public function addResource(string $resourceName, int|float $amount, bool $save_planet = true): void
+    {
+
+        if (in_array($resourceName, array_column(ResourceType::cases(), "value")) && isset($this->planet->{$resourceName})) {
+            $this->planet->{$resourceName} += $amount;
+            if ($save_planet) {
+                $this->save();
+            }
+        } else {
+            throw new Exception('Invalid Resource');
+        }
+
     }
 
     /**
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -992,7 +992,7 @@ class PlanetService
     }
 
     /**
-     * @param string $resourceName
+     * @param ResourceType $resource
      * @param int|float $amount
      * @param bool $save_planet
      * @return void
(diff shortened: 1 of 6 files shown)


jackbayliss:
Not sure why the docker test failed, but hoping it's resolved with a rerun @lanedirt

app/Services/PlanetService.php
```diff
@@ -990,6 +991,27 @@ public function addResources(Resources $resources, bool $save_planet = true): vo
...
+            $this->planet->{$resourceName} += $amount;
+            if ($save_planet) {
+                $this->save();
+            }
+        } else {
+            throw new Exception('Invalid Resource');
+        }
+
+    }
+
```
lanedirt: Nice idea to add a separate method for the addResource(), will come in handy in other places as well  that are only adding or subtracting a single resource type.

I do would like to add a suggestion: could you change the `$resourceName` parameter in this method to be a `ResourceType` enum case instead of a string. This so we have static type safety for all places that call this method. I did see you added a test for the invalid resource name detection behaviour which is good too. But if we have static types then PhpStan will be able to detect any incorrect callers automatically during testing.

After this change you will probably have to change the DeveloperShortcutsController.php file too and do the string to enum mapping there before calling the `addResource` method. All in all I think this would make it more robust.

resources/views/ingame/admin/developershortcuts.blade.php
```diff
@@ -17,52 +17,73 @@
...
+                                    <div class="fieldwrapper">
+                                        <div class="smallFont">You can enter positive or negative values to add or subtract to the selected resource.</div>
+                                        <label class="styled textBeefy">Amount of resources to add / subtract :</label>
+                                        <div class="thefield">
+                                            <input type="text" pattern="^([-+,0-9.]+)" class="textInput w50 textCenter textBeefy" value="1" size="2" name="amount_of_resources">
+                                        </div>
+                                    </div>
+                                    <div class="fieldwrapper" style="margin-bottom: 50px;">
+                                        @foreach (\OGame\Models\Enums\ResourceType::cases() as $resource)
+                                            <input type="submit" name="resource_{{ $resource->value }}" class="btn_blue" value="{{$resource->value}}">
```
lanedirt: 💡 Small nitpick: could you make the first character of the label of the input button be uppercase? So to make it `Metal` instead of `metal`. I think this looks nicer and conforms to the rest of the game where resources are mentioned.

lanedirt (changes requested):
Hi @jackbayliss,

Thanks for your time in adding this feature, much appreciated! I tested the changes and its working very well my end. Handy feature to have!

I did add two comments while reviewing the code. Could you take a look at this?

@tests: I just reran the tests and they seem to be working fine now, so maybe its caused by some flakiness. If it happens more often I'll take a look at it. But I don't think it has anything to do with these changes so we can ignore it for now. 😄

jackbayliss:
@lanedirt  Thanks for the feedback, appreciated. Agreed the Enum type is much better to use in this case- saves having to check it's one of the expected Enums too, and reduces the tests.  👍🏻 🥇

lanedirt (approved):
Works and looks good to me! Thanks for adding this feature and applying the requested changes, much appreciated 🚀.

Issue #143, opened 2024-05-05
Add method to calculate current used fields (building levels) on planet


lanedirt:
For various features we will need a method that can calculate the amount of currently used fields on a planet. One example where this is shown is on the overview page as seen below.

I'm not sure yet how the amount of fields used should be calculated exactly. But this can probably be found out via wikis or basic reverse engineering.

TODO:
- [x] Add a method to `PlanetService` that calculates the amount of used fields on a planet and returns it as an int.
- [x] Use new method on overview page to show the current amount of fields used on the current planet.
- [x] Use new method on planet sidebar menu in planet tooltips.

![Image](https://github.com/lanedirt/OGameX/assets/6917405/2ade8014-7ee7-450c-a6b0-902598a439de)
![Image](https://github.com/lanedirt/OGameX/assets/6917405/c52343a1-13c8-493d-952f-ccd15882e6bb)

Chewbaka69:
In other Ogame Clone, when you build a building, it increase the a field (field_current)
but an other approach can be to count on the fly the number of building level.

If planet type === Planet : sum level of building by type building && can be construct on Planet
else if planet type === Moon : sum level of building by type building && can be construct on Moon

lanedirt:
> In other Ogame Clone, when you build a building, it increase the a field (field_current) but an other approach can be to count on the fly the number of building level.
> 
> If planet type === Planet : sum level of building by type building && can be construct on Planet else if planet type === Moon : sum level of building by type building && can be construct on Moon

I prefer the option to calculate this value on-the-fly. 👍  We could probably cache the value if performance becomes an issue (later). But being able to recalculate the value in case of issues with the initial calculation logic seems like the most optimal choice.

jackbayliss:
I believe this one can be closed :)

Issue #110, opened 2024-04-28
Implement building downgrade/tear down functionality


lanedirt:
Screenshot example:

<img width="674" alt="Screenshot 2024-10-11 at 10 31 24" src="https://github.com/user-attachments/assets/55f0b4c9-72c9-4afc-bf3b-f5046e74f5dd">

PR #390, opened 2024-10-10
Calculate planet used buildings and max buildings

jackbayliss:
This I believe is the beginning of https://github.com/lanedirt/OGameX/issues/143 so I've marked this as a draft for more detail.

I'm not sure where the /188 comes from, does this also need to be generated?    Maybe something you can answer @lanedirt - I'm assuming its the max level each building count get to.

This adds a new method to get the buildings associated with the planet, which essentially is just a count on the queue items to see how many items have been built for that planet.. as each built item is just a level, so the count should match. 

Feel free to take over and edit 😄 

List of things to do:
- [ ] Figure out where the max int comes from
- [ ] Add new max int into the overview 
- [ ] Add the same overview logic to the planet sidebar menu


lanedirt:
Hi @jackbayliss,

Thanks for starting work on this feature! I notice I marked issue #143 as a `Good first issue` when I created it, but now I'm starting to doubt if this feature is really as easy as I thought it was. 😃 It might involve changing more things than I originally thought it would. But anyway, I'll try to help by answering your questions and give you better context as much as I can.

I checked the WIP code and your questions. I'll try to give some more context based  on what I know how the official game works.

---

1. The `/188` max fields should be equal to the `fields_max` value that already exists in the `planets` table. Currently the `/188` shown in the template seems to be static so this static value should be replaced with the actual value from the database.
 
2. Additionally: this `field_max` value is currently randomly generated by the `PlanetServiceFactory::createPlanet()` method which is executed when a new planet is created (executed upon player registration or colonisation). Currently this random assignment looks like this:
```
        $planet->field_max = rand(140, 250);
```

This will probably need to be updated to reflect how the official game works. AFAIK the planet size is random for every planet, but the position of the planet in the system (e.g. for [1:2:3] i mean the `3` part) does play an important part for how small or big a planet can be. More information can be found here: https://ogame.fandom.com/wiki/Planet_Size.

3. I checked your WIP code and I noticed that currently you are attempting to calculate the amount of used fields based on the `BuildingQueue` table contents (in `PlanetService::getBuildingCount()`). Based on the features currently implemented in the game I think this could work, but from the top of my head there are some additional game logic features that can causes issues with this approach. 

Some examples:

- Not all buildings take up a planet field. E.g. the building `Space Dock` has this description: `Since the Space Dock floats in orbit, it does not require a planet field.`. I'm not sure if there are more buildings like this, but because of this we cannot make a direct comparison based on how many building (levels) are built and how much planet fields are taken up.
- Buildings can be demolished/downgraded. This is not implemented yet, but for an example screenshot see issue #110. So if someone has decided to upgrade their solar plant to level 20, then afterwards they can downgrade it back to level 10 if they want.

In order to make it work for all usecases I think the method for calculating the amount of used fields might need to take a different approach. 

My own idea was to loop through all of the building objects via `BuildingObjects::get()` and `StationObjects::get()` and then use the `machine_name` property to look up the actual level in the database for each building on the planet and finally sum up the levels. I think this would be the most simple way to do this and since the `planet` DB model is already loaded in the `PlanetService` this does not require any additional database calls.

These are just my 2 cents, if you can come up with a different idea that you believe can work for the usecases described above feel free to give that a go as well. 👍  Thanks for your effort in looking into this 🚀 !

jackbayliss:
Awesome 👍 thanks for the reply, I'll have a think and see what I can do over the weekend. 

I haven't ever played the game so I'm kind of picking up bits and pieces - thanks for explaining!  thx

jackbayliss:
@lanedirt  This is now ready for review, and does the following: 
-  adds `getPlanetFieldMax` to planet service, so we can easily get the field_max value.
- Adds  `getBuildingCount` to the planet service, that sums the levels of the Building and Station Objects on the planet
- Added building max / building count to overview and planet tool tips
- Adds tests

lanedirt:
Hi @jackbayliss,

I just tested the changes, and everything looks and works great! Excellent job finding and reusing the existing `PlanetService::getBuildingArray()` method—I had forgotten we already had that. Also, thanks for taking the time to add the tests.

While testing the planet field size game logic, I noticed additional game-related logic improvements that can be worked on now, thanks to the foundation laid in this PR. For more information see:

- #391
- #392
- #393

Merging this PR now, thanks again! 🚀

Issue #126, opened 2024-04-30
Show actual unit/planet counts in galaxy overview page


lanedirt:
Currently the galaxy page shows incorrect amounts for these areas:

Top:
- "Esp. Probe"
- "Recycler"
- "IPM"

Bottom:
- "X Planets Colonized"

TODO:
- Update the parts above so it reflects the actual amount of units on the current planet. The `PlanetService::getObjectAmount()` method can most likely be used for this.

See this example screenshot:

![Image](https://github.com/lanedirt/OGameX/assets/6917405/b81dccec-f719-488b-acbe-1a270b39afa0)

gabrieltsants:
Hello!

I noticed that the "Esp. Probe" object seems to be correct. As for the "IPM", I fixed it by changing it to 'availableMissiles' => $planet->getObjectAmount('interplanetary_missile'), and I applied a similar fix to the recycler. Does this look correct, or did I miss something?

lanedirt:
Hi @gabrieltsants,

Thanks for looking into this! The fix you're describing seems to be correct to me. If you could make a PR for it I'll gladly take a look and test it. 😄

lanedirt:
Fixed by PR #397.

PR #397, opened 2024-10-16
fix units count #126

gabrieltsants:


app/Http/Controllers/GalaxyController.php
```diff
@@ -399,7 +399,7 @@ public function ajax(Request $request, PlayerService $player, PlanetServiceFacto
                 'playerId' => $player->getId(),
                 'settingsProbeCount' => 3,
                 'showOutlawWarning' => true,
-                'slotsColonized' => 3,
+                'slotsColonized' => $player->planets->count(),
```
lanedirt: The `slotsColonized` should reflect the amount of planets that are colonized in the current system. The value you are using now is the amount of planets the current player has. This results in an incorrect count. 

See e.g. this screenshot. I have two planets, but the system I'm looking at has three planets colonized. So I expect it to show `3 Planets colonized` instead of the two that I see now.

Screenshot 2024-10-16 at 15 15 26

I think the output of the existing method `GalaxyController::getGalaxyArray()` could be used to check how many planets in the system are actually colonized.

lanedirt (changes requested):
Hi @gabrieltsants,

Thanks for taking the time to look into fixing this issue! I tested the changes and the recycler / IPM counts seem to be working good.

I did however notice that the amount of planets colonized does not seem to match what is shown on screen. Check my comment below.

Could you take a look at this and see if this issue can be fixed? If you have any further questions feel free to let me know.

app/Http/Controllers/GalaxyController.php
gabrieltsants: Hi @lanedirt,

I apologize for the misunderstanding regarding the count of colonized planets. I initially thought the count referred only to the player's planets. Now that you've clarified, I see that we should only count the colonized planets within the current system.

Here is the logic I've implemented to count the colonized planets:

```PHP
public function ajax(Request $request, PlayerService $player, PlanetServiceFactory $planetServiceFactory): JsonResponse
{
   // ...
    $galaxyContent = $this->getGalaxyArray($galaxy, $system, $player, $planetServiceFactory);
    $slotsColonized = 0;

    foreach ($galaxyContent as $galaxySystem) {
        if (isset($galaxySystem['planets'])) {
            $activePlanets = array_filter($galaxySystem['planets'], function ($planet) {
                return !isset($planet['isDestroyed']) || !$planet['isDestroyed'];
            });
            $slotsColonized += count($activePlanets);                
        }
    }
   // ...
}
```
I believe this implementation accurately counts the colonized planets in the current system. If you have any further thoughts on this or if you think the logic should be placed elsewhere, such as a different method or a specific pattern followed in the project, I’m open to any suggestions for improvement.

Thank you very much for your attention and time.

lanedirt: Hi @gabrieltsants,

Thanks for tweaking the code. However, it's a bit difficult for me to review the changes based only on the snippet you provided. Could you push the changes in a new commit and add it to this PR so I can test it locally?

I'm happy to test the new version and provide feedback.

gabrieltsants: Hi @lanedirt,

I’ve just committed the changes with the refactor you requested. You can now review the updates directly in the PR. Let me know if anything else is needed or if you encounter any issues while testing!

Looking forward to your feedback.

app/Http/Controllers/GalaxyController.php
```diff
@@ -438,4 +440,26 @@ private function getPlanetActivityStatus(PlanetService $planet): array
...
+    {
+        $slotsColonized = 0;
+
+        foreach ($galaxyContent as $record) {
+            if (isset($record['planets']) && !empty($record['planets'])) {
+                $activePlanets = array_filter($record['planets'], function ($planet) {
+                    return (!isset($planet['isDestroyed']) || !$planet['isDestroyed']) && isset($planet['playerId']);
+                });
+                $slotsColonized += count($activePlanets);
+            }
```
lanedirt: Hi @gabrieltsants,

I just tested the new changes and functionality wise its looking good now. I'm getting accurate counts with this version so that seems to be working well 👍.

For the changes I have a small tip, and that is to use the `!empty()` check instead of the `isset()` as the former already checks if the key in the array is set or not. So in this case the `isset()` is not necessary and can be removed. This will make the code a little bit cleaner and easier to read.

Concrete if you could make the code like this I think it should still work and is more concise:

```
            if (!empty($record['planets'])) {
                $activePlanets = array_filter($record['planets'], function ($planet) {
                    return empty($planet['isDestroyed']) && !empty($planet['playerId']);
                });
                $slotsColonized += count($activePlanets);
            }
```

I have not tested it yet but think it would work. Can you see if you can refactor this method?

Again functionality wise it looks good to me so with this change I'll be happy to merge it. Thanks again for your time! 🚀

lanedirt (changes requested):

lanedirt (approved):
Looks good!

lanedirt:
Thanks for making the adjustments @gabrieltsants! Works and looks good to me! Merging into the main codebase now.

app/Http/Controllers/GalaxyController.php
gabrieltsants: Hi @lanedirt,

I have made the changes as per your suggestion and pushed the updated code. Please feel free to test it again, and let me know if there’s anything else needed.

Thank you for your guidance, and I remain at your disposal for any further adjustments 😁

PR #420, opened 2024-10-26
Fix for Recruit Officers Frontend.

alican22:


lanedirt:
With this change the page seems to be rendering good now. Thanks for looking into this! Merging. 👍

Issue #299, opened 2024-08-21
Extend "can building be built" requirements logic with build queue contents


lanedirt:
Currenly when trying to build a building the requirements are checked. If the requirements are not fulfilled you cannot add a building to the build queue. However when building multiple buildings via the queue the contents of the queue should be included in this check.

Example: shipyard building requires robotics factory level 2. If you have a new planet and add robotics factory level 1 and level 2 to the queue, then after those are built the shipyard should also be able to be built. But right now when level 2 is in the queue the shipyard is not eligible yet to be built. According to the official game it should be possible to add shipyard to the build queue if level 2 of robotics factory is in the queue.

This also applies to other buildings and also to research.

Todo:

- [x] Extend "requirements met" check for buildings and research: look in the existing build queue and include the to-be-built/upgraded levels in the check.
- [x] Make sure to also add proper checks when a build queue item starts in case previous queue items have been canceled. E.g. it should not be possible to add robotics factory level 1 and 2 to the queue, then add shipyard build order, cancel the robotics factory and still end up with a shipyard. Add double checks: cancellation of item should recalculate the queue and remove any where requirements are no longer met. Additionally add check for when a build queue actually starts processing if requirements are (still) met.

lanedirt:
Fixed by #421.

Issue #181, opened 2024-05-20
[BUG] Disallow research when a research lab is upgrading on any planet


lanedirt:
**Describe the bug**
When the research lab is upgrading on any planet, new research of technology should not be possible anywhere.

**To Reproduce**
1. Start upgrading research lab on Planet A.
2. Click on 'research' in menu. You can still research new technologies. This works for both Planet A and Planet B.

**Expected behavior**
1. Start upgrading research lab on Planet A.
2. Click on 'research' in menu. All research should now be blocked with status "Research Lab is being expanded". (See screenshot below). This should happen on all planets.
3. Note: this should also apply in reverse. I.e. if research is being done, upgrading research lab should not be possible on any planet.

**Screenshots**

<img width="682" alt="Screenshot 2024-05-20 at 20 29 14" src="https://github.com/lanedirt/OGameX/assets/6917405/96883f9d-b04b-4e8b-b204-50e0a053a3ab">
<img width="685" alt="Screenshot 2024-05-20 at 20 31 07" src="https://github.com/lanedirt/OGameX/assets/6917405/a9ca65b9-1cc6-41bc-b748-fecef572a4a0">

lanedirt:
Closed by #443

PR #421, opened 2024-10-26
Feature - extend building queue requirements

rautamik:
Related to: https://github.com/lanedirt/OGameX/issues/299

Changes:
- Check object prior levels when object requirements are checked. Eg. check that robot_factory 1 exists when robot_factory 2 requirements are checked. Remove old check from build and research queue services.
- Accept build and research queue items when object requirements are checked. Eg. when shipyard requirements are checked accept required robot_factory 1 and 2 from build queue.
- Check that requirements are still met when item is taken from build or research queue for processing, if requirements are not met the item building is canceled.
- Check all items in the building and research queue when building or research are cancelled and remove all objects that are not passing requirements check,
- Add tests for building queue and research queue for checking if item is in queue and canceling item when requirements are not met.

Todo:

- [x] Fix tests broken after rebase
- [x] Fix starting research/building when requirements are in queue/progress and not yet ready


rautamik:
I noticed one issue with this PR. If research lab is started to be built, then it's possible to start research even the research lab is not ready. I think this is because in research queue start function requirements check building queue items are accepted when research requirements are checked. I guess that building/research queue items should not be accepted when starting to build new object, only existing items should be accepted. I will look into this.

rautamik:
Another thing is that I didn't run the unit tests after rebase and some cases seems to fail now. I will look into this.

lanedirt:
Hi @rautamik,

Thanks for your investigation and work on this so far, sounds promising! 

I'll wait until you give the green light that the changes are working stable, afterwards I'll gladly review the code and try testing it on my part with various use cases as well.

Thanks again for your contributions, greatly appreciated 🚀 !

rautamik:
Hi @lanedirt,

Is the goal to make crosschecks between building and research queues, for example accept queueing research if required level of research lab is in building queue? If the goal is to do crosschecks then there are following use cases: research lab level 1 is currently building and user is able to add computer technology to research queue as requirements are met, but the item is immediately canceled as the requirements are not ready when researching is started.

Any other use cases to cover than the basic use case (robotics factories + shipyard)?

lanedirt:
Hi @rautamik,

Thanks for your time and progress on this feature, it's looking great!

I thought a little bit about your question, and I think, based on my current understanding of the official game, that we do not need to do crosschecks between buildings and technology.

A concrete example from the top of my head:
- Technology should only be allowed to be queued if the research lab and required level are fully constructed at that moment. For example, if a technology needs a level 5 research lab, it should only be able to be queued when level 5 research lab is actually active at that exact moment. 

Another thing that might also be relevant to this feature:
- Once a research lab exists (level > 0), both lab upgrades and new research should be able to be added to the queue at the same time. However, active research and an active lab upgrade shouldn’t be able to happen at the same time. This logic is not implemented yet but is described in a separate issue: #181.

The other example you mention of the robotics factory and shipyard is however valid because those two belong to the same building queue.

Does this answer your question? Disclaimer: I'm relying on my own understanding of the official game. I might not be fully correct on some details. But of course when the feature is ready for testing I'll be happy to help out with testing and verifying the logic and to compare it to the official game.

Thanks again for your work on this, I'm really happy about your contributions and progress 😄 !

rautamik:
Hi @lanedirt,

Thanks for the feedback! I have made some adjustments based on the feedback.

If you don't mind I would keep the #181 issue separate from this PR even though it's closely related. I can look that case next after this PR is ready if you want.

I think that everything should be in place now regarding issue #299. Unless I have missed something. You could try it out when you have time and let me know if there are any issues or it's not working as it should be, thanks.

If everything is working fine then I'm looking forward for code review and will be glad to make any needed adjustments.

It has been a pleasure to work in this project.

app/Services/ObjectService.php
```diff
@@ -329,21 +329,41 @@ public function getBuildingObjectsWithStorage(): array
...
+                    return false;
+                }
+            }
+
             foreach ($object->requirements as $requirement) {
                 // Load required object and check if requirements are met.
                 $object_required = $this->getObjectByMachineName($requirement->object_machine_name);
+                $check_queue = $queued;
+
+                // SKip queue check for research lab as it must be present for research objects
```
lanedirt: Typo in the comment "SKip"

tests/Unit/BuildingQueueServiceTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+        // Assert that shipyard is in building queue
+        $this->assertTrue($this->building_queue->objectInBuildingQueue($this->planetService, 'shipyard', 1));
+
+        // Cancel robot factory
+        $this->building_queue->cancel($this->planetService, $queue_robot_factory->id, 14);
+        $this->building_queue->cancelItemMissingRequirements($this->planetService);
+
+        // Assert that shipyard is in building queue
+        $this->assertFalse($this->building_queue->objectInBuildingQueue($this->planetService, 'shipyard', 1));
```
lanedirt: I think comment here should read: "Assert that shipyard is no longer in building queue"

tests/Unit/ResearchQueueServiceTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace Tests\Unit;
+
+use OGame\Models\ResearchQueue;
+use OGame\Services\ObjectService;
+use OGame\Services\ResearchQueueService;
+use Tests\AccountTestCase;
+
+class ResearchQueueServiceTest extends AccountTestCase
```
lanedirt: Since this test now extends from AccountTestCase it makes this a `Feature` test as it communicates with the database. 

The goal for the tests in the `tests/Unit` folder is to be able to run without a database dependency and be able to mock all services instead.

So in this case if the new logic requires a database to test properly, could you move the whole file to the `tests/Feature` folder and adjust the namespaces accordingly?

tests/Unit/ObjectServiceTest.php
```diff
@@ -2,39 +2,77 @@
...
 namespace Tests\Unit;
 
+use OGame\Models\BuildingQueue;
+use OGame\Models\Resources;
 use OGame\Services\ObjectService;
-use Tests\UnitTestCase;
+use Tests\AccountTestCase;
 
-class ObjectServiceTest extends UnitTestCase
+class ObjectServiceTest extends AccountTestCase
```
lanedirt: My comment for `ResearchQueueServiceTest` applies to this file as well.

tests/Unit/BuildingQueueServiceTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace Tests\Unit;
+
+use OGame\Models\BuildingQueue;
+use OGame\Services\BuildingQueueService;
+use OGame\Services\ObjectService;
+use Tests\AccountTestCase;
+
+class BuildingQueueServiceTest extends AccountTestCase
```
lanedirt: My comment for `ResearchQueueServiceTest` applies to this file as well.

lanedirt (changes requested):
Hi @rautamik,

I just tested and reviewed the changes and it's looking good! I do have a few minor comments regarding namespaces and typos, see my comments inline. And good idea to keep the #181 issue separate, I agree.

--

Other than the small inline comments I did come across a new game logic bug which I didn't think of before.

Example:

1. Queue robotics factory level 1 + 2 in build queue.
2. Queue shipyard level 1 in build queue.
3. Now when I navigate to the shipyard page the `Solar Satellite` object can already be built while the `Shipyard` is not yet built (it's still in the queue).

In this case for being able to build units, the shipyard level requirement should actually be finished (built) and should not look what is in the queue.

I also noticed the same thing applies to technology/research. If I queue energy technology lvl 1 + combustion drive lvl 1, the light fighter becomes immediately buildable. In this case for unit requirements it should also only look for what is actually researched/built and not look in the queue.

Perhaps this is a good use case for adding an additional test to ensure this game logic is consistently tested with every change. 

Sidenote: I think the official game logic seems relatively simple from the outside, but I have to say that as we dig deeper into implementing all the finer details, it's clear there are quite a few edge cases and potential loopholes to consider. But that’s also what makes it interesting IMO! 😄

Could you have a look at these comments and suggestions? Thanks for your amazing work so far 🚀 !

rautamik:
Hi @lanedirt,

Thanks for the feedback! Good comments and suggestions.

I will fix the unit queue issues and add more test cases. I'm new to the game and missed totally the unit queue, sorry for that.

Regarding unit/feature tests, you are totally right and I will see if I changes those back as unit tests or if not then I will move those properly as feature tests.

I will look also into other issues and suggestions that you raised, thank!

tests/Unit/ResearchQueueServiceTest.php
rautamik: I refactored test cases and moved feature tests to proper place. I also added new feature cases to cover discussed use cases. Thanks for pointing out improperly done test cases.

rautamik:
Hi @lanedirt,

I have addressed all the issues you raised and made the feature test coverage better. Please check the PR when you have time, thanks.

lanedirt:
Hi @rautamik,

Thanks for making the changes! Just tested it and works flawlessly on my end. 👍 Awesome work. Will merge the PR!

PR #426, opened 2024-10-30
Shop view fixed on frontend side.

alican22:
Shop view is better now. But there is still many work to do.

- [x] Write a clean html code
- [x] Complete missed css codes
- [ ] Complete missed js codes.

### DIFF
--- a/app/Console/Commands/Tests/TestRaceConditionGameMission.php
+++ b/app/Console/Commands/Tests/TestRaceConditionGameMission.php
@@ -6,6 +6,7 @@ use Exception;
 use GuzzleHttp\Exception\GuzzleException;
 use Illuminate\Validation\ValidationException;
 use OGame\Models\FleetMission;
+use OGame\Models\Resources;
 use OGame\Services\ObjectService;
 
 /**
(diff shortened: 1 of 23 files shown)


lanedirt:
Hi @alican22,

Thanks for looking into this. The frontend looks a lot better with these changes. 👍

One thing I would like to ask has to do with the assets: I see the new CSS code includes some images from the original game servers, like this example:
```
#shopcomponent #planet {
    background: url("//gf3.geo.gfsrv.net/cdnb2/dbba5547458345f9e3b3baa09c3ef2.jpg") no-repeat;
    height: 250px;
}
```

In order to have OGameX be independent of the original game the goal is to include all required assets into the repository. I already did this for all existing assets which looks like this:

```
#messagecontainercomponent .content .singleTab.marker {
    background: url("/img/icons/112960c1ace80c7dcb03ca88d4b6fc.png") no-repeat;
}
``` 

Could you take a look at copying over the relevant files and replacing the external paths in the CSS with the new local relative paths? This will ensure that OGameX will always properly work, now and in the future.

Thanks again! 🚀

Issue #401, opened 2024-10-18
[BUG] Fix flaky espionage dispatch mission test


lanedirt:
**Describe the bug**
This test sometimes fails when running jobs:

```
  FAILED  Tests\Feature\FleetDispatch\FleetDispatchEspionageTest > dispatch…   
  Target planet was not updated after espionage mission has arrived. Check target planet update logic on mission arrival.
Failed asserting that 1729147297 is less than 1729147297.

  at tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php:191
```

For example, see this GitHub Action run:
https://github.com/lanedirt/OGameX/actions/runs/11373489520/job/31640153417

**To Reproduce**
1. See GitHub action result: https://github.com/lanedirt/OGameX/actions/runs/11373489520/job/31640153417
2. Most likely if you run the tests locally many times it will also sometimes fail. However I have not tested or confirmed this yet. Verify this by executing `php artisan test` in the `ogame-app` docker container.

**Expected behavior**
Test to always succeed, especially if nothing regarding espionage missions has been changed. The problem in this case can be caused by underlying game logic that is actually faulty, or it can be caused by something else that makes it flaky (i.e. fail sometimes while it shouldn't).

rautamik:
This issue can be reproduced locally (not tested in Docker), by running the fleet dispatch espionage tests in loop:
`while php vendor/bin/phpunit --filter FleetDispatchEspionageTest --stop-on-failure; do echo; done`
```
1) Tests\Feature\FleetDispatch\FleetDispatchEspionageTest::testDispatchFleetDebrisEspionageReport
Target planet was not updated after espionage mission has arrived. Check target planet update logic on mission arrival.
Failed asserting that 1729569638 is less than 1729569638.

tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php:191
```

There is also another flaky case with the same testDispatchFleetDebrisEspionageReport that pops up when running only that test in the loop:
`while php vendor/bin/phpunit --filter testDispatchFleetDebrisEspionageReport --stop-on-failure; do echo; done`
```
1) Tests\Feature\FleetDispatch\FleetDispatchEspionageTest::testDispatchFleetDebrisEspionageReport
Debris field metal resources are not visible in espionage report.
Failed asserting that 2674 matches expected 1337.

tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php:198
```

lanedirt:
Hi @rautamik,

[...]

PR #433, opened 2024-11-01
Fix flaky tests and refactor ObjectService

lanedirt:

PR #434, opened 2024-11-01
Set current time as test time in setup

rautamik:
Bugfix for #401

Changes:
Set current time as test time in setup

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -2,6 +2,7 @@
 
 namespace Tests;
 
+use Carbon\Carbon;
 use Illuminate\Contracts\Console\Kernel;
 use Illuminate\Foundation\Application;
 use Illuminate\Foundation\Testing\TestCase as BaseTestCase;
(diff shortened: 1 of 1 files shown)


rautamik:
Hi @lanedirt,

I noticed that you reopened flaky espionage dispatch mission test and thought that maybe we could simply set the current time in test setup to get these timing related flakies fixed. I have run the espionage dispatch mission test locally in loop with this fix and wasn't able to reproduce it anymore.

There is another problematic flaky test, testDispatchFleetCombatUnitsLostAndResourceGained, still randomly failing like while testing this PR.

Sorry to jump into something that you are already looking into, just came with an idea and wanted to see if it works...

lanedirt:
Hi @rautamik,

Thanks for also looking into this, more eyes is alway better 😄 ! 

I have already made a new PR #433 that includes the same as what you're proposing but with more tweaks to all existing tests. Now in the main account testcase the time is set statically to 2024-01-01 and individual tests simply advance time by calling `$this->travel()` which is a Laravel wrapper that works with Carbon. There were also some issues I noticed that (sometimes) occur when tests are ran against an empty database which has to do with setting/changing server settings.

All in all I think that PR #433 should address all issues and that we can close this PR as duplicate. Do you agree?

rautamik:
Yes, I agree that this is a duplicate and can be closed.

Nice to see that you have refactored lot of test cases, hopefully all flaky ones are now fixed. I was also thinking that travel would be more readable to use.

Issue #441, opened 2024-11-09
Add database management GUI via PhpMyAdmin docker container


lanedirt:
### Issue Description
Add a phpMyAdmin Docker container to the development setup in `docker-compose.yml` for easy database management, with default features for database export/import. Ensure the setup is secure for potential production deployment, allowing for customization based on environment.

### Key Requirements
1. **Development Default:**  
   Enable the phpMyAdmin container by default in the development `docker-compose.yml` setup.

2. **Production Toggle:**  
   Keep the phpMyAdmin container disabled in production by default, see `docker-compose.prod.yml`. Perhaps we could use `.env` file options to allow server admins to enable or disable phpMyAdmin as needed, ensuring it isn’t exposed unintentionally in development and/or production environments.

3. **Security Enhancements:**  
   - The current docker setups use default database credentials, see `.env.example`; improve security by generating unique credentials during the initial Docker setup.
   - Verify that phpMyAdmin access aligns with intended usage and security requirements to avoid data exposure risks.

**Note:** Document these setup steps and environment-specific configurations for ease of deployment and maintenance.

jackbayliss:
@lanedirt  Good idea, but rather than coding it and maintaining it, you could just attach phpmyadmin to the docker setup: https://hub.docker.com/_/phpmyadmin?tab=description  (and lock it down to IP etc and user / pass) 

And then you can access it and export / import it that way?

lanedirt:
> @lanedirt Good idea, but rather than coding it and maintaining it, you could just attach phpmyadmin to the docker setup: https://hub.docker.com/_/phpmyadmin?tab=description (and lock it down to IP etc and user / pass)
> 
> And then you can access it and export / import it that way?

Great suggestion! On second thought, you're absolutely right that coding and maintaining this ourselves could add unnecessary complexity. Especially when trying to import a new database while authenticated with a user from the existing database.

I'll update the issue 👍.

PR #445, opened 2024-11-10
Add database management 

jackbayliss:
This closes https://github.com/lanedirt/OGameX/issues/441 

### Default Setup
I've added the default phpmyadmin setup into the dev docker setup,  and the default port will be `8080`  - this can be edited.

### Production
For the production setup, rather than leaving it out for production based on env, I've included it, but I've added a volume line that essentially locks it down by IP, this means that essentially no one can access it, unless its the IP address in the `.htaccess` ie for example `Allow from 1.1.1.1`. 
This should then give you the ability for it to setup and work, but be locked down to your own IP address etc, so not just anyone can access it. 

If you try to access it without the right IP, you get the following - therefore if you alter this in production, it'll only be locked to you allowing easy import and export ability : 
image

You'd login with the DB / PASS the same as laravel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,9 @@ For local development you can run the default docker-compose file that is includ
   ```
   $ docker compose up -d
   ```
-  > Note: The default setup binds to ports 80/443. Modify `docker-compose.yml` if needed.
+  > Note: The default setup binds to ports 80/443. Modify `docker-compose.yml` if needed. PhpMyAdmin is also included for database management and is bound to port 8080. Modify docker-compose.yml if needed.
+
+
   
 4. Access the "ogame-app" Docker container:
   ```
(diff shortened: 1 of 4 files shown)


lanedirt:
Hi @jackbayliss,

Thanks for looking into this. I think this is a nice solution with the conditional .htaccess so access is disabled by default for production but the server admin can open it up if they want. Good idea! 👍 

As a bonus I think it would be good if we could update the README.md file to include information about the new exposed port 8080 and how the .htaccess file works for restricting access production.

My idea would be to add a small sentence to these parts:

> Note: The default setup binds to ports 80/443. Modify docker-compose.yml if needed.

E.g. we could make it this for the development section:

> Note: The default setup binds to ports 80/443. PhpMyAdmin is also included for database management and is bound to port 8080. Modify `docker-compose.yml` if needed.

And this for the production section:

> Note: The default setup binds to ports 80/443. PhpMyAdmin is also included for database management and is bound to port 8080, however to access it you need to explicitly specify your IP addresses via `./docker/phpmyadmin/.htaccess` for safety purposes. Modify `docker-compose.yml` or `.htaccess` if needed.

What do you think about this?

jackbayliss:
@lanedirt  All done, I've had to do it on my phone so please do adjust if anything is off :)

lanedirt:
Looks good to me, thanks for updating the readme. Merging!

Issue #457, opened 2024-11-22
[BUG] Shipyard queue does not work correctly on `resources` page


lanedirt:
**Describe the bug**
The shipyard queue on the resources page does not work correctly when trying to build solar satellites via the `resources` page.

**To Reproduce**
- I go to the resources page and build 20 solar satellites.
- Only one solar satellite appears in the buildings queue on the resources page. Once that satellite finishes, the page refreshes, showing an empty queue.
- However, on the overview page, I can still see the solar satellites actively building in the shipyard queue, while the shipyard queue on the resources page remains empty.

So it seems there may be a bug where the shipyard queue on the resources page isn't displaying its contents properly. 

**Expected behavior**
When 1 or more solar satellites are added to the queue, they should show up in the `shipyard queue` and not the `resources` queue. The correct amount should be visible and with the correct icon.

**Screenshots**
This is what shows up after trying to build 5 solar satellites. Notice how the solar satellite is incorrectly shown in the resources queue while the shipyard queue is empty. The solar satellites should instead be visible in the shipyard queue.
<img width="673" alt="Screenshot 2024-11-22 at 11 56 09" src="https://github.com/user-attachments/assets/643c50ae-f9ec-4074-9fba-909fdfd8acce">

**Additional context**
Related to changes made in #436.

Issue #405, opened 2024-10-20
[BUG] Building more than one Solar Satellite on the Resources page doesn't work


gabrieltsants:
**Description**
When attempting to build more than one Solar Satellite, only one is added to the build queue.

**To Reproduce**
1. Go to the 'Resources' page
2. Set more than one Solar Satellite to build
3. Click "Build"
4. Only one Solar Satellite will be queued for construction

**Expected behavior**
The specified number of Solar Satellites should be added to the build queue.

lanedirt:
Hi @gabrieltsants,

Thank you for reporting this issue. 👍  I can confirm it on my local version as well. It seems to be related to the resources page. The `solar satellite` object is a bit of a special case, as this unit can be built from both the `resources` page and the `shipyard` page.

When I try to build multiple solar satellites from the shipyard page, it works fine. Therefore, I believe the resources page is missing a check that fails to properly handle or respect the quantity field.

Additionally, I noticed that the solar satellite image does not display correctly when built from the resources page. Please see the screenshot for reference:
<img width="216" alt="Screenshot 2024-10-20 at 19 58 01" src="https://github.com/user-attachments/assets/b2ca9388-ff6c-434f-a5a2-60fe448d119b">

gabrieltsants:
Hi @lanedirt,

I noticed that the image issue also happens with the HyperSpace Drive.
![image](https://github.com/user-attachments/assets/204a3f08-25d8-4c10-8e62-e6db6e71fc8c)

Also, I’m not sure about the base version of the game, but the remaining time is the same for all three counters. In the screenshot above, you can see both the image error and the same countdown for all, **11m 12s**.

Regarding the Solar Satellite issue, I’ve been digging into the code and found that the Solar Satellite object is being treated as a Resource instead of a Ship. I started working on a fix, but I discovered there are many other areas that would need to be changed to make it work correctly. I have a lot of demands in my work at the moment, but as soon as I have more free time, I’ll continue looking into a solution. 😁

lanedirt:
Hi @gabrieltsants,

The timer which is the same for all three counters looks to me like a bug. Maybe this is a regression caused by a recent change. Feel free to open up a new issue for this if you can reproduce it, haven't tried it yet myself.

And thanks for looking further into the issue. The `solar satellite` is indeed a "strange" game object in the sense that this is the only object (that I know of) that can be built from both the resources and shipyard page. So if you find workarounds in the code this is probably the reason.

[...]

PR #458, opened 2024-11-22
Populate shipyard queue on resources page for solar satellites

lanedirt:
Fixes:
- #457 

Remaining todo's:
- [x] When solar satellite are building, when one unit is done the actual unit count is not updated on the resources page. This does work when building units from shipyard/defense page. Check what is missing to make this work. --> class where count is shown should be "amount" for unit objects instead of "level".


lanedirt:
Will close #457 and fixes additional bugs reported in #405

Issue #172, opened 2024-05-19
Implement "Intergalactic Research Network" logic


lanedirt:
**Is your feature request related to a problem? Please describe.**
The `Intergalactic Research Network` technology allows multiple research labs on multiple planets to work together to reduce research times. This time reduction logic is not implemented yet.

**Describe the solution you'd like**
- [ ] Reverse engineer how this IRN technology works and how it should apply to production times. Write specs.
- [ ] Implement specs.

**Additional context**

rautamik:
Hi @lanedirt,

Is this reliable enough for the specs?
https://ogame.fandom.com/wiki/Intergalactic_Research_Network

lanedirt:
Hi @rautamik,

Yep that looks like a nice and extensive overview of how it should work. Thanks for the link!

Geda173:
Closed in #460?

lanedirt:
@Geda173 Yep, thanks! Closing.

PR #460, opened 2024-11-23
Feature/Intergalactic Research Network

rautamik:
Feature #172 

Changes:
Add function to get Intergalactic Research Network Research Lab level.
Add Intergalactic Research Network Research Lab level feature test.

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -738,6 +738,28 @@ class PlanetService
         // Research speed is = (economy x research speed).
         $universe_speed = $this->settingsService->economySpeed() * $this->settingsService->researchSpeed();
 
+        // Intergalactic Research Network technology allows multiple research
+        // labs on multiple planets to work together to reduce research times.
+        $irn_level = $this->player->getResearchLevel('intergalactic_research_network');
+        if ($irn_level) {
+            $research_labs = $this->player->retrievePlanetResearchLabs()->filter(function (Planet $planet) use ($machine_name) {
+                // Check if object requirements are met in the planet
+                $planetService = resolve(PlanetService::class, ['player' => $this->player, 'planet_id' => $planet->id]);
+                if (!ObjectService::objectRequirementsMet($machine_name, $planetService, $this->player)) {
+                    return false;
+                }
+
+                // Exclude current planet as it's already included as a base value
+                if ($planet->id === $this->planet->id) {
+                    return false;
+                }
+
+                return true;
+            });
+
+            $research_lab_level += $research_labs->take($irn_level)->sum('research_lab');
+        }
+
         // The actual formula which return time in seconds
         $time_hours =
             (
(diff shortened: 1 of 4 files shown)


rautamik:
Hi @lanedirt,

I have implemented "Intergalactic Research Network" logic and feature test.

For some reason some tests like testDispatchFleetColonizeEmptyPlanet fails when I increased the number of planets created for feature tests from 2 to 4 in order to be able to test with multiple planets. Do you have any idea what is the correlation between number of planets created on  and AccountTestCase::dispatchFleet to succeed? The coordinates given to dispatchFleet are empty.

lanedirt:
> Hi @lanedirt,
> 
> I have implemented "Intergalactic Research Network" logic and feature test.
> 
> For some reason some tests like testDispatchFleetColonizeEmptyPlanet fails when I increased the number of planets created for feature tests from 2 to 4 in order to be able to test with multiple planets. Do you have any idea what is the correlation between number of planets created on and AccountTestCase::dispatchFleet to succeed? The coordinates given to dispatchFleet are empty.

Thanks for your time in implementing the logic! 

Regarding your question: I checked the correlation between amount of colonies and some tests failing. After some digging I found out that certain existing tests such as colonisation tests assume the player has only 2 planets at the start of the test. This is important for the `Astrophysics` technology as this controls the max. amount of colonies a player can have. 

As an example the test `FleetDispatchColoniseTest::testDispatchFleetColonizeEmptyPlanet()` sets the astrophysics to level 3 which would allow the user to have 1 planet + 2 colonies max. However with your change to give every test user 4 planets the astrophysics level is now too low and when the test tries to colonise an additional planet the user would have a total of 1 planet + 3 existing + 1 new (= 5 in total) while the user can only have 3 total with astrophysics level 3.

I think in this case it would be better if we refactor the test logic so every test can individually control how many colonies the user has at the start of the test instead of setting it globally in `AccountTestCase`. This will make it easier to add certain tests that require a different amount of colonies.

We could integrate that refactor in this PR, but as this increases the scope I'm also fine with merging this PR as-is and then fixing the failing tests with a refactor in a separate PR. Let me know what you think.

rautamik:
Thanks for looking into this issue and explaining what happens. It sounds a good idea to refactor the test logic to set the number of colonies per test. I can do it in this PR, no worries about the scope increasing. This is small PR anyways.

rautamik:
@lanedirt I ended up to a simple solution to pass amount of planets before test parent setup, what you think about it? Other possible solution could be to create and setup user in new function and call that one from every child after parent setup.

Btw, the failing FleetDispatchAttackTest test should be a flaky one.

lanedirt:
Hi @rautamik,

Looks good to me, nice and simple. In the future we could always refactor this part if it's necessary, but for now since it works I'm happy 😄!

I'll merge the PR. Thanks again for your time and help in making OGameX feature-complete 🚀 !

PR #500, opened 2024-12-17
[9.0] Windows entrypoint adjustments

jackbayliss:
This closes https://github.com/lanedirt/OGameX/issues/501 : 
- added a shebang to ensure the container knows what interpreter to use 
- uses [[ instead of [, which is more robust I believe
- directly invoke exec php-fpm, rather than passing it a string 
- adjusts spacing 

Also wrapped some ports I found in quotes 

Not sure which fixed the entrypoint issue, but seems to work locally now 

This works on Windows, but please check on mac etc, but I believe it should be fine :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -83,7 +83,7 @@ services:
     container_name: ogamex-phpmyadmin
     restart: always
     ports:
-      - 8080:80
+      - "8080:80"
     environment:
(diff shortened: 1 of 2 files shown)


lanedirt:
Just tested it on Mac, looks to be working fine there. Linux also works as the tests show. Thanks, merging!

Issue #391, opened 2024-10-14
Implement correct planet size calculation when creating a new planet


lanedirt:
Currently when a new planet is created a random planet size is chosen based on this logic in `PlanetServiceFactory::createPlanet()`:

```
        $planet->field_max = rand(140, 250);
```

So currently the planet size is fully random and does not take into account any other information.

The official game however implements certain rules for a small or big a planet can become based on the position of the planet. Up-to-date information can be found here: https://ogame.fandom.com/wiki/Planet_Size.

Todo:
- Change the planet size calculation logic to conform to how the official game works.

lanedirt:
Fixed by #507.

Issue #310, opened 2024-08-24
Implement correct planet temperature and use it in formula for solar satellite energy and deuterium mine production.


lanedirt:
The planet min/max temperature should be properly used/implemented for the things below.

Todo:

- [x] Update correct planet min/max temperature ranges which is determined and saved to the planet when creating a planet.
- [x] Solar satellite energy production is dependent on average planet temperature. The closer a planet is to the sun (e.g. position 1 in galaxy) the warmer the planet and the higher the solar satellite output.
- [x] Update feature test implemented in #182 to check that solar satellite output is different for various planets.
- [x] Deuterium mine output is also dependent on temperature: the lower the temperature the more deuterium is harvested: https://ogame.fandom.com/wiki/Deuterium_Synthesizer.
- [x] Add new DB migration to set `solar_satellite_percent` to 10 (default) for existing players. The new column added in a previous update has a default value of 0 in the database so it is set to 0% now for all existing players which is not desirable.

lanedirt:
Fixed by #508

PR #507, opened 2024-12-23
planet size calculation #391 fixes

yazilimmz:
I made changes based on [this page](https://ogame.fandom.com/wiki/Colonization). Adjustments were made according to the first planet.

### DIFF
--- a/app/Factories/PlanetServiceFactory.php
+++ b/app/Factories/PlanetServiceFactory.php
@@ -381,8 +381,6 @@ class PlanetServiceFactory
      */
     private function createPlanet(PlayerService $player, Coordinate $new_position, string $planet_name, PlanetType $planet_type): PlanetService
     {
-
-
         $planet = new Planet();
         $planet->user_id = $player->getId();
         $planet->name = $planet_name;
@@ -679,5 +677,4 @@ class PlanetServiceFactory
             'solar_satellite_percent' => 10,
         ];
     }
-
 }
--- a/app/Factories/PlanetServiceFactory.php
+++ b/app/Factories/PlanetServiceFactory.php
@@ -439,7 +439,6 @@ class PlanetServiceFactory
         $planet->field_max = rand($planet_data['fields'][0], $planet_data['fields'][1]) + $this->settings->planetFieldsBonus();
         $planet->diameter = (int) (36.14 * $planet->field_max + 5697.23);
 
-        // TODO: temperature range should be dependent on the planet position.
         $planet->temp_min = rand($planet_data['temperature'][0], $planet_data['temperature'][1]);
         $planet->temp_max = $planet->temp_min + 40;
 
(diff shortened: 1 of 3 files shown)


app/Factories/PlanetServiceFactory.php
```diff
@@ -428,13 +430,17 @@ private function setupMoonProperties(Planet $planet): void
...
-        $planet->diameter = 300;
-        $planet->field_max = rand(140, 250);
+
+        $planet_data = $this->planetData($planet->planet, $is_first_planet);
+
+        // Random field count between the min and max values and add the Server planet fields bonus setting.
+        $planet->field_max = rand($planet_data['fields'][0], $planet_data['fields'][1]) + $this->settings->planetFieldsBonus();
+        $planet->diameter = (int) (36.14 * $planet->field_max + 5697.23);
 
         // TODO: temperature range should be dependent on the planet position.
```
lanedirt: Can you remove this todo line since its now implemented

app/Factories/PlanetServiceFactory.php
```diff
@@ -443,11 +449,232 @@ private function setupPlanetProperties(Planet $planet): void
...
-        $planet->deuterium_synthesizer_percent = 10;
-        $planet->solar_plant_percent = 10;
-        $planet->fusion_plant_percent = 10;
-        $planet->solar_satellite_percent = 10;
+        $planet->metal_mine_percent = $planet_data['metal_mine_percent'];
+        $planet->crystal_mine_percent = $planet_data['crystal_mine_percent'];
+        $planet->deuterium_synthesizer_percent = $planet_data['deuterium_synthesizer_percent'];
+        $planet->solar_plant_percent = $planet_data['solar_plant_percent'];
+        $planet->fusion_plant_percent = $planet_data['fusion_plant_percent'];
+        $planet->solar_satellite_percent = $planet_data['solar_satellite_percent'];
```
lanedirt: Currently in OGameX the `*_percent` fields in the `planets` table are associated with the production percentage that the player can configure on the resource settings page, example: 
Screenshot 2024-12-23 at 21 30 47

If I read your code correctly then based on the position of the planet a bonus modifier should be added to the mine output. However to make this work we'll probably need to add this bonus modifier to the resource production formula. See `PlanetService::updateResourceProductionStatsInner()` for how this is currently implemented.

So for this PR I would propose to:
1. Revert this specific change back to the original version (set everything to "10" which represents 100% in the screenshot above)
2. Integrate the planet resource bonus modifier into the `PlanetService::updateResourceProductionStatsInner()` logic. However as this is outside of the scope of #391 we could also make a new issue for this to take on separately.

lanedirt (changes requested):
Hi @yazilimmz,

Thanks for your contribution and for looking into implementing the planet size calculations! 

Overall the changes to the field calculation looks good to me. I reviewed the changes and did add two comments, see below. Could you take a look at these?

If you have any questions feel free to let me know. Thanks again for your time in helping out, much appreciated! 🚀 😃

yazilimmz:
Hello,
Thank you for starting and leading this project; we truly appreciate your efforts.
You're absolutely right—I hadn't realized that the percentage pointed there. I've made adjustments accordingly.

Using this link:
https://proxyforgame.com/en/ogame/calc/production.php

I noticed that it corresponds to the base production, so I updated it based on that.
If you could create an issue and assign it to me for calculating energy and deuterium based on the planet's temperature, I’d be happy to look into it as well.

app/Services/PlanetService.php
```diff
@@ -1462,6 +1520,7 @@ public function getPlanetBasicIncome(): Resources
      */
     private function updateResourceProductionStatsInner(Resources $production_total, int|float $energy_production_total, int|float $energy_consumption_total, bool $save_planet = true): void
     {
+
```
lanedirt: It looks like nothing changed in this method `updateResourceProductionStatsInner()` except adding and removing a newline, can you revert the change on line 1523 and line 1482?

yazilimmz: > It looks like nothing changed in this method `updateResourceProductionStatsInner()` except adding and removing a newline, can you revert the change on line 1523 and line 1482?

I have updated the code again for your reference. Since the production calculations are added to the base production, you can see it in PlanetService::getPlanetBasicIncome().

lanedirt:
Hi @yazilimmz,

Thanks for the adjustments, functionally it looks good to me now. I noticed that some tests were failing, but as I was typing this I saw your new commit come in and now it looks like its working properly, nice 😄! Will merge this PR now. Welcome as an official contributor to the OGameX project 🚀 !

For the other changes you mentioned earlier regarding energy and deuterium output an issue already exists, see:
- #310

If you have the time to look at this it would be greatly appreciated :-)

PR #508, opened 2024-12-24
fixed #310 energy and deuterium mine production

### LINKED ISSUE #310: Implement correct planet temperature and use it in formula for solar satellite energy and deuterium mine production.

yazilimmz:
I've reviewed all of these, and it seems that they have all been resolved with this pull request :)

### DIFF
--- a/app/GameObjects/BuildingObjects.php
+++ b/app/GameObjects/BuildingObjects.php
@@ -78,7 +78,7 @@ class BuildingObjects
         $deuteriumSynthesizer->valid_planet_types = [PlanetType::Planet];
 
         $deuteriumSynthesizer->production = new GameObjectProduction();
-        $deuteriumSynthesizer->production->deuterium = 'return ((10 * $object_level * pow((1.1), $object_level)) * (-0.002 * $planet_temperature + 1.28))  * (0.1 * $building_percentage);';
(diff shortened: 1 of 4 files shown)


yazilimmz:
Additionally, I checked the formulas from the source code of [this site](https://proxyforgame.com/en/ogame/calc/production.php) and verified them through the OGame forums.

lanedirt:
LGTM, thanks for implementing the improved formulas! The dynamic solar satellite energy production also seems to work fine on my local environment. Merging! 🚀

PR #509, opened 2024-12-26
fixed #156 fixed #155 fixed #494

### LINKED ISSUE #155: Implement fleet dispatch mission deuterium consumption
Currently sending fleets does not consume any deuterium. Logic needs to be implemented to correctly calculate the deuterium consumption based on the fleet contents:

- Amount of ships
- Type of ships
- Distance between the start planet and target planet

We will also need to add unit/feature tests for this calculation logic.

### LINKED ISSUE #156: Implement max cargo carry capacity check when sending fleets
Currently there is no backend logic to check how much resources the user wants to send with a fleet. 

TODO:

- [ ] A check needs to be added  to all missions  which calculates the max cargo capacity of a fleet and uses this as sanity check when dispatching fleets.
- [ ] When dispatching a fleet, the deuterium consumption needs to be deducted from the current planet.

### LINKED ISSUE #494: [BUG] Duration of fleet mission shown in interface does not match actual duration
**Describe the bug**
Duration of flight shown in fleet dispatch screen does not match the actual flight time which is affected by game speed and/or research levels. See screenshots for the page where the incorrect flight time is shown.

**To Reproduce**
1. Via developer shortcuts set all research to level 10.
2. Send one or more light fighters to an attack mission to another planet.
3. The fleet dispatch screen will show `Duration of flight (one way)` with a certain duration. Take a mental note of this.
4. Send fleet. Now compare the actual duration with the previously shown duration. The actual duration is faster.

**Expected behavior**
I expect the shown duration in the fleet dispatch screen to match the actual duration.

**Screenshots**
Screenshot 2024-12-16 at 10 56 22
Screenshot 2024-12-16 at 10 56 41

**Additional context**
The issue is most likely caused by the Javascript that shows the duration of flight on the fleet dispatch screen is not provided with all correct server settings and/or research levels.

yazilimmz:
fixed #156 fixed #155 fixed #494

Reverse engineering was used for deuterium calculation.
Errors in flight durations were resolved.
The issue in flight distance calculation was fixed.
The circular universe, which has become standard in the normal game, was activated.
In deployment, when a fleet is deployed, it now leaves half of the deuterium on the target planet, just like in the original game.
The flight duration was always taken as 100%, but now it considers the selected percentage.
A few more fixes were also made. :)

### DIFF
--- a/app/Services/FleetMissionService.php
+++ b/app/Services/FleetMissionService.php
@@ -124,7 +124,7 @@ class FleetMissionService
         $diffSystem = abs($fromCoordinate->system - $to->system);
         $diffPlanet = abs($fromCoordinate->position - $to->position);
 
-        // Eğer galaksiler farklıysa
+        // If the galaxies are different
         if ($diffGalaxy != 0) {
             $diff2 = abs($diffGalaxy -  $this->settingsService->numberOfGalaxies());
 
@@ -135,7 +135,7 @@ class FleetMissionService
             }
         }
 
-        // Eğer sistemler farklıysa
+        // If the system are different
         if ($diffSystem != 0) {
             $diff2 = abs($diffSystem - 499);
             $deltaSystem = 0;
@@ -151,12 +151,12 @@ class FleetMissionService
             return $deltaSystem * 5 * 19 + 2700;
         }
 
-        // Eğer gezegenler farklıysa
+        // If the planet are different
         if ($diffPlanet != 0) {
             return $diffPlanet * 5 + 1000;
         }
 
-        // Eğer koordinatlar aynıysa
+        // If the coordinates are the same
         return 5;
     }
 
@@ -170,7 +170,7 @@ class FleetMissionService
      * @param $mission
      * @return float|mixed
      */
-    public function calcConsumption(PlanetService $fromPlanet, UnitCollection $ships, Coordinate $target_coordinate, int $holdingTime, float $speed_percent)
+    public function calculateConsumption(PlanetService $fromPlanet, UnitCollection $ships, Coordinate $target_coordinate, int $holdingTime, float $speed_percent)
     {
         $consumption = 0;
         $holdingCosts = 0;
--- a/app/Services/FleetMissionService.php
+++ b/app/Services/FleetMissionService.php
@@ -388,13 +388,13 @@ class FleetMissionService
      * @return FleetMission
      * @throws Exception
      */
-    public function createNewFromPlanet(PlanetService $planet, Coordinate $targetCoordinate, PlanetType $targetType, int $missionType, UnitCollection $units, Resources $resources, Resources $consumption_resources, float $speed_percent, int $parent_id = 0): FleetMission
+    public function createNewFromPlanet(PlanetService $planet, Coordinate $targetCoordinate, PlanetType $targetType, int $missionType, UnitCollection $units, Resources $resources, float $speed_percent, int $parent_id = 0): FleetMission
     {
         $missionObject = $this->gameMissionFactory->getMissionById($missionType, [
             'fleetMissionService' => $this,
             'messageService' => $this->messageService,
         ]);
-        return $missionObject->start($planet, $targetCoordinate, $targetType, $units, $resources, $consumption_resources, $speed_percent, $parent_id);
+        return $missionObject->start($planet, $targetCoordinate, $targetType, $units, $resources, $speed_percent, $parent_id);
     }
 
     /**
--- a/app/Services/FleetMissionService.php
+++ b/app/Services/FleetMissionService.php
@@ -180,10 +180,12 @@ class FleetMissionService
 
         $speedValue = max(0.5, $duration * $this->settingsService->fleetSpeed() - 10);
         foreach ($ships->units as $shipEntry) {
-            // $shipEntry içeriğini çözüyoruz
-            $ship = $shipEntry->unitObject; // ShipObject alınır
-            $shipAmount = $shipEntry->amount; // Gemi sayısı
-            $ship_speed = $ship->properties->speed->calculate($fromPlanet->getPlayer())->totalValue; // Geminin hızı
+            // Get the ship object and amount
+            $ship = $shipEntry->unitObject; // Ship object
+            $shipAmount = $shipEntry->amount; // Amount of ships
+
+            // Calculate the speed of the ship
+            $ship_speed = $ship->properties->speed->calculate($fromPlanet->getPlayer())->totalValue;
 
             if (!empty($shipAmount)) {
                 $shipSpeedValue = 35000 / $speedValue * sqrt($distance * 10 / $ship_speed);
--- a/app/Services/FleetMissionService.php
+++ b/app/Services/FleetMissionService.php
@@ -68,43 +68,149 @@ class FleetMissionService
      * @param UnitCollection $units
      * @return int
      */
-    public function calculateFleetMissionDuration(PlanetService $fromPlanet, Coordinate $to, UnitCollection $units): int
+    public function calculateFleetMissionDuration(PlanetService $fromPlanet, Coordinate $to, UnitCollection $units, float $speed_percent = 10): int
     {
         // Get slowest unit speed.
         $slowest_speed = $units->getSlowestUnitSpeed($fromPlanet->getPlayer());
+        $distance = $this->calculateFleetMissionDistance($fromPlanet, $to);
 
-        // Calculate distance between current planet and target planet.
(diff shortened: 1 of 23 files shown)


yazilimmz:
"The tests don't give any errors on my side either. Can you please try it on your side, @lanedirt ?

> www@d1dd6a39b15e:/var/www$ ./vendor/bin/phpunit
> PHPUnit 11.5.1 by Sebastian Bergmann and contributors.
> 
> Runtime:       PHP 8.4.2
> Configuration: /var/www/phpunit.xml
> 
> ....................................."Moon was created after amount of attempts: 1" // tests/Feature/FleetDispatch/FleetDispatchAttackTest.php:945
> ..........................  63 / 208 ( 30%)
> ............................................................... 126 / 208 ( 60%)
> ............................................................... 189 / 208 ( 90%)
> ...................                                             208 / 208 (100%)
> 
> Time: 04:12.284, Memory: 250.50 MB
> 
> OK (208 tests, 2782 assertions)

app/Services/FleetMissionService.php
```diff
@@ -68,42 +68,144 @@ public function __construct(PlayerService $player, MessageService $messageServic
...
         $fromCoordinate = $fromPlanet->getPlanetCoordinates();
-        $distance = 0;
-        if ($fromCoordinate->galaxy !== $to->galaxy) {
-            $distance = 20000 * abs($to->galaxy - $fromCoordinate->galaxy);
+
+        $diffGalaxy = abs($fromCoordinate->galaxy - $to->galaxy);
+        $diffSystem = abs($fromCoordinate->system - $to->system);
+        $diffPlanet = abs($fromCoordinate->position - $to->position);
+
+        // Eğer galaksiler farklıysa
```
jackbayliss: Can these be English? Seeing as the rest of the comments are :)

-
jackbayliss, line 1: This is an empty file, and not needed :)

app/Services/FleetMissionService.php
jackbayliss: Same for other comments

app/Services/FleetMissionService.php
```diff
@@ -68,42 +68,144 @@ public function __construct(PlayerService $player, MessageService $messageServic
...
+     * Calculate the consumption of a fleet mission based on the current planet, target coordinates and fleet.
+     * @param PlanetService $fromPlanet
+     * @param UnitCollection $ships
+     * @param Coordinate $target_coordinate
+     * @param int $holdingTime
+     * @param float $speed_percent
+     * @param $mission
+     * @return float|mixed
+     */
+    public function calcConsumption(PlanetService $fromPlanet, UnitCollection $ships, Coordinate $target_coordinate, int $holdingTime, float $speed_percent)
```
jackbayliss:
```suggestion
    public function calculateConsumption(PlanetService $fromPlanet, UnitCollection $ships, Coordinate $target_coordinate, int $holdingTime, float $speed_percent)
```
This should probably be calculate, like the other methods :)

jackbayliss:
@yazilimmz  The test that is failing is the `php artisan test:race-condition-unitqueue` test, if you run this command locally (via the php container) you should see the same error :)

yazilimmz:
> @yazilimmz The test that is failing is the `php artisan test:race-condition-unitqueue` test, if you run this command locally (via the php container) you should see the same error :)

Yesterday, I was getting an error in a PHPUnit test.

The same thing happened again.

While Git throws an error, the test passes successfully on my local machine.

The test error given by Git:

> ----------------------------------------------------------------------------------
> 
> Not enough resources on the planet to send the fleet.
> 
>   at tests/FleetDispatchTestCase.php:246
>     242▕             'speed' => 10,
>     243▕             ...$unitsArray
>     244▕         ]);
>     245▕ 
>   ➜ 246▕         $post->assertStatus($assertStatus);
>     247▕ 
>     248▕         $this->reloadApplication();
>     249▕ 
>     250▕         $this->get('/ajax/fleet/eventbox/fetch')->assertStatus(200);
> 
>   1   tests/FleetDispatchTestCase.php:143
>   2   tests/Feature/FleetDispatch/FleetDispatchAttackTest.php:676
> 
> 

However, this test passes successfully on my side, and I get 208 passed tests.

app/Services/FleetMissionService.php
```diff
@@ -68,42 +68,144 @@ public function __construct(PlayerService $player, MessageService $messageServic
...
+        $distance = $this->calculateFleetMissionDistance($fromPlanet, $target_coordinate);
+        $duration = $this->calculateFleetMissionDuration($fromPlanet, $target_coordinate, $ships, $speed_percent);
+
+        $speedValue = max(0.5, $duration * $this->settingsService->fleetSpeed() - 10);
+        foreach ($ships->units as $shipEntry) {
+            // $shipEntry içeriğini çözüyoruz
+            $ship = $shipEntry->unitObject; // ShipObject alınır
+            $shipAmount = $shipEntry->amount; // Gemi sayısı
+            $ship_speed = $ship->properties->speed->calculate($fromPlanet->getPlayer())->totalValue; // Geminin hızı
+
```
lanedirt: Can you update these comments to be in english

tests/Feature/FleetDispatch/FleetDispatchGenericTest.php
```diff
@@ -100,6 +104,30 @@ public function testFleetDurationCalculation(): void
         $units->addUnit(ObjectService::getShipObjectByMachineName('small_cargo'), 5);
 
         // Should take 2h:18m:05s to travel to the target planet 1 system away with base speed of 5000.
-        $this->assertEquals(8285, $fleetMissionService->calculateFleetMissionDuration($this->planetService, $targetPlanetCoords, $units));
+        $this->assertEquals(4788, $fleetMissionService->calculateFleetMissionDuration($this->planetService, $targetPlanetCoords, $units));
```
lanedirt: Can you update the comment stating how long the fleet is expected to take (in hour:minute:second format)

lanedirt (changes requested):
Hi @yazilimmz,

Thanks for the changes! Looks impressive and quite in-depth. 👍 And thanks @jackbayliss for your initial review, I agree with the comments. I found two more small things, can you check these?

I'll also try and see if I can reproduce the test error on my environment locally. I'll see if I can do this tonight.

lanedirt:
Hi @yazilimmz,

I checked the tests on my local environment and I got the error like 50% of the time when I ran the tests 10 times. I think it has to do with the deuterium costs especially with the `testLargeScaleAttackWithDebrisField()` method.


This test method sends a fleet to a random planet within a range of -15 systems to +15 systems. The test itself sends a fleet of 700k cruisers and 100k battleships. Based on this new code if I try to send this fleet to a planet 15 systems away the deuterium cost is 105M:

Screenshot 2024-12-27 at 17 15 14

And the test currently provides only 100M starting deuterium costs:
```
    public function testLargeScaleAttackWithDebrisField(): void
    {
        // Prepare attacker fleet
        $this->planetAddUnit('cruiser', 700000);
        $this->planetAddUnit('battle_ship', 100000);

        $this->planetAddResources(new Resources(5000, 5000, 100000000, 0));
```

So because of the random nature of the tests I think if we increase the deuterium on the planet at the start of the test this issue should be solved. Probably good to set it to like 150M to be safe?

yazilimmz:
> Hi @yazilimmz,
> 
> I checked the tests on my local environment and I got the error like 50% of the time when I ran the tests 10 times. I think it has to do with the deuterium costs especially with the `testLargeScaleAttackWithDebrisField()` method.
> 
> This test method sends a fleet to a random planet within a range of -15 systems to +15 systems. The test itself sends a fleet of 700k cruisers and 100k battleships. Based on this new code if I try to send this fleet to a planet 15 systems away the deuterium cost is 105M:
> 
> Screenshot 2024-12-27 at 17 15 14
> And the test currently provides only 100M starting deuterium costs:
> 
> ```
>     public function testLargeScaleAttackWithDebrisField(): void
>     {
>         // Prepare attacker fleet
>         $this->planetAddUnit('cruiser', 700000);
>         $this->planetAddUnit('battle_ship', 100000);
> 
>         $this->planetAddResources(new Resources(5000, 5000, 100000000, 0));
> ```
> 
> So because of the random nature of the tests I think if we increase the deuterium on the planet at the start of the test this issue should be solved. Probably good to set it to like 150M to be safe?

Nice approach, I hadn’t noticed the +15 -15 situation. I updated the code again. Thank you

lanedirt:
LGTM! One test failed but that one has been flaky for longer so probably doesn't have to do with this change. Merging the PR!

Thanks again for your work and contribution 😄 !

PR #523, opened 2025-01-06
Make navigation links in planet and moon tooltips work

lanedirt:


resources/views/ingame/layouts/main.blade.php
```diff
@@ -1095,12 +1095,21 @@ function initChatAsync() {
...
+                                       data-link="{{ $urlToPlanetWithUpdatedParam }}" title="<b>{{ $planet->getPlanetName() }} [{{ $planet->getPlanetCoordinates()->asString() }}]</b><br/>
+                                        @lang('Lifeform'): Humans<br/>
+                                        {{ OGame\Facades\AppUtil::formatNumber($planet->getPlanetDiameter()) }}km ({{ $planet->getBuildingCount() }}/{{ $planet->getPlanetFieldMax() }})<br>
```
OceanicNomad, line 1103: Add dynamic planet temperatures

{{ $planet->getPlanetTempMin() }} to {{ $planet->getPlanetTempMax() }}°C<br/>

lanedirt: Thanks for pointing this out! It isn't directly a part of this PR as this was only for fixing the menu links.

But this should indeed be an easy fix as well. 

If you have the time a PR for this would be welcome! 👍

Issue #531, opened 2025-01-18
[BUG]  Windows project docker installation fails


jackbayliss:
**Describe the bug**
Windows docker installation seems to fail 

**To Reproduce**
Setup the project, use Windows and docker - get the following:
```
[+] Running 7/7image sha256:f1bc270ddd9833dc5854317c26c983053c3e18d084c9ad6ef8d6e75af5dad074                       0.0s
 ✔ Service ogamex-app           Built                                                                             42.3s
 ✔ Service ogamex-scheduler     Built                                                                              4.4s
 ✔ Container ogamex-phpmyadmin  Running                                                                            0.0s
 ✔ Container ogamex-db          Running                                                                            0.0s
 ✔ Container ogamex-scheduler   Recreated                                                                          0.1s
 ✔ Container ogamex-webserver   Created                                                                            0.0s
 ✘ Container ogamex-app         Error                                                                              1.6s
dependency failed to start: container ogamex-app is unhealthy

```

It can't find the entrypoint in the docker container for me locally, even with fresh images etc.

PR #532, opened 2025-01-18
Fix windows docker build

jackbayliss:
## Description

- Removes the bash shebang and changes it for sh which seems to work for Windows 
- Changed the mariadb environment variables to be maria specifc, rather than mysql as made more sense as it's the mariadb docker image?

### Type of Change:
- [x] Bug fix

## Related Issues

Fixes https://github.com/lanedirt/OGameX/issues/531

## Checklist

Builds passes, so assuming it works on mac etc.. had a right nightmare on Windows.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,8 +12,8 @@ services:
     ports:
       - "3306:3306"
     environment:
-      MYSQL_DATABASE: laravel
-      MYSQL_ROOT_PASSWORD: toor
+      MARIADB_DATABASE: laravel
+      MARIADB_ROOT_PASSWORD: toor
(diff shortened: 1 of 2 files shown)


lanedirt:
Thanks for the Windows fixes! I'm not using Windows as a primary OS so did not notice this before. 

I'll try and look into if we can run additional GitHub actions on a Windows runner to assert if the docker build and entrypoint work correctly. A lot of people are running OGameX on Windows, at least for local dev, so I think this make sense to add to the automated tests.

Anyway thanks again for the PR, merging!

jackbayliss:
No problem! Yes an additional workflow would be a great idea

PR #542, opened 2025-01-21
Update HighscoreService.php

jackbayliss:
## Description

I missed this when doing it, but the logic should be inside the remember.

### Type of Change:
- [x] Bug fix


lanedirt:
LGTM, thx for the fix!

Issue #543, opened 2025-01-21
[BUG] Various resource DB fields are still int rather than float


jackbayliss:
**Describe the bug**
Various DB fields are ints for resources, rather than a float-- for example: building_queue,research_queues, unit_queues and planets (Looking at the migrations) 

We recently changed the debris table resources, to float, which makes me think this should be applied everywhere for resources?

PR #544, opened 2025-01-22
Resources Int to float

jackbayliss:
## Description
Various resource fields were still ints, rather than floats, this makes it uniform, 
I also changed the planet _max values to floats as they're already done in the migrations. 

I wonder if some of the _production columns on the planet table may also need to be changed - it would be good to go through all the columns ideally,  (one for another PR)

### Type of Change:
- [x] Bug fix

## Related Issues

Link to any issues or discussions that this PR addresses:
Fixes https://github.com/lanedirt/OGameX/issues/543

- [x] **Testing:**

### DIFF
--- a/app/Models/BuildingQueue.php
+++ b/app/Models/BuildingQueue.php
@@ -14,9 +14,9 @@ use Illuminate\Database\Eloquent\Model;
  * @property int $time_duration
  * @property int $time_start
  * @property int $time_end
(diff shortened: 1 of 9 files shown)


lanedirt:
Thanks again for looking into this and fixing the issue! Agreed that it would be good to check all data types in the future to see where floats make more sense.

PR #563, opened 2025-02-03
Update Dockerfile

### LINKED ISSUE #562: Optimize dockerfile for performance
The current dockerfile can cause performance issues because of this part

```
# Set write permissions for required directories
RUN chown -R www-data:www-data \
    /var/www/storage \
    /var/www/bootstrap/cache \
    /var/www/rust
```

Running a chown inside the dockerfile causes it to create a new layer and when these directories contains lots of files it can take minutes for this command to succeed. Replacing it with this instead should also work but with increased performance:

```
# Copy only necessary files first with correct permissions
COPY --chown=www-data:www-data storage /var/www/storage
COPY --chown=www-data:www-data bootstrap/cache /var/www/bootstrap/cache
COPY --chown=www-data:www-data rust /var/www/rust

# Then copy remaining files
COPY . /var/www/
```

lanedirt:
## Description
Dockerfile performance improvement

### Type of Change:
- [x] Other (please describe): performance improvement

## Related Issues
Fixes #562

## Checklist
- [ ] **Manual test:** Test with clean checkout that installation works as usual with this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,12 @@ RUN if [ $OPCACHE_ENABLE = "1" ]; then \
 # Install composer
 RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
 
-# Copy existing application directory contents
+# Copy only necessary files first with correct permissions
+COPY --chown=www-data:www-data storage /var/www/storage
+COPY --chown=www-data:www-data bootstrap/cache /var/www/bootstrap/cache
+COPY --chown=www-data:www-data rust /var/www/rust
+
+# Then copy remaining files
 COPY . /var/www/
 
 # Check if .env file exists, fail if it doesn't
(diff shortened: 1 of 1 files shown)


lanedirt:
These changes cause a new installation to throw this error:

```
The stream or file "/var/www/storage/logs/laravel.log" could not be opened in append mode: Failed to open stream: Permission denied The exception occurred while attempting to log: The stream or file "/var/www/storage/logs/laravel.log" could not be opened in append mode:
```

This needs some more testing.

lanedirt:
I just rebased the changes in this PR so its in-sync again with the main branch. After that I tested it again on my local Macbook and tested it on a Linux server. With these changes it works for me now for both new and existing installations. I did not get any permission errors anymore.

@jackbayliss when you have time, could you try and test this branch on your machine too and rebuild docker containers as a double-check?

jackbayliss:
Seems good for me locally, would be nice to reduce the copys into one and use && which should reduce the docker layers = faster

jackbayliss (approved):

lanedirt:
> Seems good for me locally, would be nice to reduce the copys into one and use && which should reduce the docker layers = faster

Thanks for testing! And good catch, just pushed the fix that combines all operations in one layer.

PR #594, opened 2025-02-27
Improve docker build

jackbayliss:
## Description

Wip to test if any increases can be made, 

- [x] Use image to build app and scheduler, rather than rebuilding 2x 
- [ ] Switch to fpm alpine for a quicker build time
- [ ] Sort rust issues once doing the above ^ 
- [ ] While here look at permission issues / see if we can replicate 

### Type of Change:
- [ ] Bug fix
- [X] Feature enhancement
- [ ] Documentation update


jackbayliss:
Rust stuff has baffled me, so for now comparing image changes,  will revisit 😴

Issue #363, opened 2024-09-27
Implement worker queue / cron system to chronologically process fleet missions


lanedirt:
## Current Situation
- Fleet missions are processed during user requests via `PlanetService::updateFleetMissions()`.
- This approach works for the current user but causes issues with multiple players and missions.

### Inconsistent mission processing
The current system can lead to inconsistent mission processing:

1. Player A schedules an attack on planet [1:1:1] to arrive at 10:00.
2. Player B schedules an attack on the same planet [1:1:1] to arrive at 10:15.
3. If Player A logs off and Player B stays online, the 10:00 mission isn't processed until Player A returns.

This creates potential conflicts and unfair gameplay scenarios.

### Large battle processing
Large battles with millions of units can take a relatively long time to process. PR #517 introduced the Rust Battle Engine which is up to 200x more performant than the PHP Battle Engine, which already saves a lot of time. However there are still usecases where even the Rust Battle Engine takes up too much time to process during a user request.

E.g. issue #503 mentioned a usecase where the defender player had 28M destroyer units on the planet. With a battle of 15M attacker units and 28M defender units the Rust Battle Engine takes up to 800MB of memory and 10+ seconds to process.

Moving over the fleet mission processing to the worker queue will prevent user requests from hanging for multiple seconds in case of very large battles.

## Proposed Solution
Implement an independent worker queue / cron system for background fleet mission processing.

### Requirements
1. Process fleet missions in real-time, independent of user logins.
2. Handle missions in strict chronological order.
3. Prevent conflicts between simultaneous or closely timed missions.

### Key Considerations
- [ ] Aim for near real-time processing: missions should be executed within 1-3 seconds of their scheduled time.
- [ ] Ensure scalability to handle multiple concurrent missions across the game universe. E.g. missions for different planets that do not conflict with each other can be processed concurrently and do not necessarily have to wait for each other.
- [ ] Implement logging and error handling for mission processing.
- [ ] Consider edge cases like server downtime or processing delays.

## Next Steps
1. Design the background processing system architecture.
2. Choose appropriate technologies (e.g., Redis for queue, Laravel's job system).
3. Implement a prototype and test with various scenarios.
4. Gradually migrate from the current system to the new background processing system.

## Additional Notes
- This change will significantly improve gameplay fairness and consistency.
- May require adjustments to other game mechanics that interact with fleet missions.
[...]

PR #598, opened 2025-03-01
Add Discord exception handler & various improvements

jackbayliss:
## Description

- Adds a a way to post exceptions to Discord, as it may give easier visibility to anyone running OGameX of any exceptions occurring - it may also help people that may not have server access, but can easily see any exceptions. You can lock these channels down to specific users etc, as you see fit-- so no issue here. 
- Adds env changes for the above, leaving webhook blank
- Removed `Kernel.php` as no longer in Laravel 11/12, and changed over into the `app.php` 
- The custom exception handler wasn't working, this adds a singleton to sort that, which allows us to do the Discord stuff.
- Adds queue worker to the docker containers, responsible for running the queue jobs (when a message is posted to discord the package sends a job), but this is also useful for https://github.com/lanedirt/OGameX/issues/363

### Type of Change:
- [X] Feature enhancement

Looks like:
image

## Checklist

- [x] **Testing:**
    **There's no tests for the scheduler or queue worker as feels a bit redundant to test**

- [x] **Documentation:** **Might be worth documenting all the different containers, and why - dont think we do atm, but something to do in the future ? **


jackbayliss:
In order to test the new container, you should be able to [create a webhook on Discord](https://support.discord.com/hc/en-us/articles/228383668-Intro-to-Webhooks) and throw an exception, you should then see it output (re image above)

app/Exceptions/Handler.php
```diff
@@ -5,49 +5,50 @@
...
      */
-    public function render($request, Exception|Throwable $e): \Symfony\Component\HttpFoundation\Response
+    protected function sendToDiscord(Throwable $e): void
     {
-        return parent::render($request, $e);
+        if (config('app.discord_alert_webhook') && $this->shouldReport($e)) {
+            $stackTrace = explode("\n", $e->getTraceAsString());
+
+            // Limit to the first two stack trace lines
+            $limitedStackTrace = implode("\n", array_slice($stackTrace, 0, 2));
```
jackbayliss, line 36: We could change this to more lines, but at least for now it's something.

lanedirt: Looks good to me, thanks for the work on this. I'll merge this PR and set up the discord webhook to test if it works correctly 😄 !

lanedirt (approved):

Issue #590, opened 2025-02-24
[BUG] Building ship while spaceyard is updating should not be possible


Lightounet:
Construction of ships is possible when spaceyard is also in construction, should not be possible.

PR #606, opened 2025-03-05
fix : Ensure ships and defense cant be built when shipyard upgrading and shipyard cant be upgraded when either unit types are being built

jackbayliss:
## Description

Ensures if the shipyard is upgrading ships can't be built. Not sure if this is part of the original game, hence review asked for :)  

### Type of Change:
- [X] Bug fix

## Related Issues

Fixes https://github.com/lanedirt/OGameX/issues/590

## Checklist

## Additional Information

This seems like the cleanest way to make sure ships can't be built when the shipyard is being upgraded,  I've also made sure the user gets a nice error back without the need to alter the JS etc. It overrides the abstract method from the parent to add this additional piece of logic.

image


lanedirt:
Thanks for looking into this! I just checked how it works in the official game, see these two screenshots:

Screenshot 2025-03-06 at 10 55 57
Screenshot 2025-03-06 at 10 56 12

I also checked what happens if via developer tools I enable the "build" button and try to build it anyway, then this shows up:
Screenshot 2025-03-06 at 10 59 16

So based on what I see:
- When the shipyard page loads it should already check if the shipyard building is currently being upgraded. If so, then all ships shown on the page should already be disabled with the tooltip text "Shipyard is being upgraded.".
- The "build" button when clicked on a specific ship should be disabled and should also show the tooltip text "Shipyard is being upgraded.".
- When a build request is received while the shipyard is being upgraded, the error that you already added should be shown with the text "Shipyard is being upgraded.".

I also just checked and noticed that this same checking mechanism has already been implemented for `Research lab` building and `Research` page. So when the research lab is upgrading, you cannot start research (and vice versa). So might warrant to take a look at how this is implemented for the Research page and build queue in case some methods can be reused. 😄

jackbayliss:
@lanedirt  Thanks for that! It should now be sorted: 

Initial hover over shows the tool tip too
image

When clicking the object, the button shows the tooltip
image

If you bypass the disabled
image

resources/views/ingame/ajax/object.blade.php
```diff
@@ -184,12 +184,15 @@
...
+                            @php
+                                $disabled_shipyard_upgrading = ($object->type == \OGame\GameObjects\Models\Enums\GameObjectType::Ship && $shipyard_upgrading)
+                            @endphp
+                            @if (!$enough_resources || !$requirements_met || !$valid_planet_type || $build_queue_max || !$max_build_amount || $research_lab_upgrading || ($object->machine_name === 'research_lab' && $research_in_progress || $disabled_shipyard_upgrading))
                                 disabled
                             @else
                             @endif
                             data-technology="{{ $object->id }}">
-                        <span class="tooltip" title="">
+                        <span class="tooltip" title="{{ $disabled_shipyard_upgrading ? 'Shipyard is being upgrading' : '' }}">
```
lanedirt: Small typo here, should be 'Shipyard is being upgraded'

jackbayliss:
@lanedirt  Thanks! Will sort this later today and we should be there, appreciate you walking me through what it should be- don't play the game so 😂

jackbayliss:
image
image

We should now be closer! Let me know if there's anything else I need to do / missed / changes :)

app/Http/Traits/ObjectAjaxTrait.php
```diff
@@ -55,8 +55,6 @@ public function ajaxHandler(Request $request, PlayerService $player): JsonRespon
         // Switch
         $production_time = '';
         $production_datetime = '';
-        $research_lab_upgrading = false;
```
jackbayliss, line 58: Pointless declaring these as false so changed these to null coalescing to false

lanedirt:
Thanks for the fixes and the effort you’ve put into this! Some things may seem simple on the outside with OGame but can turn out to be more complex than they appear. 😺 I just reviewed the latest version and noticed two more things:

1. When the shipyard is busy, it shows it correctly here. However the little green arrow in the top left corner (which is the quick build) should be invisible if the building cannot be build. Also I checked if I click it, the shipyard build order gets executed. So the check if the building can actually be built seems to be missing from the quick build mechanism.
Screenshot 2025-03-07 at 23 12 14

I have a screen recording here that shows that the quick build works even though the building should not be able to be build:
mov2gif_Screen Recording 2025-03-07 at 23 15 14


2. When I navigate to the `/defense` page, I get the following error. I guess this might be caused by the fact that the unit template is used for both the shipyard and defense pages.
Screenshot 2025-03-07 at 23 12 28

jackbayliss:
@lanedirt  Thanks for that! Should be OK now, apologies for the back and forth - I guess the simple things turn out to be much more difficult :D

app/Http/Traits/ObjectAjaxTrait.php
```diff
@@ -65,8 +63,13 @@ public function ajaxHandler(Request $request, PlayerService $player): JsonRespon
 
                 // Research Lab upgrading is disallowed when research is in progress
                 $research_in_progress = $player->isResearching();
+
+                // Shipyard upgrading is now allowed when ship/s in progress.
+                $ship_in_progress = $player->isBuildingShips();
                 break;
             case GameObjectType::Ship:
+                $shipyard_upgrading = $player->isBuildingObject('shipyard');
+                break;
```
lanedirt, line 75: Making this GameObjectType::Ship case separate because of the added `$shipyard_upgrading` variable makes the `Ship` case miss the `$production_time` and `$production_datetime` variables.

I noticed this because the `production duration` for the ships was suddenly empty for me:
Screenshot 2025-03-08 at 00 54 38

I think adding these two variables to the `Ship` case should fix this regression

lanedirt (changes requested):
Thanks! All major mechanics now seem to work as they should 👍 . I did notice one minor regression issue though, see my comment above.

app/Http/Traits/ObjectAjaxTrait.php
jackbayliss: Thanks! Nice catch, feel free to re-review when you get time, no rush.

lanedirt: Looks good now! Thanks for the fix!

lanedirt:
Great work! I just tested it again, and after playing around with it a bit more, I realized there's one more detail I overlooked in my initial tests. (This was also not mentioned in the original issue.)

Right now, all restrictions seem to be working well for the `Shipyard` page/queue. If a ship is being built, the shipyard is blocked, and vice versa. This part is working as expected.

However, what I realized just now is that the `Defense` page/queue is also considered part of the `Shipyard`. That means if a defense unit is being built, the shipyard should be blocked from being upgraded. Similarly, if the shipyard is being upgraded, the defense page should be blocked and not allow building defense units.

Would it be possible to implement these added checks for the `Shipyard` page and `ship` units so these apply to the `Defense` page and `defense` units too so that the same restrictions apply?

jackbayliss:
@lanedirt  No trouble, I think it should now work for defense units from what I can see locally: 

image

Will pick up anything else tomorrow now 🫡

lanedirt (approved):
Just stress tested the latest version locally, everything I throw at it seems to be working fine now. Great stuff, nice addition to make the gameplay more fair.  Thanks for your effort on this!

Good to merge from me 👍!

Issue #609, opened 2025-03-10
[BUG] Changing fleet destination while a param is empty causes exception


jackbayliss:
**Describe the bug**
When using the fleet system, if you fully empty one of the inputs, ie the galaxy, and switch from say planet to moon, it gets stuck loading and causes an exception on the backend 

**To Reproduce**
1. /fleet
2. empty the first co-ordinates box
3. Click moon or whatever you like
4. See loading, and network error

**Expected behavior**
It works and lets you switch without posting the data, if there's empty params.

PR #610, opened 2025-03-10
fix : Ensure galaxy,  system, position and type available

jackbayliss:
## Description
When using the fleet system and emptying the galaxy box for instance, if you switch from say planet to moon, it will cause an exception and forever load.  

Found this while looking into exceptions via webhook. 

This basically ensures all the params are there, otherwise it'll stop loading - this seems to make the most sense here as it's where the params are being prepared to post.

### Type of Change:
- [X] Bug fix

## Related Issues
Fixes https://github.com/lanedirt/OGameX/issues/609

## Checklist

### DIFF
--- a/public/js/ingame.js
+++ b/public/js/ingame.js
@@ -76903,6 +76903,12 @@ FleetDispatcher.prototype.fetchTargetPlayerData = function () {
     params.recycler = 1;
   }
 
+    // Ensure all our params are existent, otherwise no point posting the data.
+    if (!params.galaxy || !params.system || !params.position || !params.type) {
+        this.stopLoading();
+        return;
+    }
+
   $.post(this.checkTargetUrl, params, function (data) {
     let status = data.status || 'failure';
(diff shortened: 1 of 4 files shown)


lanedirt:
Looks good to me! 👍  I did try and check just now though how the official game handles this, and if I reproduce it on there I see that the fleet check call still comes with a valid response when the position is empty:

Screenshot 2025-03-10 at 09 30 22

So I'm guessing that they are defaulting the position (or any other coordinates) to "0" on the serverside, so even if the client does not pass along a value the server side call still succeeds and returns a valid response (with no planets/missions found, which is fine). I'm not sure how complicated it is to change the server side logic for this call to make it work without changing the client side JS, but either way seems like a fine solution to me to prevent the exception/hanging loader.

jackbayliss:
Merging for now, as will prevent the errors-- then will potentially refactor the backend at a later point.

PR #614, opened 2025-03-15
Update Dockerfile

jackbayliss:
## Description
If there's no .env present, we can ensure it copies the example.

The prod .env copy is still required, as it ensures the prod .env works.


lanedirt:
Yea that sounds like a good idea to simplify the setup. 👍 

If this works then perhaps we can remove the manual copy step from the README.md documentation, or leave it in but mark it as optional. And then also good to modify the `.github/workflows/run-docker-compose-prod.yml` workflow file to remove the .env copy step, so it can test this default copy behavior.

jackbayliss:
@lanedirt  This PR is the best I could come up with to sort it automatically, we'll still use the prod env copy for the prod test, but maybe easier for new comers.

Happy to close this if you don't think it's worth it 🤔

lanedirt:
> @lanedirt This PR is the best I could come up with to sort it automatically, we'll still use the prod env copy for the prod test, but maybe easier for new comers.
> 
> Happy to close this if you don't think it's worth it 🤔

Looks good to me as-is! Thanks for the troubleshooting on this :).

I agree that making it easier for getting a development instance running locally is always a good thing.

lanedirt (approved):

PR #628, opened 2025-03-26
Fix wrong ressources stats behavior & Change Ressource Setting view

### LINKED ISSUE #375: Resource production does not match shown values in tooltips
## Problem summary
The resource production that the tooltip shows does not match the resource production that the "?" popup shows. I'm not sure yet if this has to do with server speed settings (economy) or if there is a bug somewhere else.

The screenshot for example shows the following differences all for metal mine level 1 with server settings economy speed/research speed/fleet speed all set to `1`.
- Tooltip shows per hour: `96`
- The metal mine "?" properties popup shows per hour: `220`
- The resource settings page shows per hour: `63` (30 base + 33 metal mine level 1)

## Todo
Investigate what causes these differences and adjust the logic so that the following places all show the correct resource production:

- [ ] Tooltip when hovering over the metal/crystal/deuterium/energy on top of the screen.
- [ ] The "?" properties popup of the respective buildings (metal mine, crystal mine, deuterium synthesizer, solar plant, fusion reactor)
- [ ] The resource settings page (URL `/resources/settings`) 
- [ ] Test with various server settings for economy speed to ensure all values always match.

### Screenshots

Screenshot 2024-10-04 at 09 29 06

Screenshot 2024-10-04 at 09 33 54

veikoon:
## Description
Please include a summary of the changes made to OGameX and their purpose. Clearly describe what this PR does and why it is necessary.

### Type of Change:
- [x] Bug fix
- [x] Feature enhancement

## Related Issues
Fixes #375 

## Additional Information

- Remove energy from Resources sum() function as the sum function is used to get the ammount of resource product by a Building (therefor - the energy used should not be part of the calcul)
- Add a verification on TechTreeController to check if the Building is an energy producer to get the energy production instead of sum() --> As the Resource sum function does not take energy anymore
- Fix Resource getFormatted() functions as in ressource blade a parametter were given (nb of hour for day and week) but were not implemented. --> The numbers of materials product are now formatted
- Change the Energy collumn in the Ressource setting to transform the view as in the real Ogame (energy_used / energy_max) instead of the addition of energy diff over days and week as cumulation of energy doesn't really make sense because energy diff is lost.
- Fix the calculation of Building Techinfo (the building percentage is define on a 10 base instead of a 100 one)
- Fix the Header Current production by removing a updateResourceProductionStatsInner (the Current production were actualy base income + 2 * Building income because of the double call) --> I do read the following explanation :
// 1. First time in order to ensure the energy production is up-to-date.
// 2. Second time for the mine production to be updated according to the up-to-date (actual) energy production
But I didn't make sense in my mind, and removing one call and testing just patch the issue.

{2384199D-99E5-45C2-B761-91A754823D7E}
--> 515 + 36.9 = 551.9 ~= 551

{C835CEA1-EA3B-4F29-ABD4-C1F93D9A960F}
--> The stats are corresponding to the actual production

{66F2BBDD-A035-47A2-95DF-17E590C22992}
--> The formatting is correct, you can view the exact number with your mouse & the energy is the live consumption


veikoon:
Mmh It seems that the double update for planet needs to be done for the createAndSetPlanetModel from Unit Tests.

app/Http/Controllers/TechtreeController.php
```diff
@@ -105,6 +105,8 @@ public function getProductionTable(GameObject $object, PlayerService $player): V
             return view('empty');
         }
 
+        $energy_building = ($object->id === 4 || $object->id === 12) ? true : false;
```
lanedirt: Can you change this so the comparison is done based on the `machine_name` property instead of the `id`. This makes it more readable and in line with the rest of the codebase.

app/Models/Resources.php
```diff
@@ -38,7 +38,7 @@ public function add(Resources $other): void
      */
     public function sum(): float
     {
-        return $this->metal->get() + $this->crystal->get() + $this->deuterium->get() + $this->energy->get();
+        return $this->metal->get() + $this->crystal->get() + $this->deuterium->get();
```
lanedirt, line 41: Can you update the comment for this method to read: `Returns sum of all resources (except energy)`

app/Services/PlanetService.php
```diff
@@ -509,6 +509,28 @@ public function energy(): Resource
...
+     * Get planet energy generated.
+     *
+     * @return int
+     * Energy amount.
+     */
+    public function getEnergyMax(): int
+    {
+        return $this->planet->energy_max;
+    }
+
```
lanedirt: There are already implementations for these two methods called:
- `energyProduction()`
- `energyConsumption()`

Can you remove these duplicate methods and call the existing methods instead?

app/Services/PlanetService.php
```diff
@@ -1459,7 +1481,6 @@ public function updateResourceProductionStats(bool $save_planet = true): void
         // 1. First time in order to ensure the energy production is up-to-date.
         // 2. Second time for the mine production to be updated according to the up-to-date (actual) energy production.
         $this->updateResourceProductionStatsInner($production_total, $energy_production_total, $energy_consumption_total);
-        $this->updateResourceProductionStatsInner($production_total, $energy_production_total, $energy_consumption_total);
```
lanedirt, line 1462: Great catch on identifying the duplicate method call! After revisiting this, it’s clear that this is one of the main causes of the discrepancy in resource production. The method itself directly appends the calculated production values to the planet object. So when called twice, it will append these values again, which isn’t correct.

The reason the method is called twice was to respect the correct order of operations: first, energy production/consumption needs to be calculated, and then those totals are used to calculate mine production, as they rely on accurate energy data. However now that you have removed the second call, in regular gameplay it would now take two separate requests for everything to be properly calculated. This is why you saw the unit tests failing, as everything is done in a single request there.  Adding back the second pass only in the unit tests is not ideal as the tests now have different logic compared to real player gameplay requests, though I understand your reasoning behind it for trying.

It would likely be better to refactor `updateResourceProductionStatsInner()` further and break it into distinct steps:

1. Calculate energy production/consumption for all objects and set it on the planet model. We can ignore the resource production (metal/crystal/deut) at this stage.
2. Make a second pass that only handles resource production, ignoring energy consumption, so that the two steps don't overwrite each other.

For naming, perhaps:
1. `updateEnergyProductionStats()`
2. `updateResourceProductionStats()`

Then, for the third step, the logic for the fusion plant would rely on the output from step 2, so that could be handled as a separate part.

If you have time to look at this further to improve it, feel free to do so. But if you don't I'm also happy to merge your current change as-is as it's already better than the previous implementation. The additional refactoring can then happen in a later PR.

lanedirt (changes requested):
Hi @veikoon,

Thanks for putting the time in to investigate the resource production issue and opening up a PR, nice work!

I have tested the code locally and reviewed the changes. I have a few comments. Could you take a look at these? If you have any thoughts, feel free to share.

I do need to say: kudos for the depth of your debugging! Seems like you did find what is most likely the root cause in that duplicate resource production calculation. So good job on the deep-dive! 😄 With this version all calculations on my local testing environment now indeed seem to match so I'm very happy 👍 !

veikoon:
Hi @lanedirt !

I'm glad you appreciate my humble work 😃 
I listened to you and applied the changes you requested.

As you said I split the updateResourceProductionStatsInner() behavior, please review it as it is not quite easy to explain.
Here is the logic :
- Get the energy consumption & production from all building except the Fusion Reactor
- "Save" the energy max & used

If player has at least 1 deuterium:
- Calculate the energy produced by the fusion plant and "save" it
- Calculate the deuterium produced by the deuterium synthesizer (In this case the Fusion Reactor can help the production of more deuterium)

If player don't have any deuterium left:
- Calculate deuterium production (without Fusion Reactor)
- Calculate Fusion reactor production

If the production is still positive :
- Add deuterium & energy production (Without the fusion reactor boosting the Deuterium production)

Else:
- Nothing is produced ? Honestly I don't know the game that much, It make sens to me to not produce energy at all rather than calculate a ratio based on how much deuterium is produced compared to the maximum needed.

Then:
- Calculate the Metal & Crystal production (With the energy boost from Fusion Reactor)

I took profit from a new code session to patch other strange behavior here is the list :

Fusion Reactor :
Add building_percentage to the operation as mentionned in the wiki
Remove resource_production_factor from the deuterium (minus) production of the fusion reactor as it should not impact deuterium consumption
Remove the Fusion Reactor from the getResourceProductionFactor calculation as the fusion reactor should not impact the factor for other buildings
Set the fustion reactor stat in ressource setting view overmarked to be more visible

Set the production_factor to float as the building resource production result is rounded anyway, and the sum of energy consumed by buildings was not equal to the available energy (due to an approximation on the factor)

Please tell me what you're thinking of all these changes, for now I won't push new features on this MR but I will correct things as you desire.

app/Services/PlanetService.php
```diff
@@ -1643,17 +1663,23 @@ public function getObjectProduction(string $machine_name, int|null $object_level
...
-        $energy_technology_level = 0; // Implement energy technology level getter.
+        $energy_technology_level = $this->player->getResearchLevel('energy_technology');
         $universe_resource_multiplier = $this->settingsService->economySpeed();
 
         $production = new Resources(0, 0, 0, 0);
 
         $production->metal->set((eval($gameObject->production->metal) * $universe_resource_multiplier) * ($resource_production_factor / 100));
         $production->crystal->set((eval($gameObject->production->crystal) * $universe_resource_multiplier) * ($resource_production_factor / 100));
-        $production->deuterium->set((eval($gameObject->production->deuterium) * $universe_resource_multiplier) * ($resource_production_factor / 100));
+        // Fusion plant does not require energy tu consume deuterium so it should not be impacted by resource_production_factor
```
lanedirt, line 1676: Typo: `tu consume` --> `to consume`

app/Services/PlanetService.php
```diff
@@ -1563,49 +1562,70 @@ public function getProductionForPositionBonuses(int $position): array
...
+            // As a professional ingeeneer, we will suppose that the fusion plant will switch to Safety program
+            // And will stop producing energy if not enough deuterium is available.
+            // To ensure that the working technicians stay safe the fusion plant will still consume the deuterium to avoid explosions.
+            // TODO : Ask to an Ogame professional Engineer Doctor what the fusion plant is supposed to do.
         }
 
+        // Calculate the metal & crystal production normaly.
+        $production_total->add($this->getObjectProduction('metal_mine'));
+        $production_total->add($this->getObjectProduction('crystal_mine'));
+
```
lanedirt, line 1624: Instead of hardcoding production of metal/crystal/deuterium by looking directly at the buildings `metal_mine`, `crystal_mine` (and `deuterium_synthesizer` in this method above), it would be better to stick to the original implementation where it loops through all possible buildings and calculates the total of all production.

This is important because it's possible that in the future other buildings will be added that can produce metal/crystal/deut, and this method should be generic enough that it automatically picks them up without needing to change the code.

So I advise to look at the original implementation before your changes, and use that as a reference point for how to do the calculation.

> Edit: see my general comment above, perhaps better to simply revert the changes to this method and then we can take up these changes in new PR's.

tests/Unit/ResourceProductionTest.php
```diff
@@ -118,12 +118,26 @@ public function testFusionPlantEnergyProductionScalesWithDeuterium(): void
             'deuterium_synthesizer_percent' => 10,
             'fusion_plant' => 20,
             'fusion_plant_percent' => 10,
+            'solar_plant' => 30,
+            'solar_plant_percent' => 10,
```
lanedirt, line 122: These lines should be removed as this interferes with the actual test assertion. The test checks for `Fusion plant should produce energy when deuterium production is positive but deuterium storage is 0`

Since you're adding positive solar plant energy to the mix, it's not able to test the fusion plant energy output in isolation.

lanedirt (changes requested):
Thanks so much for all the effort you've put into this PR,  I really appreciate your time and thought that went into these changes. 🙏

That said, I've tried testing the updated logic and while some parts are indeed improved, others seem to have unintended side effects or regressions (e.g. where the existing unit test that is now broken without the added solar plant production change). With multiple core mechanics touched in a single PR, it becomes quite difficult to properly test and validate that everything still works as expected.

> Honestly I don't know the game that much, It make sens to me to not produce energy at all rather than calculate a ratio based on how much deuterium is produced compared to the maximum needed.

Totally understandable. These resource production calculations are tricky and often non-obvious, especially if you're not deeply familiar with the official game mechanics. I strongly recommend playing the original game a bit so you can manually compare certain usecases to get a better sense of expected outcomes.

To help us move forward, I actually would like to suggest we revert the changes made to `PlanetService::updateResourceProductionStatsInner()` back to the original implementation. Then I'll be able to review all other changes in isolation and we can get this PR merged.

Then for the rest of the work to the core resource production calculation, we can try break these improvements into smaller, focused PRs. That way, we can properly review, test, and merge changes incrementally, reducing the risk of breaking other parts of the game.

Let me know what you think! I'm happy to help review follow-up PRs in smaller chunks.

lanedirt:
Closing due to inactivity and in favor of #747 which also addresses some of these issues.

PR #692, opened 2025-07-17
Change laravel mix to vite

jackbayliss:
## Description
 Laravel mix is no longer maintained, so we should use vite to prevent packages being outdated, [see here
](https://laravel.com/docs/12.x/mix)  as far as I'm aware it minifys and versions etc already. 

There are already [some Vulnerabilities](https://github.com/lanedirt/OGameX/security/dependabot) reported, so this fixes those too.

I've ran this locally and it seems correct, but would be good to have an external test.

Below is a full list of changes
 ## Changes 
-  Changed checkout actions to v4 
- Changed the docker yml workflows so they rebuild fresh each time (this was causing me issues as it was grabbing cached versions when edited - maybe a better way?)  
- Included node in php image (so we can build etc right away) 
- Updated the entrypoint.sh so for the app container, it'll npm install and build the assets automatically 
- Updated docs to point at new commands etc 
- Updated  `469500b3cd5158332fb20a56b14b2c.css` as it had some broken css causing issues with Vite which was fun to fix
- Created a ingame outgame js and css files that imports - the same as what mix was doing but contains it in a file for Vite to use 
- Switched mix to vite and links etc.
- I haven't used Laravels vite plugin, because it doesn't play nice with how the project works ie we need our vite scripts to load before the blade scripts, and using laravels plugin / directives mean they're deferred. 

 ## Issues 
 When testing locally, I had the below issues ... hoping to figure it out but a list for anyone interested
- [ ]  Jquery is undefined  
- [ ] Global vars etc dont work, which is due to how vite works.. they're all scoped, so the only way around it from what I can think of is by declaring things like ogame, as window.ogame, so then it's not scoped. 

### Type of Change:
- [X] Other / Improvement

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,8 @@
     "private": true,
     "scripts": {
         "dev": "vite",
-        "build": "vite build",
-        "watch": "vite build --watch"
-    },
+        "build": "vite build -c vite.config.js && vite build -c vite.config-ingame.js && vite build -c vite.config-outgame.js",
+        "watch": "npx concurrently \"vite build --watch -c vite.config.js\" \"vite build --watch -c vite.config-ingame.js\" \"vite build --watch -c vite.config-outgame.js\""    },
     "devDependencies": {
         "axios": "^1.10.0",
         "laravel-vite-plugin": "^2.0.0",
--- a/resources/js/ingame.js
+++ b/resources/js/ingame.js
@@ -1,5 +1,5 @@
 import './ingame/jquery-1.12.4.min';
-import './ingame/jquery.js';
+// import './ingame/jquery.js';
 import './ingame/chat.js';
 import './ingame/inventory.js';
 import './ingame/jquery-spinners.js';
--- a/vite.config-ingame.js
+++ b/vite.config-ingame.js
@@ -1,0 +1,25 @@
+import { defineConfig } from 'vite';
+import path from 'path';
+
+export default defineConfig({
+    build: {
+        minify: false,
+        minifyIdentifiers: false,
+        outDir: 'public',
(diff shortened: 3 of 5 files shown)


lanedirt:
Hi @jackbayliss,

Awesome work, thanks for looking into this! I just tested the latest version locally and can also reproduce the jQuery is not defined error that you mentioned:
image

By the way, I do remember vaguely looking into using vite myself when doing refactoring last year during a new design implementation. I think I also ran into quite a few errors back then, so decided to revert the changes. But it looks like your version is already ahead in terms of issues fixed. 👍

jackbayliss:
@lanedirt Thanks for checking it out, I believe the main issue is because vite is defering the script by default. So the other scripts don't have jQuery etc available right away 

Hoping to play with it and try and figure it out 👀 but feel free to have a go if you want to 😁

lanedirt:
Update as FYI:

After running `docker compose up -d --build` with the new version on my Macbook, I do notice that the `yarn.lock` reports differences, looks to be because it uses platform specific packages during yarn/composer install (?):
image

`package.lock.json` also reports differences:
image

Might not be critical, but I don't think this happened before.

jackbayliss:
Closing as sadly not straight forward. Using Laravel's vite plugin means the js files are deferred, which then means the blade files need those items - I found a way around, but then because the way vite works, the var ogame etc is only local to that file- so when we try to access ogame in the blade that doesn't work.  

I managed to work around it by assigning it to the window, but it just felt wrong.  

Ideally it's possible but we'd need to change all the JS in the views to load via files I assume. May reopen, but hit a wall atm.

Issue #701, opened 2025-08-14
[BUG] Shipyard upgrade queued up as second item blocks Shipyard usage on all planets


Kays:
**Describe the bug**
Queuing up a Shipyard upgrade on ONE planet as the second item in the build queue prevents construction of any items that requires a Shipyard across ALL planets. It seems to behave properly if Shipyard is not the second queued item, for example - third queued item. I don't know why the second place is special. I've tested this multiple times, including logging out, refreshing the page, clearing cache. At first, I wasn't sure why I was blocked because the planet had no Shipyard building and the UI wouldn't let me construct a Solar Satellite.

**To Reproduce**
1. On one planet, queue up one building, then queue up Shipyard.
2. Go to a different planet, and make sure no Shipyard is being upgraded, and try to construct a Solar Satellite.
3. The Build button is grayed out and tooltip says Shipyard is under construction, which it isn't.

**Expected behavior**
Regardless of queue order, Shipyard upgrade from one planet should not block usage in a different planet.

jackbayliss:
This looks to be down to the fact `$player->isBuildingObject('shipyard')` - `isBuildingObject` considers all planets, rather than the current. I guess ideally it should only be the current planet.

PR #704, opened 2025-08-16
Alter isBuildingObject

jackbayliss:
## Description
Closes https://github.com/lanedirt/OGameX/issues/701

- Adjusts the shipyard logic building to take into account the current planet, rather than as the whole player.
- Alters the planet service `isBuildingObject` method level to accept null, which means we'll get the level if not provided

### Type of Change:
- [X] Bug fix

## Checklist

- PlayerService has no tests... so didnt add one. Could we perhaps make an issue for it to cover it more? dunno if its worth it unit test wise..

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -1759,8 +1759,8 @@ class PlanetService
     public function isBuildingObject(string $machine_name, int|null $level = null): bool
     {
         $object = ObjectService::getObjectByMachineName($machine_name);
-        
-        if($level === null) {
+
+        if ($level === null) {
             $level = $this->getObjectLevel($machine_name) + 1;
         }
 
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -1756,9 +1756,13 @@ class PlanetService
      *
      * @return bool
      */
-    public function isBuildingObject(string $machine_name, int $level): bool
+    public function isBuildingObject(string $machine_name, int|null $level = null): bool
     {
         $object = ObjectService::getObjectByMachineName($machine_name);
+        
+        if($level === null) {
+            $level = $this->getObjectLevel($machine_name) + 1;
+        }
 
         // Check only building queue objects
         if ($object->type !== GameObjectType::Building && $object->type !== GameObjectType::Station) {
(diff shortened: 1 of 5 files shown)


Kays:
I've just returned to OGame after being away from the game for many, many years so my knowledge of mechanics is rusty at best, and I don't have an account in OGame to test this - does upgrading a Research Lab in Planet A with IRN active prevent doing research in Planet B? Or would this be fine, but IRN would not use the Research Lab level from Planet A? If so, then this change should accommodate it. Are there any mechanics in the game where upgrading a building in one planet affects the other? I ask because there's a `isBuildingObject()` function in both PlayerService and PlanetService. There would be no need to have two of them if one of them can never happen.

jackbayliss:
@lanedirt may know the answer to that  as I don't know either 😆

lanedirt:
Hi @jackbayliss @Kays,

Thanks for the work on this! I just checked it on one of my accounts on OGame.org, upgrading the research lab building on planet A does indeed prevent planet B from starting a research (with reason: "Research lab is being expanded"). This applies to both having IRN and not having IRN.

So with this in mind, my guess is the best way to fix #701 would be to change the parts in the code that explicitly check for shipyard being upgraded, like:
```
     if (!$player->isBuildingObject('shipyard')) {
```

And we change these explicit checks to use the `$planet` context instead of `$planet`, so it only checks if the current planet is building a shipyard. Then for the research_lab checks, we can continue to use the (old) $player implementation. So concrete:
- Shipyard being upgraded **(on current planet)** --> use $planet->isBuildingObject()
- Research lab being upgraded **(on any planet)** --> use $player-->isBuildingObject()?

Thoughts?

app/Services/PlanetService.php
```diff
@@ -1756,10 +1756,14 @@ public function isBuilding(): bool
      *
      * @return bool
      */
-    public function isBuildingObject(string $machine_name, int $level): bool
+    public function isBuildingObject(string $machine_name, int|null $level = null): bool
     {
         $object = ObjectService::getObjectByMachineName($machine_name);
 
+        if ($level === null) {
```
jackbayliss, line 1763: This just simplifies what we were doing elsewhere- so rather than needing to do this logic everywhere it encapsulates it within this method.

jackbayliss:
@lanedirt  Great idea, makes sense - thanks for checking it out as I wasn't sure what the implementation was like in the real game 🤓

lanedirt:
Looking good!

PR #747, opened 2025-09-23
Add ProductionIndex and convert production logic and resource settings view to accommodate correct displays

### LINKED ISSUE #708: [BUG] Planet's slot income bonus is only applied on basic income
**Describe the bug**
From what I can glance at, it looks like the planet position bonus income only applies to basic income and not to the overall mine generation. If it's supposed to be on basic income only, the bonus is useless because it's applied to a very small number.

This grabs the basic income, multiply by universe speed
https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1474

... then it gets multiplied by slot bonus here
https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1512

Looking at mine generation, starting with `getObjectProduction()`, the bonus is not applied:
https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1627

We don't want to have to factor this multiplier into each GameObject, so I assume this logic has to be higher up like in `getObjectProduction()` function or its callees. Looking at one of them in `updateResourceProductionStatsInner()`:
https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1564

... does not look like it gets applied there either. In-game, looking at table charts between a planet in position 1, 6, and 15 show the same numbers for all mines.

I'd like to start a convo on what the solution should be. I'm not seeing a clear path on the best place to make this change with the current structure. There is this other issue for tech bonus - https://github.com/lanedirt/OGameX/issues/170 - which involves Plasma Technology. That is an update that can be applied to the GameObjects for mine buildings as inline eval code, but then we need to decouple the value when presented in the "Resource Settings" screen in-game. Then we need to factor in this slot bonus multiplier, and maybe commander multipliers, class multipliers, alliance multipliers, etc. in the future. Maybe a better design is to use a Resource-like object that will calculate the resulting value for each bonus with props like basic_income, mine_production, plasma_tech_bonus, planet_slot_bonus, etc. and each prop will have its own value calculated and can be `sum()` to get the final value, but the individual items will help with the table breakdown layout design in "Resource Settings" screen as well.

Short term, at least related to this ticket's topic, we could add a one-liner in `getObjectProduction()` (https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1666) before the return, and that should be okay, and can be fine until a better structure is used, which would take more time to engineer.
```php
$production = $this->calculatePlanetProductionBonuses($production);
```

In terms of calculation, I assume each bonus works like this:
```
total_production = basic_income + mine_production
plasma_tech_bonus = total_production * %
slot_bonus = total_production * %
officer_bonus = total_production * %
etc.
```

In other words, the bonuses do not get compounded by other multipliers. If that is the case, it may be trickier to get the order correct so they can be displayed properly in the "Resource Settings" screen.

Kays:
## Description
@lanedirt  can use your review and feedback. I don't know my way around this framework and I've never used it before so I'm still trying to figure out how to do some of the checklist items, particularly the unit/feature tests part. I don't understand the error messages that it's showing me. I welcome help and advice on that. For now, just the code and logic bits.

Note: this MR does not fix the Techinfo panel for mines/buildings. The sum() on Resources object still factors in the energy component, when it shouldn't.

Edit - some issues:
- I realized I left the `has*` officer functions in PlayerService to `return true`. I had it on to test and forgot to set it back to false. I will update that in a follow-up commit later.

### Type of Change:
- [x] Bug fix
- [x] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #708 

## Checklist

- [x] **Testing:**

## Additional Information

### DIFF
--- a/resources/views/ingame/resources/settings.blade.php
+++ b/resources/views/ingame/resources/settings.blade.php
@@ -82,7 +82,7 @@
                                     {{ $row['production']->crystal->getFormattedLong() }}
                                 </span>
                                 </td>
-                                <td class="{{ $row['production']->deuterium->get() > 0 ? 'overmark' : ($row['production']->deuterium->get() < 0 ? 'undermark' : 'normalmark') }}">
+                                <td class="{{ $row['production']->deuterium->get() > 0 ? 'undermark' : ($row['production']->deuterium->get() < 0 ? 'overmark' : 'normalmark') }}">
                                 <span class="tooltipCustom "
                                       title="{{ $row['production']->deuterium->getFormattedFull() }}">
                                     {{ $row['production']->deuterium->getFormattedLong() }}
@@ -140,17 +140,17 @@
                                 </td>
                                 <td>
                                 </td>
-                                <td class="{{ $row['production']->metal->get() > 0 ? 'overmark' : ($row['production']->metal->get() < 0 ? 'undermark' : 'normalmark') }}">
+                                <td class="{{ $row['production']->metal->get() > 0 ? 'undermark' : ($row['production']->metal->get() < 0 ? 'overmark' : 'normalmark') }}">
                                 <span class="tooltipCustom " title="{{ $row['production']->metal->getFormattedFull() }}">
                                     {{ $row['production']->metal->getFormatted() }}
                                 </span>
                                 </td>
-                                <td class="{{ $row['production']->crystal->get() > 0 ? 'overmark' : ($row['production']->crystal->get() < 0 ? 'undermark' : 'normalmark') }}">
+                                <td class="{{ $row['production']->crystal->get() > 0 ? 'undermark' : ($row['production']->crystal->get() < 0 ? 'overmark' : 'normalmark') }}">
                                 <span class="tooltipCustom " title="{{ $row['production']->crystal->getFormattedFull() }}">
                                     {{ $row['production']->crystal->getFormatted() }}
                                 </span>
                                 </td>
-                                <td class="{{ $row['production']->deuterium->get() > 0 ? 'overmark' : ($row['production']->deuterium->get() < 0 ? 'undermark' : 'normalmark') }}">
+                                <td class="{{ $row['production']->deuterium->get() > 0 ? 'undermark' : ($row['production']->deuterium->get() < 0 ? 'overmark' : 'normalmark') }}">
                                 <span class="tooltipCustom "
                                       title="{{ $row['production']->deuterium->getFormattedFull() }}">
                                     {{ $row['production']->deuterium->getFormattedLong() }}


app/Http/Controllers/ResourcesController.php
```diff
@@ -105,61 +108,54 @@ public function settings(Request $request, PlayerService $player): View
...
-
         $production_factor = $this->planet->getResourceProductionFactor();
 
         return view('ingame.resources.settings')->with([
             'basic_income' => $this->planet->getPlanetBasicIncome(),
             'planet_name' => $this->planet->getPlanetName(),
             'building_resource_rows' => $building_resource_rows,
             'building_energy_rows' => $building_energy_rows,
-            'production_total' => $production_total,
+            'production_total' => $productionIndexTotal,
```
lanedirt: Can you refactor the naming style of this variable so it matches the existing surrounding code, in this case make it snake_case: `$production_index_total`

app/Models/ProductionIndex.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace OGame\Models;
+
+class ProductionIndex
+{
+    // basic income
```
lanedirt: Multiline comment style is preferred for public variables and methods which should include the type, so like the following:

```
 /**
  * Basic income
  *
  * @var Resources
  */
```

Applies to all the public vars in this files.

app/Models/Resource.php
```diff
@@ -58,31 +58,34 @@ public function getRounded(): int
...
      *
+     * @param int $multiplier
      * @return string
      */
-    public function getFormatted(): string
+    public function getFormatted(int $multiplier = 1): string
     {
-        return AppUtil::formatNumberShort($this->rawValue);
+        return AppUtil::formatNumberShort($this->rawValue * $multiplier);
     }
```
lanedirt: Making the `getFormatted()` handle both formatting and multiplication makes it not that readable IMO. Also, there already is an existing method in the `Resource.php` class for multiplication:

```
    /**
     * Multiply resource by a factor.
     *
     * @param int $factor
     * @return void
     */
    public function multiply(int $factor): void
    {
        $this->rawValue = $this->rawValue * $factor;
    }
```

So instead of doing this:
```
$production_total->total->deuterium->getFormattedLong(168)
```

You can do it like this:
```
$production_total->total->deuterium->multiply(168)->getFormattedLong()
```

**Edit:** I see that the multiply() method returns void. So we might need to change the multiply method so it returns itself (`Resource` type) so calls can be properly chained.

app/Services/PlanetService.php
```diff
@@ -1459,7 +1461,6 @@ public function updateResourceProductionStats(bool $save_planet = true): void
         // 1. First time in order to ensure the energy production is up-to-date.
         // 2. Second time for the mine production to be updated according to the up-to-date (actual) energy production.
         $this->updateResourceProductionStatsInner($production_total, $energy_production_total, $energy_consumption_total);
-        $this->updateResourceProductionStatsInner($production_total, $energy_production_total, $energy_consumption_total);
```
lanedirt, line 1462: If removing the second call is on purpose, please update the comment above to explain the new flow.

lanedirt:
Hi @Kays,

Thanks for looking into this and your PR! I added some code style comments, see above. I also see that the Laravel Pint Code Style Checker in general is failing. My advice would be to first look into this and update the code to match the prescribed style guidelines.

I also notice that some tests are failing, looks like this has to do with the resource calculation that is not returning the values that the existing tests expect. I can help look into this, but will probably take me a couple of days to make some time for that.

So if you could look into the comments first, then afterwards I'll test the feature locally and check the failing tests. 😄

app/Http/Controllers/ResourcesController.php
Kays: How about `$productionIndex_total`? Looks weird so I did `$productionIndexTotal`. `ProductionIndex` is refers to the object type, so it's like a proper noun. Or maybe `$productionindex_total`?

app/Models/Resource.php
Kays: In `resources/views/ingame/resources/settings.blade.php`, it has usage like `->getFormatted(168)`. That was already there. It wasn't part of my change. But the call doesn't work properly because the function doesn't take any arguments. I agree with your assessment that it looks awkward and `->multiply()` would be better. In this situation, should `->multiply()` be a mutable function? I can see arguments to both. I know Resources's multiply is a mutation. But if I'm only needing it for formatting in a specific instance, that would not be good. I would have to clone the object for every use.

Edit: I can change to using AppUtil directly for formatting and pass over something like `->metal->get() * 168`.

app/Services/PlanetService.php
Kays: Will do. It is on purpose. There should not be a need to calculate it twice. The production factor can be calculated on the spot - https://github.com/lanedirt/OGameX/pull/747/files#diff-24965fd880ba30948f1ea3baf4502c554c8aa8a831e6b8ea594141cbf98617b5R1611

Then multiplying that to the total production should be the same thing. There is a weird cyclic dependency structure in the current setup where calculating the energy metrics for production factor calls `getObjectProduction()` but `getObjectProduction()` needs the energy metric to provide the correct values, which hasn't been calculated yet.

So I think the correct way is to include this change, which I haven't added to this PR yet - https://github.com/Kays/OGameX/commit/a47b78a2f8397d449b7529416fdb1f6c55e80f64
Edit: added the commit into this PR now

My thinking is that `$force_factor` was incorrectly used in `$building_percentage`. Building percentage is set by the user. That's a fixed value and I can't think of a scenario where we want to alter that value for any calculations. Applying it to the production factor makes sense, particularly in the event the production factor becomes 0. This would be a huge problem because that forces all values to be 0, permanently. So the logic is to force production factor to 1, do the calculations, then calculate the new production factor by energy consumption, then multiply it to the production values.

I'm open to any loopholes or flaws in this logic that I'm not seeing.

Kays:
@lanedirt fixed all the formatting stuff and a few bug fixes as well. Added comments for the other stuff. For some reason, the checks are still failing but I don't know why.

On a side note, I wish I could run these tests locally, which I've tried, and they messed up my entire instance:
- changing my server settings
- adding a bunch of users and planets

app/Http/Controllers/ResourcesController.php
lanedirt: `$productionindex_total` also looks fine to me in snake_case 👍. That makes it consistent with the other code in the function.

app/Models/Resource.php
lanedirt: > Edit: I can change to using AppUtil directly for formatting and pass over something like ->metal->get() * 168.

Yes I agree, I also think that would be the best option for now. Because indeed, when using `multiply()` it would mutate the object itself that is used in multiple places, requiring to have unique clones for each display making it a bit too verbose IMO. So using AppUtil directly should be fine here. 👍

lanedirt:
Thanks for the fixes you've applied, looking good! I added additional feedback to the remaining open points.

> On a side note, I wish I could run these tests locally, which I've tried, and they messed up my entire instance:
> 
> * changing my server settings
> * adding a bunch of users and planets

Yes, running the `php artisan test` locally will change/reset server settings and add test users and planets. This is because the feature tests run against the real (local) database for proper end-to-end testing without mocks.

If you’re also running a local game instance that you actually play on, I recommend setting up a separate environment just for development/testing. An easy way is to clone the project into a second folder, run it on different ports, and dedicate that instance to testing. That way your “play” environment stays untouched.

Being able to freely run the tests locally is crucial for being able to test your changes locally and ensure there are no regressions.

Kays:
@lanedirt  fixed all the GitHub Actions errors. For some reason, PHPStan is stuck and I don't know why. I will fix the variable name tomorrow, but the PR should be good for you to test locally now. You can enable the officers to see the change reflected in the Resource Settings. Unfortunately, they can't stay enabled due to failing solar sats unit test due to officers providing a bonus that's unaccounted for in the test.

As for the dev env, I do have a separate env that I use for testing, but it still clutters the instance with hundreds of dummy users, planets, etc. May need to work out some sort of regular db backports, but that's equally as annoying. So I'll just defer to GitHub Actions for now. Would be nice if there's a way to run the tests in a separate container that gets removed so the actual instance won't grow with clutter after each test run. Can map the same codebase, but with the same dummy test db setup that GitHub Actions use. Doesn't sound like it should be hard.

app/GameObjects/Models/Fields/GameObjectProduction.php
```diff
@@ -2,10 +2,209 @@
...
+    public ?\Closure $metal_formula = null;
+    public ?\Closure $crystal_formula = null;
+    public ?\Closure $deuterium_formula = null;
+    public ?\Closure $energy_formula = null;
+
+    public PlanetService $planetService;
+    public PlayerService $playerService;
+
+    public int $universe_speed = 1;
+    public int $plasma_technology_level = 0;
```
lanedirt: Could you please add comments to all public vars and methods which include params/types and return type descriptions. You can check the `PlayerService.php` for examples. This will improve maintainability in the long run.

app/Http/Controllers/ResourcesController.php
```diff
@@ -105,61 +107,54 @@ public function settings(Request $request, PlayerService $player): View
...
 
         // Buildings that provide resource income
         // Get all buildings that have production values.
         foreach (ObjectService::getGameObjectsWithProduction() as $building) {
-            // Retrieve all buildings that have production values.
-            $production = $this->planet->getObjectProduction($building->machine_name);
-            $production_total->add($production);
+            $level = $this->planet->getObjectLevel($building->machine_name);
+            $productionIndex = $this->planet->getObjectProductionIndex($building, $level);
+            $productionindex_total->add($productionIndex);
```
lanedirt: Can you change the `$productionIndex` variable to also follow the snake_case convention in this function, so make this `$productionindex`.

app/Services/PlanetService.php
```diff
@@ -1587,25 +1590,62 @@ private function updateResourceProductionStatsInner(Resources $production_total,
...
 
+        $this->planet->energy_used = (int) $energy_consumption_total;
+        $this->planet->energy_max  = (int) $energy_production_total;
+
+        $production_factor = $this->getResourceProductionFactor() / 100;
+
+        // i don't know if modifying energy will affect anything, but to be safe
+        $total_energy = $building_production_total->energy->get();
+        $building_production_total->multiply($production_factor);
+        $building_production_total->energy->set($total_energy);
```
lanedirt: Can you rewrite this comment to make it more clear what the purpose is. The `i don't know if` can be confusing to the reader. :)

app/Services/PlayerService.php
```diff
@@ -781,4 +781,47 @@ public function isBuildingObject(string $machine_name): bool
...
+    {
+        // TODO: don't know if this is correct logic
+        //       like if this has to be purchased specifically
+        //       and buying all officers individually won't work
+        //       I don't know
+        return $this->hasCommander()
+            && $this->hasAdmiral()
+            && $this->hasEngineer()
+            && $this->hasGeologist()
+            && $this->hasTechnocrat();
```
lanedirt: I just checked the original game implementation. And yes, whether all officers have been activated individually or as a package, it should both result in `hasCommandingStaff()` to be `true`. So you can remove the comment.

lanedirt (changes requested):
Hi @Kays,

Thanks for the fixes! I did some testing with your version locally and the resource production numbers seem to make way more sense now. Really great job!

I added some style comments, see above. Could you look into these?

As well I noticed a potential issue with the plasma tech calculation. For example, according to the plasma tech description it should give a 1% bonus to metal per level:

```
A further development of ion technology which accelerates high-energy plasma, which then inflicts devastating damage and additionally optimises the production of metal, crystal and deuterium (1%/0.66%/0.33% per level).
```

On the official game, this seems to work as expected:
Screenshot 2025-09-27 at 09 42 36
- Metal mine production: 226.152
- Plasma tech level 7 bonus: 226.152 * 0,07 = 15.830 

So here the plasma bonus value matches what I expect.

---

However on OGameX this is what happens for me locally:
image

- Metal mine production: 39.719
- Plasma tech level 10 bonus: 39.719 * 0.10 = 3.971

However as you can see in the screenshot it comes down to 3.229 which does not match.

---

I checked the code, and I'm not sure if this might have to do with the `planet_slot` bonus that seems to be calculated independently (while in the official game it looks to be part of the mine output itself?), or something else.

Also please check out the `ResourceProductionTest.php` which already contains some assertions on resource production output. The plasma tech bonus calculation might be a good candidate to add a unit test for as well. You can run unit tests locally (without affecting the database) via:

```bash
php artisan test tests/Unit
```

Kays:
@lanedirt Good catch on the Plasma Technology issue. Bit of an awkward interaction.

app/GameObjects/Models/Fields/GameObjectProduction.php
Kays: Please pardon my lack of experience with this framework - what do the comments do, particularly the params/types and return type in comment block? The class properties are type-hinted, and all function parameters and returns are type-hinted.

app/Http/Controllers/ResourcesController.php
Kays: Being honest here, and with all due respect - it doesn't make sense to me to do it, so I will not, but I don't mind if you would like to change them. I don't like my renaming of the variable to `$productionindex_total` earlier either.

I referenced and tried to name my variables in convention to what I've seen in the codebase. For example:
- In `BuildingObjects`, all the vars of building names are in camelCase - `$metalMine`, `$crystalMine`, etc.
- In `PlanetService`, the `GameObject` variable is `$gameObject`.

As for the sticking to conventions in the function, referencing the `index` function in the same file, `ResourcesController`, there is this block:
```php
$unitQueue = resolve(UnitQueueService::class);
$unit_full_queue = $unitQueue->retrieveQueue($this->planet);
```

I didn't write those, I was just following those conventions. If there's a new convention, I don't mind, but I'd like a better way to determine them other than you pointing out each case.

As for snake_case, it looks like they're used for non-proper nouns, which I followed for my additions in `GameObjectProduction` as well as other variable names.

app/GameObjects/Models/Fields/GameObjectProduction.php
lanedirt: Hi @Kays,

PHP Docblocks for all public methods are preferred because it helps static analysis & IDE Support. Tools like PHPStan and IDEs can interpret PHPdoc for more advanced checks that normal PHP types cannot always do. It will also allow to generate API level documentation later if needed in the future.

Having a uniform style across the project makes the codebase easier to read and maintain for all contributors.

app/Http/Controllers/ResourcesController.php
lanedirt: Hi @Kays, to keep the codebase readable, the rule is: when adding code to an existing method, follow the conventions already used in that method. This isn’t about personal preference but about consistency. So yes, it can happen that one method in a class uses a different convention than another (not preferred, but this might have grown historically). The goal is simply to stay as consistent as possible within the method you’re working on, especially when it’s pointed out in a PR review.

I know this might feel like a small and even frustrating change, but consistency helps a lot with long-term maintainability. So I kindly ask you again to please make this adjustment. If not, I’ll need to do it myself before merging, but I’d prefer if you do it as the contributor.

And I agree that conventions can be made more clear. But it’s a come-and-go process, and each PR review helps clarify and move closer toward a more consistent standard across the codebase. I have just now updated the CONTRIBUTING.md with this guideline: https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md#1-follow-existing-conventions-in-the-code-youre-working-with. 

Thanks again for your work on this!

Kays: Yes; I understand the concern with consistency and as I've explained, I try to keep in line with that based on what I've seen. Also, I don't mind if the style is your personal preference or not. My qualm was that if I'm making changes and I don't fully know what the rules are, and referencing other existing code in the code base is not a proper indicator, then I don't know if what I'm doing is consistent or not and that is more frustrating. Of course, I don't want my MRs to repeatedly go back-and-forth on variable styling. I will comb through my changes and try to conform to that style guide. I rather do one sweep than an ad-hoc and on-demand style, correcting them as they're reported.

So with my changes in `GameObjectProduction`, per the guideline, should I change my class properties like `metal_formula` to `metalFormula` or change the instances of `playerService` to `playerservice` or `player_service`?

Thank you for that update to the guidelines!

app/GameObjects/Models/Fields/GameObjectProduction.php
Kays: Understood. `protected` and `private` vars and functions do not require them?

Kays:
I know the official OGame does not have the planet position row in Resource Settings but I think it's a useful breakdown item. Reduces some confusion players may have when comparing mine outputs between planets with position bonus and planets without position bonus. Thoughts?

app/Http/Controllers/ResourcesController.php
lanedirt: Thanks for your understanding, much appreciated!

For `GameObjectProduction`, IMO you can leave it as-is as it already contained mix styles so there's no obvious right and wrong there. But if you want to refactor, then feel free to make all class variables `camelCase` for consistency as that's used in most places already and also used by Laravel itself. 

Again I agree, standards should be more explicitly stated somewhere (which we can address in a different PR). For now when changing code that already uses both styles mixed, either way is fine. But if you make changes to a limited scope that already explictly uses one style: adhere to that style.

Kays:
@lanedirt Ready for re-review. I think I've addressed everything in the comments. I added a unit test to check for plasma tech and planet slot. Not sure if what I wrote is in the proper unit test format. Let me know what I should do if it's not. Thank you!

Side note: while calculating stuff for unit tests, I noticed that `PlanetService::getPlanetBasicIncome()` multiplies basic energy production by universe speed. Is that intended? Object prices don't change, and other building energy production and solar sat energy production are unaffected by universe speed. It's outside the scope of this MR, but throwing it out there.

lanedirt:
Hi @Kays,

Thanks a lot for the fixes! I just tested the calculations on my local instance and everything seems to match up perfectly now. Also thanks for adding the unit tests, general code structure looks good to me! One small comment on the resource settings screen below, other than that, LGTM! 👍 

> I know the official OGame does not have the planet position row in Resource Settings but I think it's a useful breakdown item. Reduces some confusion players may have when comparing mine outputs between planets with position bonus and planets without position bonus. Thoughts?

I agree that adding the planet position bonus separately would be in line with the other bonus displays, _however_ the goal for OGameX is to be a 100% 1-on-1 clone with the official game as much as possible, which includes replicating any game mechanics that might be considered unoptimal (on purpose). 

So as the official game includes the planet bonus in the normal mine output, could you please tweak the `resources/views/ingame/resources/settings.blade.php` file and merge the planet position bonus with the mine output itself: 
image

> Side note: while calculating stuff for unit tests, I noticed that PlanetService::getPlanetBasicIncome() multiplies basic energy production by universe speed. Is that intended? Object prices don't change, and other building energy production and solar sat energy production are unaffected by universe speed. It's outside the scope of this MR, but throwing it out there.

Good point! By default the planet basic income does not contain energy, however the server admin can change that of course. I agree that tweaking the logic so that the `energy` production is not affected by the universe speed would make sense for basic income. I think it would be good to take this up in a different PR to avoid scope creep in this one, but feel free to create an issue and/or PR for this.

Thanks again!

Kays:
@lanedirt Just my opinion, but I think it would be nice if OGameX has some artistic flair while keeping the core mechanics the same. For example, being able to click on a DF to send Recyclers. We can click on a planet to send Espionage Probe in the official game, yet can't click on a DF to mine it when it's the only possible action. I feel like that's a mechanic OGameX could add, yet would not alter the game whatsoever other than quality-of-life improvement. OGame already exists so if OGameX is an exact clone, just play the official version. Nevertheless, I will respect and honor your request.

As for these tests. I get the feeling that something weird is going on. It's like they can fail randomly. I've ran into this quite a bit in this one PR and I really curious. What in my two commits that I just pushed out that have any impact on why an Expedition test case would fail? All tests were passing fine before.

lanedirt:
Hi @Kays,

Awesome, LGTM! 👍 Thanks for your work and persistence in addressing all the comments, much appreciated! Merging!

As for the tests, yes indeed some of are a bit flaky. Especially with the combat simulation as the outcome of that is not deterministic. So some test assertions might be a bit too specific in hindsight, needs to be looked at. But I've just reran them and now it works fine.

For your idea regarding UI and other improvements: there has been a similar idea proposed before in:
- #113 

This calls for an optional modules system, which is something I would support for the OGameX project: having the base game still be a 1-to-1 exact replica, but allowing server admins to optionally enable additional modules that can change/expand/improve game mechanics. This would allow UX improvements as you suggest, but also allow more complicated features like e.g. bots to be enabled for solo offline play etc.

lanedirt (approved):

Issue #231, opened 2024-06-17
[BUG] Base speed of ships that have upgraded drives should be different


lanedirt:
**Describe the bug**
Currently all ships have a static base speed. E.g. "small cargo" now has a base speed of 5.000. However when the drives get upgraded when impulse drive reaches level 5, this base speed should be changed to 10.000.

The upgraded drive logic is already supported for bonus percentages to base speed, but it doesn't change the base speed itself yet.

See https://ogame.fandom.com/wiki/Base_Speed with a full overview. This changed base speed should be supported for the following objects:

- Small cargo
- Recycler
- Bomber

TODO:
- Add option to configure different base speeds for different drive upgrades.

PR #753, opened 2025-10-11
Update SpeedPropertyService.php

Geda173:
Fix: honor speed_upgrade thresholds (e.g. Small Cargo switches to Impulse 5), previously the base speed of the Impulse Drive was not applied.

## Description
This pull request fixes an issue where ships that define a speed_upgrade threshold (such as the Small Cargo) did not properly switch to the higher drive’s base speed when the player reached the required research level.

Previously, even after reaching Impulse Drive 5, the Small Cargo continued to use Combustion Drive’s 10% per level speed multiplier instead of upgrading to Impulse Drive’s 20% per level.

### Type of Change:
- [x] Bug fix
- [X] Feature enhancement

## Related Issues
Fixes  [#231](https://github.com/lanedirt/OGameX/issues/231) (partially)

## Checklist

- [X] Code Standards: Verified with Laravel Pint (PSR-12 compliant).
- [X] Static Analysis: Passes PHPStan static analysis.
Testing:
- [X] Small Cargo speed now increases according to Impulse Drive at level 5+.
- [X]  Behavior of all other ships unchanged.
- [X] No regressions in fleet speed calculations.
- [X] CSS & JS Build: N/A – no frontend changes.
- [X] Documentation: In-code docblocks updated.

## Additional Information
What this PR does:
	•	Updates SpeedPropertyService::getBonusPercentage() to correctly:
	•	Detect and apply the appropriate drive technology when the player meets or exceeds its speed_upgrade threshold.
	•	Prioritize drives in logical order (Hyperspace > Impulse > Combustion).
	•	Fall back to the ship’s base drive requirement if no upgrade thresholds are met.
	•	Maintains full backward compatibility for all other ships and logic.

The old logic looped upgrades but failed to properly “override” when multiple drives existed. The new implementation uses an ordered priority check and returns the first qualifying drive bonus.

### DIFF
--- a/tests/Feature/FleetDispatch/FleetDispatchGenericTest.php
+++ b/tests/Feature/FleetDispatch/FleetDispatchGenericTest.php
@@ -106,7 +106,7 @@ class FleetDispatchGenericTest extends FleetDispatchTestCase
         // With Impulse ≥5, Small Cargo switches to Impulse:
         // base speed becomes 10,000 and at Impulse 10 the bonus is +200% (20% × 10) → effective speed 20,000.
         // This reduces the duration to 56m 28s (3388 seconds) for a 1-system jump at fleet_speed=1.
-        $this->assertEquals(3388, $fleetMissionService->calculateFleetMissionDuration($this->planetService, $targetPlanetCoords, $units));
+        $this->assertEquals(4788, $fleetMissionService->calculateFleetMissionDuration($this->planetService, $targetPlanetCoords, $units));
     }
 
     public function testFleetDeuteriumConsumptionCalculation(): void


Kays:
I can't reproduce the premise. The original code works fine for me - Small Cargo swaps over to Impulse Drive at level 5 correctly, and the bonus calculation is correct.

> The old logic looped upgrades but failed to properly “override” when multiple drives existed.

The only ship that has multiple drive upgrades is the Recycler - Impulse at level 17, Hyperspace at level 15. And I've tested that one, and the original code works. I suppose the way it would fail is if the `speed_upgrade` entries were ordered incorrectly and Hyperspace is listed above Impulse, but that's not how the code is written.

As for the code, it looks like a neat refactor, but I don't see any logical change other than processing by priority drives, but considering almost all ships have at most one drive upgrades, such a change doesn't feel necessary.

Geda173:
> I can't reproduce the premise. The original code works fine for me - Small Cargo swaps over to Impulse Drive at level 5 correctly, and the bonus calculation is correct.
> 
> > The old logic looped upgrades but failed to properly “override” when multiple drives existed.
> 
> The only ship that has multiple drive upgrades is the Recycler - Impulse at level 17, Hyperspace at level 15. And I've tested that one, and the original code works. I suppose the way it would fail is if the `speed_upgrade` entries were ordered incorrectly and Hyperspace is listed above Impulse, but that's not how the code is written.
> 
> As for the code, it looks like a neat refactor, but I don't see any logical change other than processing by priority drives.

Thanks for checking! You’re absolutely right that the drive bonus percentage already switches correctly at Impulse 5, I confirmed that too after retesting.

What I ran into (and what this patch was meant to address) is the base-speed change that occurs at that same threshold. In the current logic, the Impulse 5 switch applies the 20 % bonus correctly, but it still uses the Combustion base speed (5 000).

In the official OGame behavior, the Small Cargo’s base speed doubles to 10 000 once it gets its Impulse drive. The same principle applies later for Recyclers and Bombers, just like the issue was raised in Bug #231.

The change in this PR doesn’t alter that base-speed behavior yet - it was aimed at clarifying and making the upgrade logic more explicit, but my testing led me to realize that the underlying code doesn’t currently support different base speeds for ships that refit to new drives (Small Cargo, Recycler, Bomber). A follow-up RP to adress this entirely is needed.

Kays:
> What I ran into (and what this patch was meant to address) is the base-speed change that occurs at that same threshold. In the current logic, the Impulse 5 switch applies the 20 % bonus correctly, but it still uses the Combustion base speed (5 000).
> 
> In the official OGame behavior, the Small Cargo’s base speed doubles to 10 000 once it gets its Impulse drive. The same principle applies later for Recyclers and Bombers, just like the issue was raised in Bug #231.
> 
> So this fix wasn’t about the bonus percentage order, but about enabling ships to have different base speeds once a drive upgrade threshold is met. The new base_speed field in GameObjectSpeedUpgrade simply makes that configurable.

The only change I see in this PR is to `SpeedPropertyService` and that only calculates the bonus, which is independent of the base speed of the ship. Are there changes missing from this PR?

Kays:
> The change in this PR doesn’t alter that base-speed behavior yet - it was aimed at clarifying and making the upgrade logic more explicit, but my testing led me to realize that the underlying code doesn’t currently support different base speeds for ships that refit to new drives (Small Cargo, Recycler, Bomber). A follow-up RP to adress this entirely is needed.

I think the change for this is relatively simple.

1. Remove the `$base_value` property from `ObjectPropertyService`, both from the class and the constructor. There is no need for this because it has `$parent_object` and we can look up the values needed from there. This requires you to remove the same parameter from all the other `*PropertyService` classes - `GameObjectProperties`, `AttackPropertyService`, `ShieldPropertyService`, etc.
2. In the same class `ObjectPropertyService`, in `calculateProperty` function, reference the base value as such, and replace all instances of `$this->base_value` with `$baseValue`.
```php
$baseValue = $this->parent_object->properties->{$this->propertyName}->rawValue;
```
3. In `SpeedPropertyService`, you can just set `$object->properties->speed->rawValue = <new-value>;`.

That should do it. As a side effect, speed upgrades affect fuel costs. Small Cargo's fuel cost goes from 10 to 20 when fitted with Impulse Drive. We may want to account for that somehow.

Geda173:
> > The change in this PR doesn’t alter that base-speed behavior yet - it was aimed at clarifying and making the upgrade logic more explicit, but my testing led me to realize that the underlying code doesn’t currently support different base speeds for ships that refit to new drives (Small Cargo, Recycler, Bomber). A follow-up RP to adress this entirely is needed.
> 
> I think the change for this is relatively simple.
> 
> 1. Remove the `$base_value` property from `ObjectPropertyService`, both from the class and the constructor. There is no need for this because it has `$parent_object` and we can look up the values needed from there. This requires you to remove the same parameter from all the other `*PropertyService` classes - `GameObjectProperties`, `AttackPropertyService`, `ShieldPropertyService`, etc.
> 2. In the same class `ObjectPropertyService`, in `calculateProperty` function, reference the base value as such, and replace all instances of `$this->base_value` with `$baseValue`.
> 
> ```
> $baseValue = $this->parent_object->properties->{$this->propertyName}->rawValue;
> ```
> 
> 3. In `SpeedPropertyService`, you can just set `$object->properties->speed->rawValue = <new-value>;`.
> 
> That should do it. As a side effect, speed upgrades affect fuel costs. Small Cargo's fuel cost goes from 10 to 20 when fitted with Impulse Drive. We may want to account for that somehow.

Thanks for the clarification! I haven’t done the `$base_value` cleanup yet. I kept this PR focused on the drive refit and base speed changes to avoid blowing up the scope too much. The refactor to resolve values directly from `$parent_object` definitely makes sense as a next step once this is merged.

I’d also prefer to handle the fuel cost adjustments in a separate PR. It’s closely related to the drive upgrades but touches a few other systems, so I’d rather keep that isolated and test it properly later.

Right now, this PR just finalizes the base speed upgrade behavior and leaves things ready for the next round of improvements.

PR #769, opened 2025-10-15
Deathstar expedition exclusion

Geda173:
## Description
PR to exclude `death_star` from the `$expeditionLevels` list, as per @Kays's suggestion. Also removed `solar_satellite` from `$nonFindableShips`.

### DIFF
--- a/app/GameMissions/ExpeditionMission.php
+++ b/app/GameMissions/ExpeditionMission.php
@@ -306,11 +306,9 @@ class ExpeditionMission extends GameMission
         $fleetUnits = $this->fleetMissionService->getFleetUnits(mission: $mission);
         $objectService = app(ObjectService::class);
 
-        // Define expedition hierarchy levels. These determine the *ceiling* for findable ships.
-        // Some ships (e.g. deathstar, colony/recycler/solar) are included for hierarchy but excluded from finds below.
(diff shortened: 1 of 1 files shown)


Kays:
@Geda173  nice work!

A few of the comments added in your last PR are inaccurate. As far as I'm aware, there are no ships that partake in the hierarchy that cannot be "found". So these comments are inaccurate, moreover, `$nonFindableShips` is unnecessary because only ships listed in `$expeditionLevels` can be found and Deathstar, Colony Ship, Recycler, and Solar Satellite (for technicality) are excluded.
```php
        // Some ships (e.g. deathstar, colony/recycler/solar) are included for hierarchy but excluded from finds below.
```
```php
        // Exclude ships that should never be found (even if they affect hierarchy)
        // IMPORTANT: These ships can affect the hierarchy ceiling but MUST never be found as expedition rewards.
```

Geda173:
Hey, looks like there are some remains from previous PR.

>         // Some ships (e.g. deathstar, colony/recycler/solar) are included for hierarchy but excluded from finds below.
This was previously true because `death_star` was part of the hierarchy. The other ships were included to reflect their "unfindability" but I guess that is a bit obtuse. I'll modify them momentarily.

lanedirt:
Thanks @Geda173 and @Kays for the update, merging!

PR #772, opened 2025-10-17
Fix: Ensure at least one ship is always awarded on expeditions (Armada Fix)

### LINKED ISSUE #768: [BUG] Expedition “Armada” event sometimes yields no ships (legacy behavior)
**Describe the bug**
When running expeditions, players sometimes receive a “We found the remains of an armada” message — but no ships are actually added to their fleet.
The message implies a successful “Gain Ships” event, yet the ship list is empty.

**To Reproduce**
1.	Send a small expedition (e.g. 1 cruiser + 10–15 large cargos).
2.	Wait for a “Gain Ships” expedition event.
3.	Occasionally, the result message will show the “armada” text but list no ships.

**Observed behavior**
- Message class ExpeditionGainShips is triggered.
- However, the generated UnitCollection is empty, producing a blank ship list.
- This happens when the resource budget rolled is lower than the cost of any ship randomly selected for recovery and thus every amount rounds down to 0.

**Expected behavior**
- In official OGame, the “armada” message always includes at least one ship in modern versions (v6+).
- Historically (pre-2014), zero-ship results did occur due to rounding errors, but Gameforge patched them out.
- OGameX currently reproduces the legacy behavior.

**Additional context**
`processExpeditionGainShipsOutcome()` uses: `$amount = (int) floor($randomResources / $shipPrice);`

If all `$amount` values are zero, `$units` stays empty, but the message is still sent.
This occurs even with large cargo space and sufficient fleet composition, depending on random ship cost selection and low RNG resource rolls.

Suggested solution:

Add a fallback that guarantees at least one ship when GainShips triggers and the budget can afford any ship:

```if ($units->isEmpty()) {
    $cheapest = min_by($possibleShips, fn($s) => $s->price->resources->sum());
    if ($cargoCapacityConstrainedAmount >= $cheapest->price->resources->sum()) {
        $units->addUnit($cheapest, 1);
    } else {
        // fallback to “nothing salvageable” message
    }
}

### PR DESCRIPTION
**Pull request by Geda173**
## Description
This pull request introduces the Armada Fix, which corrects an edge case in expedition ship recovery.
Previously, expeditions could occasionally return no ships at all, even when the fleet had sufficient cargo capacity and the random ship distribution yielded none.

The new logic ensures that:

- If the ship distribution result is empty ($units->isEmpty()),
- and if the effective “budget” (based on expedition loot factor) allows for at least one ship,

then the cheapest available ship (e.g., Small Cargo) will always be awarded.
This guarantees that expeditions never return completely empty-handed due to RNG imbalance, while still respecting expedition economy and fleet cargo limits.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #768

## Checklist

- [X] **Testing:**
    - Ran local tests but it is not included in this PR, can commit if needed.
- [ ] **CSS & JS Build:** Not applicable, no frontend changes.
- [X] **Documentation:** Inline comments clarify intent and expected behavior.

## Additional Information
This fix improves expedition consistency and player experience by ensuring expeditions always produce logical outcomes.
No breaking changes or balance alterations outside expedition RNG handling.

### DIFF
--- a/app/GameMissions/ExpeditionMission.php
+++ b/app/GameMissions/ExpeditionMission.php
@@ -123,7 +123,7 @@ class ExpeditionMission extends GameMission
                 if ($foundUnits->getAmount() > 0) {
                     $units->addCollection($foundUnits);
                 } else {
-                    // No ships could be granted for this fleet -> treat as Failed (per maintainer review)
+                    // No ships could be granted for this fleet -> treat as Failed
                     $this->processExpeditionFailedOutcome($mission);
                 }
                 break;
@@ -431,9 +431,7 @@ class ExpeditionMission extends GameMission
             }
         }
 
-        // --- Armada fix: Guarantee at least one ship if possible (before building $message_params) ---
         if (empty($units->units)) {
-            // Switch to UnitCollection helper
             $cheapest = $units->findCheapestShip($possibleShips);
 
             if ($cheapest !== null && $cargoCapacityConstrainedAmount >= $cheapest->price->resources->sum()) {

### CODE REVIEW

File: .github/workflows/run-tests-docker-compose.yml [hunk 1]
```diff
@@ -33,7 +33,15 @@ jobs:
...
           timeout 60 bash -c 'until docker compose exec -T ogamex-app php artisan migrate:status > /dev/null 2>&1; do echo "Waiting for application to be ready..." && sleep 2; done'
-      - name: Run Tests
+      # Ensure a clean environment before running the tests
+      - name: Prepare Laravel for tests
+        run: |
+          docker compose exec -T ogamex-app php artisan key:generate --force
+          docker compose exec -T ogamex-app php artisan config:clear
+          docker compose exec -T ogamex-app php artisan cache:clear
+          docker compose exec -T ogamex-app php artisan route:clear
+          docker compose exec -T ogamex-app php artisan migrate:fresh --force
```
lanedirt: Can you explain why this change is necessary? The GitHub actions seem to be working fine AFAIK and during first container start these initialization steps should already be done automatically.

app/GameMissions/ExpeditionMission.php
```diff
@@ -369,8 +369,10 @@ private function processExpeditionGainShipsOutcome(FleetMission $mission): UnitC
             }
         }
 
-        // If no ships can be found, return empty collection
+        // If no ships can be found at all for this fleet composition, fallback to a failure-style outcome
         if (empty($possibleShips)) {
+            $message_variation_id = ExpeditionFailed::getRandomMessageVariationId();
+            $this->messageService->sendSystemMessageToPlayer($player, ExpeditionFailed::class, ['message_variation_id' => $message_variation_id]);
             return new UnitCollection();
         }
```
lanedirt: Instead of sending a `ExpeditionFailed` message here inside the `processExpeditionGainShipsOutcome()` method (twice actually, see my other comment), can you change it to simply return early with an empty UnitCollection here (as you already did).

And then add an extra check to the `switch ($outcome)` part, e.g. make it like this. This avoids duplicate code for sending expedition failed messages.

```php
switch ($outcome) {
  case ExpeditionOutcomeType::GainShips:
      $foundUnits = $this->processExpeditionGainShipsOutcome($mission);
  
      if ($foundUnits->getAmount() > 0) {
        $units->addCollection($foundUnits);
      }
      else {
        $this->processExpeditionFailedOutcome($mission);
      }
      break;
}
```

app/GameMissions/ExpeditionMission.php
```diff
@@ -425,6 +427,29 @@ private function processExpeditionGainShipsOutcome(FleetMission $mission): UnitC
...
+                $units->addUnit($cheapest, 1);
+            }
+        }
+
+        // If still empty here (no eligible or affordable ships), fallback to a failure-style message
+        if (empty($units->units)) {
+            $message_variation_id = ExpeditionFailed::getRandomMessageVariationId();
+            $this->messageService->sendSystemMessageToPlayer($player, ExpeditionFailed::class, ['message_variation_id' => $message_variation_id]);
+            return new UnitCollection();
+        }
```
lanedirt: See my other comment. Simply return an empty collection here without sending a message.

app/GameMissions/ExpeditionMission.php
```diff
@@ -425,6 +427,29 @@ private function processExpeditionGainShipsOutcome(FleetMission $mission): UnitC
...
+                fn ($carry, $s) => $carry === null || $s->price->resources->sum() < $carry->price->resources->sum()
+                    ? $s
+                    : $carry,
+                null
+            );
+
+            if ($cheapest !== null && $cargoCapacityConstrainedAmount >= $cheapest->price->resources->sum()) {
+                $units->addUnit($cheapest, 1);
+            }
+        }
```
lanedirt: This part feels a bit like code golfing, it technically works but I personally find this a bit hard to read. It would be clearer if we avoided the inline `array_reduce` and moved the “find cheapest ship” logic into a dedicated method to the `UnitCollection` class, similar to how other methods in `UnitCollection` (like `getSlowestUnitSpeed()`) are structured.

For example, add this to `UnitCollection.php`:

```php
/**
 * Get the cheapest ship from the given list.
 *
 * @return ?Unit
 */
public function findCheapestShip(): ?Unit
{
    $cheapest = null;
    foreach ($this->units as $entry) {
        if ($cheapest === null || $entry->price->resources->sum() < $cheapest->price->resources->sum()) {
            $cheapest = $entry;
        }
    }

    return $cheapest;
}
```

Then the main logic can become much cleaner and easier to read:

```php
if (empty($units->units)) {
    $cheapest = $possibleShips->findCheapestShip();
    if ($cheapest && $cargoCapacityConstrainedAmount >= $cheapest->price->resources->sum()) {
        $units->addUnit($cheapest, 1);
    }
}
```

This keeps the intent obvious: “find the cheapest ship and add one if affordable” without inline complexity. And also allows us to use the find cheapest ship logic in other places in the future if required.

lanedirt (changes requested):
Hi @Geda173, thanks for looking into fixing this, much appreciated! I added a few comments, can you take a look at these?

.github/workflows/run-tests-docker-compose.yml
Geda173: Previous to all tests coming back as passed, the test run failed once, but only here on Github. On my local machine the tests passed every time, leading me to believe that GitHub Actions can reuse cached layers or partially initialized containers between workflow steps, which sometimes leaves stale configs, cached routes, or mismatched keys. I may be wrong though and this change can be discarded, but during the last commit it failed the docker-compose.yml test.

lanedirt: @Geda173, got it. This might actually have to do with some existing tests that are prone to RNG like combat and expedition outcomes are known to be flaky and can throw a false-negative once in a while. This should be looked at separately. Can you revert this change so we can see if tests will succeed in a second attempt?

Geda173: Absolutely, I just reverted the change and committed it. Waiting for GitHub Actions to do its thing. I will get on the other requested changes later in the day.

app/GameMissions/ExpeditionMission.php
Kays: I think the core issue is that the code gets to this point with zero ships gained.

My suggestion:
Move the calculation of `$cargoCapacityConstrainedAmount` to above `$possibleShips = [];` Then in the loop that adds ship objects into `$possibleShips`, check if a single ship's resource requirement is greater than `$cargoCapacityConstrainedAmount`. If so, do not add it into the array. This way, we don't need to handle this situation at all. Either the array is empty, or it will only contain ships that can be gained based on cargo space.

Site note: I believe that in the official game, Deuterium is exempt from this calculation. IIRC, `resources->sum()` includes Deuterium. I defer if we want to make that change at this moment.

Geda173: @lanedirt Thank's a ton for the detailed feedback. I have given my best to incorporate the requested changes and I agree that this is a much more elegant solution.

@Kays I just saw your comment and I wasn't able to factore that contribution in yet. Filtering out ships whose total cost exceeds the fleet’s effective cargo/resource budget before random selection would indeed prevent us from ever hitting a “zero ships found” edge case in the first place. At this point, the current version still includes the Armada safeguard, but I definitely like the direction of your suggestion. It makes the outcome generation deterministic and elegant. I'll incorporate it in the next iteration so that the expedition reward pool only includes affordable ship types right from the start.

As for the Deuterium: I’ll leave it as-is for now to stay consistent with the existing resource model, but a TODO will go in to revisit that once we confirm how the original game handles it or how @lanedirt would like to have it handled, whichever comes first!

app/GameMissions/ExpeditionMission.php
```diff
@@ -425,6 +431,22 @@ private function processExpeditionGainShipsOutcome(FleetMission $mission): UnitC
             }
         }
 
+        // --- Armada fix: Guarantee at least one ship if possible (before building $message_params) ---
+        if (empty($units->units)) {
+            // Switch to UnitCollection helper
```
lanedirt: This comment can be removed

app/GameMissions/ExpeditionMission.php
```diff
@@ -120,7 +120,12 @@ protected function processArrival(FleetMission $mission): void
                 break;
             case ExpeditionOutcomeType::GainShips:
                 $foundUnits = $this->processExpeditionGainShipsOutcome($mission);
-                $units->addCollection($foundUnits);
+                if ($foundUnits->getAmount() > 0) {
+                    $units->addCollection($foundUnits);
+                } else {
+                    // No ships could be granted for this fleet -> treat as Failed (per maintainer review)
```
lanedirt: The `(per maintainer review)` comment can be removed, does not really add meaning for future maintenance

Geda173:
> Hi @Geda173,
> 
> Looking good! Thanks for the adjustments. I have two small comment tweaks that can be removed, other than that looking good to merge. Let me know when you're ready with any final tweaks.

All done! Once again, thank you for you lightning-fast responses.

lanedirt:
Awesome, thanks for your contribution, looking good! Merging!

PR #780, opened 2025-10-21
Fix Solar Satellite description to display actual energy production per planet

### LINKED ISSUE #756: [BUG] Solar Satellites displaying wrong energy production in description

Geda173:
## Description
Fixes the bug where Solar Satellite description always displays "produces 25 energy on this planet" regardless of the actual energy production, which varies based on planet temperature.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #756 

## Checklist

- [X] **Testing:** Tested across multiple planets with varying temperatures and energy situations. Description now correctly matches green (+X) production bonus in all scenarios.
- [ ] **CSS & JS Build:** No changes made.
- [X] **Documentation:** Inline comments to reflect the changes and thoughts behind them.

## Additional Information
The description now uses the same calculation as the green production bonus:
comparing `getObjectProduction()` for current vs next satellite count. This automatically accounts for both temperature variations and energy deficits.

Implementation:
- Add `getDescription()` method to `UnitViewModel` for dynamic descriptions
- Add `getObjectDescription()` helper to `ObjectAjaxTrait`
- Calculate energy per satellite using production difference method
- Ensure description always matches displayed production bonus

image

### DIFF
--- a/app/ViewModels/UnitViewModel.php
+++ b/app/ViewModels/UnitViewModel.php
@@ -4,7 +4,6 @@ namespace OGame\ViewModels;
 
 use OGame\GameObjects\Models\Abstracts\GameObject;
 use OGame\Models\Resource;
-use OGame\Services\PlanetService;
 
 class UnitViewModel
 {
@@ -38,20 +37,5 @@ class UnitViewModel
     public function getFormattedLong(): string
     {
         return $this->getResource()->getFormattedLong();
-    }
-
-    /**
-     * Get the description with dynamic values replaced based on planet context.
-     * This method handles special cases like Solar Satellite where the description
-     * needs to show planet-specific energy production values.
-     *
-     * @param PlanetService $planet
-     * @return string
-     */
-    public function getDescription(PlanetService $planet): string
-    {
-        // Return the original description without modifications
-        // Dynamic descriptions should only be shown in specific contexts (like AJAX object details)
-        return $this->object->description;
     }
 }


app/Http/Traits/ObjectAjaxTrait.php
```diff
@@ -197,4 +197,37 @@ public function ajaxHandler(Request $request, PlayerService $player): JsonRespon
...
+
+            // Calculate energy per single satellite (the difference between current and next level)
+            $energyPerUnit = abs($production_next->energy->get() - $production_current->energy->get());
+
+            // Replace any occurrence of "produces [number] energy" with the calculated value
+            $description = preg_replace(
+                '/produces \d+ energy/',
+                "produces {$energyPerUnit} energy",
+                $description
+            );
```
lanedirt: The calculation logic in general looks good! However this string replacement part here can be a bit tricky. In general, the OGameX project has the ambition to make all texts translatable. A proof-of-concept has been added for switching between English (default) and Dutch (via footer link). When switching you can check the "Metal mine" description which changes. 

For solar satellites the description is not made translatable yet, but if it would be done, this string replacement logic won't work anymore as it looks for the strings "produces [x] energy" which is different in every language.

1) To prevent the multi-lang issue above, I think it would be better to update the original description to have a placeholder token so it becomes `produces :energy_production energy`. And then look for that placeholder here and replace this specific token, making it work regardless of the actual surrounding (translated) text.

2) Another thing (which I think I commented on your previous PR for this issue), is that the last sentence in the solar satellite description in the official game seems to be sort of dynamic where it's shown in the ObjectAjaxTrait view, but it isn't shown in other places like the technology tab:

This is what it looks like right now in OGameX with this PR (value is not replaced as it's a different view)
Image

This is what it looks like on the official OGame (last sentence is not shown at all):

Image

In order to make this work, I think we'll need to introduce a new (third) `$espionageProbe->description` variation which contains this "dynamic" text where tokens can then be replaced in, and we should then update the UI to determine whether to show this "dynamic text" or not.

lanedirt (changes requested):
Hi @Geda173, thanks for looking into this! Please see my comments below.

app/Http/Traits/ObjectAjaxTrait.php
Geda173: Hi @lanedirt! Thanks for the review. I understand the importance of future-proofing for multi-language support. As a reaction to your comment I have:

- Removed dynamic description logic from `UnitViewModel` to prevent it appearing in the techtree
- Updated the base Solar Satellite description to use generic text without hardcoded energy values
- Modified `ObjectAjaxTrait` to dynamically replace the generic text with planet-specific energy production values
- This ensures the techtree shows a generic description while the building/shipyard menu shows actual energy production per satellite

Note: The techtree description issue you mentioned in the screenshots was already present in the main branch. That wasn’t introduced by my changes. The base description in `CivilShipObjects.php` has always included the energy production text, which explains why it appeared in both techtree and object details views originally. At least the server my players are using has the same display and I haven't touched anything there.

Regarding the translation placeholder approach: I understand your concern about the string replacement not working with future translations. However, I’d like to clarify the scope before proceeding further:
	1.	Should I implement the placeholder token system (`:energy_production: `) as part of this PR?
	2.	Or would you prefer to accept this current fix and handle the translation infrastructure in a separate issue/PR when Solar Satellite descriptions are actually made translatable?

The current implementation follows the same pattern used elsewhere in the codebase and solves the immediate bug. It can be easily refactored later when the translation system is extended to cover this specific text.

Thanks a ton for taking the time to work on this!

lanedirt: Hi @Geda173, thanks for the fixes.

I see the new string replacement implementation, however it looks like its trying to replace text that does not appear in the updated description. E.g. on my test environment I don't see the actual energy production mentioned at all anymore:

image

> Regarding the translation placeholder approach: I understand your concern about the string replacement not working with future translations. However, I’d like to clarify the scope before proceeding further:
> 1. Should I implement the placeholder token system (:energy_production: ) as part of this PR?
> 2. Or would you prefer to accept this current fix and handle the translation infrastructure in a separate issue/PR when Solar Satellite descriptions are actually made translatable?

Yes option 2 is fine by me, I agree it focuses this PR on the immediate fix. Then later when the translation system is added for these ship objects the string/token replacement can be part of that refactor. Your current approach is to the point and allows for an easy refactor later. 👍

Geda173: Ah shit, something must have been borked when I pushed the changes yesterday. I looked into it again and it should be working now:

image

image

lanedirt: Looks good now, thanks for the fix!

app/ViewModels/UnitViewModel.php
```diff
@@ -38,4 +39,19 @@ public function getFormattedLong(): string
...
+     *
+     * @param PlanetService $planet
+     * @return string
+     */
+    public function getDescription(PlanetService $planet): string
+    {
+        // Return the original description without modifications
+        // Dynamic descriptions should only be shown in specific contexts (like AJAX object details)
+        return $this->object->description;
+    }
```
lanedirt: I just tested the PR again locally, fix looks to be working good now!

However this method `getDescription()` seems to have no usages anymore, probably due to refactoring (?). So this method can be removed.

Geda173: Yes, you're right. Initially, I thought the techtree used `UnitViewModel::getDescription() ` to display Solar Satellite descriptions. But as it turns out that's not the case, so this whole block became obsolete. I removed it as well as the now  unused`PlanetService` import at the top.

lanedirt:
Looking good, thanks for the work on this! Merging!

PR #786, opened 2025-10-26
Fix research queue using wrong planet's lab level for time calculation

### LINKED ISSUE #781: [BUG] Queued research uses wrong duration when next item starts
Describe the bug
When a new research item starts automatically from the research queue, the wrong research time can be applied. This occurs when the research queue progresses while the user is viewing a planet with a lower-level research lab than on the planet where the queue was initiated.

To Reproduce

1. Go to your Home Planet (HP) with a level 12 research lab.
2. Queue multiple research tasks.
3. Switch to a different planet with a lower-level research lab (for example, level 8).
4. Wait until the first queued research finishes while you remain on the planet overview with the weaker lab.
5. Observe that the next research in the queue uses the wrong (slower) research time based on the currently viewed planet instead of the one where the research queue was started.

Expected behavior
Queued research items should always use the research time and lab level of the planet where the queue was initially created, regardless of which planet is currently being viewed by the user.

Screenshots
No screenshots available.

Additional context
This issue seems to occur when switching planets between queued research completions or being offline. The problem might be tied to how the server determines the “active” planet when starting the next research item.

Geda173:
Use the planet where research was queued instead of currently viewed planet for calculating research time. Fixes incorrect research duration when viewing a different planet with lower lab level.

## Description
This PR fixes a bug in the research queue system where starting a new research item from the queue would use the currently viewed planet’s research lab level instead of the originating planet’s research lab level to calculate research time.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #781 

## Checklist

- [X] **Testing:** Tests successfully run locally.
- [ ] **CSS & JS Build:** No changes.
- [ ] **Documentation:** No changes.

## Additional Information
This is a minimal one-line fix that addresses the root cause of the issue. The variable  `$planet`  was already being correctly loaded from the queue item’s  `planet_id`, but was not being used for the research time calculation. This fix ensures consistency throughout the queue processing logic.

### DIFF
--- a/app/Services/ResearchQueueService.php
+++ b/app/Services/ResearchQueueService.php
@@ -253,7 +253,7 @@ class ResearchQueueService
 
             // See if the planet has enough resources for this research attempt.
             $price = ObjectService::getObjectPrice($object->machine_name, $planet);
-            $research_time = $player->planets->current()->getTechnologyResearchTime($object->machine_name);
+            $research_time = $planet->getTechnologyResearchTime($object->machine_name);
 
             // Only start the queue item if there are no other queue items researching
             // for this planet.


lanedirt:
Thanks for the fix! LGTM 👍

PR #788, opened 2025-10-26
Display 'Deep space' as destination for expedition fleet missions

### LINKED ISSUE #766: [BUG] Expeditions do not show a labelled destination in fleet movement overview
**Describe the bug**
Sending an expedition doesn't show a destination in the fleet movement overview in `/overview`:

Image

**To Reproduce**
1. Dispatch fleet to expedition on x:x:16
2. Go to "Overview"
3. Open the drop down Fleet Movement menu

**Expected behavior**
It should show the destination as "Deep space":
Image

Geda173:
## Description
Expeditions now correctly display "Deep space" instead of just coordinates/blank name

Changes:
- Added `PlanetType::DeepSpace = 4` to the `PlanetType enum`
- Modified `GameMission::start()` to `set type_to = DeepSpace` for expeditions  
- Updated `eventrow.blade.php` to display "Deep space" for `DeepSpace planet type`

Implements contributor-suggested (thanks @Kays !) approach using `PlanetType enum` for type-safe
handling of expedition destinations.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #766 

## Checklist

- [X] **Testing:** Tests successfully run locally.
- [X] **CSS & JS Build:** Added `.deep-space-text` styling
- [X] **Documentation:** Updated documentation to indicate newly added enum portions.

### DIFF
--- a/public/css/ingame.css
+++ b/public/css/ingame.css
@@ -9477,9 +9477,6 @@ figure.planetIcon.tf {
 figure.planetIcon.planet {
     display: none;
 }
-figure.planetIcon.deep-space {
-    display: none;
-}
 
 /* ****************** Small Officer and Ressource Images ************************** */
 .smallOfficer,
--- a/resources/css/ingame/base1.css
+++ b/resources/css/ingame/base1.css
@@ -6418,6 +6418,14 @@ figure.planetIcon.planet {
     display: none
 }
 
+figure.planetIcon.deep-space {
+    display: none;
+}
+
+.deep-space-text {
+    margin-left: 21px;
+}
+
 .smallOfficer,.officer20,.officer32 {
     display: inline-block;
     float: left
--- a/resources/views/ingame/fleetevents/eventrow.blade.php
+++ b/resources/views/ingame/fleetevents/eventrow.blade.php
@@ -29,7 +29,7 @@
                     <figure class="planetIcon tf js_hideTipOnMobile" title="Debris Field"></figure>debris field
                     @break
                 @case (OGame\Models\Enums\PlanetType::DeepSpace)
-                    <figure class="planetIcon deep-space js_hideTipOnMobile" title="Deep space"></figure>{{ __('Deep space') }}
+                    <span class="deep-space-text">{{ __('Deep space') }}</span>
                     @break
             @endswitch
         </td>
@@ -100,7 +100,7 @@
                     <figure class="planetIcon tf js_hideTipOnMobile" title="Debris Field"></figure>debris field
                     @break
                 @case (OGame\Models\Enums\PlanetType::DeepSpace)
-                    <figure class="planetIcon deep-space js_hideTipOnMobile" title="Deep space"></figure>{{ __('Deep space') }}
+                    <span class="deep-space-text">{{ __('Deep space') }}</span>
                     @break
             @endswitch
         </td>
@@ -147,7 +147,7 @@
                     <figure class="planetIcon tf js_hideTipOnMobile" title="Debris Field"></figure>debris field
                     @break
                 @case (OGame\Models\Enums\PlanetType::DeepSpace)
-                    <figure class="planetIcon deep-space js_hideTipOnMobile" title="Deep space"></figure>{{ __('Deep space') }}
+                    <span class="deep-space-text">{{ __('Deep space') }}</span>
                     @break
             @endswitch
         </td>
@@ -218,7 +218,7 @@
                     <figure class="planetIcon tf js_hideTipOnMobile" title="Debris Field"></figure>debris field
                     @break
                 @case (OGame\Models\Enums\PlanetType::DeepSpace)
-                    <figure class="planetIcon deep-space js_hideTipOnMobile" title="Deep space"></figure>{{ __('Deep space') }}
+                    <span class="deep-space-text">{{ __('Deep space') }}</span>
                     @break
             @endswitch
         </td>


app/Http/Controllers/FleetEventsController.php
```diff
@@ -115,7 +115,9 @@ public function fetchEventList(PlayerService $player, FleetMissionService $fleet
             $eventRowViewModel->destination_planet_coords = new Coordinate($row->galaxy_to, $row->system_to, $row->position_to);
             $eventRowViewModel->destination_planet_type = PlanetType::from($row->type_to);
 
-            if ($row->planet_id_to !== null) {
+            if ($row->mission_type == 15) {
+                $eventRowViewModel->destination_planet_name = 'Deep space';
```
lanedirt: The goal for OGameX is to be fully multilingual (in the future), so for any literal text strings that you add, could you also add the default Laravel translation wrappers, like so:

https://laravel.com/docs/12.x/localization#retrieving-translation-strings

```php
$eventRowViewModel->destination_planet_name = __('Deep space');
```

This will enable this string to be translated. During refactors I try to add these multilang wrappers to all literal strings that I find. It's a best effort way, but every little bit helps in the long run 😄

Geda173: You are right, I will try to keep that in mind for the future!

Geda173:
I have ancountered a tricky problem during more testing:

When a fleet returns after the hold time (and only then!) does the display in the fleet view become mangled, e.g. it displays both origin and destination as Deep space. This has sent me down a rabbit hole for all sorts of issues. When recalling a fleet, the view now bugs out and cant be loaded until the fleet has arrived and the event is resolved. I am currently working through these problems one by one and I will submit a fix soon(-ish).

Kays:
> I have ancountered a tricky problem during more testing:
> 
> When a fleet returns after the hold time (and only then!) does the display in the fleet view become mangled, e.g. it displays both origin and destination as Deep space. This has sent me down a rabbit hole for all sorts of issues. When recalling a fleet, the view now bugs out and cant be loaded until the fleet has arrived and the event is resolved. I am currently working through these problems one by one and I will submit a fix soon(-ish).

I think there are some ways to handle this cleanly. I'll mention the way I would tackle it if I were doing it, and few alternatives:
1. This would be the way I would do it.
- Add a new PlanetType for "DeepSpace".
- Set the database `fleet_missions.type_to` to the new type when Expedition missions are created. Currently, Expeditions are set to "Planet" type, which is inaccurate data anyway.
- In `resources/views/ingame/fleetevents/eventrow.blade.php`, add a new `switch` case for the new type and show accordingly. There are existing switch for the other three types already, so adding a new `case` is simple.
2. Keep your changes in the Controller class, but instead of checking by mission type, check by coordinates - position = 16, then set the planet names accordingly for `$eventRowViewModel->origin_planet_name` and `$eventRowViewModel->destination_planet_name`. That should work; albeit, not as clean as the first approach because "Deep space" is not a name of a planet.
3. Similar to the second approach, but do the same coordinate checks in the `eventrow.blade.php` view instead. IMO, this is cleaner than the second approach, distinguishing between display-specific details and avoiding misnomers in field names. Although, if you were going to do this approach, may as well just do the first approach.

Geda173:
> > I have ancountered a tricky problem during more testing:
> > When a fleet returns after the hold time (and only then!) does the display in the fleet view become mangled, e.g. it displays both origin and destination as Deep space. This has sent me down a rabbit hole for all sorts of issues. When recalling a fleet, the view now bugs out and cant be loaded until the fleet has arrived and the event is resolved. I am currently working through these problems one by one and I will submit a fix soon(-ish).
> 
> I think there are some ways to handle this cleanly. I'll mention the way I would tackle it if I were doing it, and few alternatives:
> 
> 1. This would be the way I would do it.
> 
> * Add a new PlanetType for "DeepSpace".
> * Set the database `fleet_missions.type_to` to the new type when Expedition missions are created. Currently, Expeditions are set to "Planet" type, which is inaccurate data anyway.
> * In `resources/views/ingame/fleetevents/eventrow.blade.php`, add a new `switch` case for the new type and show accordingly. There are existing switch for the other three types already, so adding a new `case` is simple.

That was actually a great suggestion! It took some time to implement but I _think_ I got it now.
I have now fully adopted the enum-based approach for expeditions:
- I set  `type_to`  to  `PlanetType::DeepSpace`  for expedition missions instead of relying on mission type logic or string comparisons.
- The fleet event display now shows “Deep space” by checking for this PlanetType, both for outbound and return journeys.
- I also fixed an issue with return timing for expeditions: the return journey now correctly begins after the expedition hold phase, using the intended one-way flight time, matching OGame’s convention.

Thanks again for the guidance. This method is much more robust and maintainable. If you spot anything else or have further suggestions, I’m happy to adjust!

Kays:
@Geda173 Good work! A few things:

1. https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R293
Not sure what this is for. Perhaps it's for a different, unrelated, issue. It doesn't look like a complete implementation based on what's submitted in the MR either. Just in case it helps, in command line - `git add -p <file>` - allows partially adding files for commit so you don't risk adding more than you need.

2. https://github.com/lanedirt/OGameX/pull/788/files#diff-f331d7658739dd97a140407c8e9d2eff9dc37cc6c6cb4d15f5d4124ac568d938R119
Is this still necessary?

3. https://github.com/lanedirt/OGameX/pull/788/files#diff-6ed4a697a6eeed9b55dab404e9ea1eb53cb1b902b9cada544e1f4f6fa5b74ceaR34
Perhaps this can be converted to doing what the other switch/case do - have the name on the same line as the `<figure>` output. Having a `case` that only `break` does nothing. Would be cleaner than doing an additional check.

Geda173:
> @Geda173 Good work! A few things:
> 
> 1. https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R293
>    Not sure what this is for. Perhaps it's for a different, unrelated, issue. It doesn't look like a complete implementation based on what's submitted in the MR either. Just in case it helps, in command line - `git add -p <file>` - allows partially adding files for commit so you don't risk adding more than you need.

I updated the original PR before I answered your comment and should have probably included the changes as a comment as well to reflect the timeline. This line is for the expedition return time bug. It was so frustrating to test the Deep space fix because I needed to troubleshoot the return event to see whether the problem with the planet name being overwritten by Deep space persisted. So I included the fix here as well.

> 2. https://github.com/lanedirt/OGameX/pull/788/files#diff-f331d7658739dd97a140407c8e9d2eff9dc37cc6c6cb4d15f5d4124ac568d938R119
>    Is this still necessary?

If all frontend display logic for expeditions is converted to switch/case on `PlanetType`, just as it is for planet/moon/debris, this explicit check is no longer necessary. We can safely remove the separate if and handle "Deep space" as an additional case in the blade switch/case, keeping the controller clean.

> 3. https://github.com/lanedirt/OGameX/pull/788/files#diff-6ed4a697a6eeed9b55dab404e9ea1eb53cb1b902b9cada544e1f4f6fa5b74ceaR34
>    Perhaps this can be converted to doing what the other switch/case do - have the name on the same line as the <figure>`output. Having a "case" that only "break" does nothing. Would be cleaner than doing an additional check.

Great suggestion! I’ll move the DeepSpace name/output into the main switch/case in the blade template, just like for planet/moon/debris. That keeps the markup cleaner and removes the need for a separate conditional.

Kays:
> I updated the original PR before I answered your comment and should have probably included the changes as a comment as well to reflect the timeline. This line is for the expedition return time bug. It was so frustrating to test the Deep space fix because I needed to troubleshoot the return event to see whether the problem with the planet name being overwritten by Deep space persisted. So I included the fix here as well.

What if I recall the Expedition mission?

Geda173:
> > I updated the original PR before I answered your comment and should have probably included the changes as a comment as well to reflect the timeline. This line is for the expedition return time bug. It was so frustrating to test the Deep space fix because I needed to troubleshoot the return event to see whether the problem with the planet name being overwritten by Deep space persisted. So I included the fix here as well.
> 
> What if I recall the Expedition mission?

Recalling an expedition aborts the mission before reaching the hold period at Deep space/the expedition. The timing fix for planned returns (after the hold) does not affect recalled expeditions as these always follow the standard recall mechanic. The previous fix that I had submitted did two things (wrong):

- When on a "proper" return from an expedition it would insert the planet name for both origin and destination.
- When recalling a fleet, the fleet overview broke entirely, defaulting to an infinite loading circle and only resolving once the fleet had fully returned and the event was resolved.

So recalling missions broke something in a different place, while a return after the hold time broke something else.

Or maybe I am misunderstanding your question?

Kays:
> Recalling an expedition aborts the mission before reaching the hold period at Deep space/the expedition. The timing fix for planned returns (after the hold) does not affect recalled expeditions as these always follow the standard recall mechanic. The previous fix that I had submitted did two things (wrong):
> 
>     * When on a "proper" return from an expedition it would insert the planet name for both origin and destination.
> 
>     * When recalling a fleet, the fleet overview broke entirely, defaulting to an infinite loading circle and only resolving once the fleet had fully returned and the event was resolved.
> 
> 
> So recalling missions broke something in a different place, while a return after the hold time broke something else.
> 
> Or maybe I am misunderstanding your question?

I haven't tested this on the app, so I may be wrong. I'm just going off what I'm reading from your MR change. As far as I know, the only thing that recalling a mission does is setting the `time_arrival` to the current time, then it processes `startReturn` per normal. I don't recall Expedition overriding that method, but I may be wrong. Looking at the current change in your `GameMission` class in this MR, you will add the holding time to the start time. Does that not mean that if I recall my Expedition mission, I will be penalized by the hold time + return trip difference? It's possible that you have it implemented properly in your test environment and the MR is missing all changes necessary, but from what I'm seeing, it looks bugged. If that is the case, I do suggest testing in a clean environment with only the MR changes. Again, I haven't pulled in your changes and tested them, so I may be wrong, and if so, I stand corrected. I am just raising a possible concern.

Geda173:
> > Recalling an expedition aborts the mission before reaching the hold period at Deep space/the expedition. The timing fix for planned returns (after the hold) does not affect recalled expeditions as these always follow the standard recall mechanic. The previous fix that I had submitted did two things (wrong):
> > ```
> > * When on a "proper" return from an expedition it would insert the planet name for both origin and destination.
> > 
> > * When recalling a fleet, the fleet overview broke entirely, defaulting to an infinite loading circle and only resolving once the fleet had fully returned and the event was resolved.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So recalling missions broke something in a different place, while a return after the hold time broke something else.
> > Or maybe I am misunderstanding your question?
> 
> I haven't tested this on the app, so I may be wrong. I'm just going off what I'm reading from your MR change. As far as I know, the only thing that recalling a mission does is setting the `time_arrival` to the current time, then it processes `startReturn` per normal. I don't recall Expedition overriding that method, but I may be wrong. Looking at the current change in your `GameMission` class in this MR, you will add the holding time to the start time. Does that not mean that if I recall my Expedition mission, I will be penalized by the hold time + return trip difference? It's possible that you have it implemented properly in your test environment and the MR is missing all changes necessary, but from what I'm seeing, it looks bugged. If that is the case, I do suggest testing in a clean environment with only the MR changes. Again, I haven't pulled in your changes and tested them, so I may be wrong, and if so, I stand corrected. I am just raising a possible concern.

Gotcha, I think I understand it now. The hold time of the expedition mission cannot be interrupted via recall. I haven't paid any particular attention to this problem because it never occured to me that it could be a problem. Without this PR, aborting the hold time is not possible and with this PR it should also not be possible to abort the hold time, only the initial approach to the expedition from the origin point. Once the fleet reaches the expedition and starts its hold, it is out of the players's hands. But to be certain I'll check tomorrow.

Kays:
> Gotcha, I think I understand it now. The hold time of the expedition mission cannot be interrupted via recall. I haven't paid any particular attention to this problem because it never occured to me that it could be a problem. Without this PR, aborting the hold time is not possible and with this PR it should also not be possible to abort the hold time, only the initial approach to the expedition from the origin point. Once the fleet reaches the expedition and starts its hold, it is out of the players's hands. But to be certain I'll check tomorrow.

It may be best to pull out this change and do it in a separate PR. There may be nuances. Without your change in this PR, the only thing affected is that fleets may return, nearly instantly, after their holds, but recalls are calculated correctly. With your changes, the inverse is true - recalls will be penalized by the hold time, but returns after holds are properly calculated. From a player experience perspective, the original issue is more welcomed than the new one, because the original issue never penalized the player.

Geda173:
I have reverted the return time change and implemented some other changes that now hopefully display Deep space correctly in the fleet overview menu:

- Modified GameMission.php to set `type_to` as `DeepSpace` (enum value 4) for expedition missions (that was already in there)
- Added CSS rule to hide deep-space icon, so that the text in the fleet overview is properly aligned
- Fixed issue where expeditions incorrectly displayed planet names as destinations

Kays:
@Geda173 Good work. It looks good. Final mentions:

1. https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R222
Since there's already an `if` statement literally right above it that checks for the same thing, perhaps, this change is unnecessary. You can just add another statement inside the above `if` statement to set the correct type - `$targetType = PlanetType::from(PlanetType::DeepSpace);`, then the original code to set the mission type from `$targetType` should work as intended. This keeps the Expedition-specific logic/nuances in one block. This also saves minor calculations in this `if` block: https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R232

2. I preface by saying that I prefer the UI changes that you made for Deep Space, but lanedirt is adamant in a replication effect. As such, the way the official OGame does it is as such:
image

Geda173:
> @Geda173 Good work. It looks good. Final mentions:
> 
> 1. https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R222
>    Since there's already an `if` statement literally right above it that checks for the same thing, perhaps, this change is unnecessary. You can just add another statement inside the above `if` statement to set the correct type - `$targetType = PlanetType::from(PlanetType::DeepSpace);`, then the original code to set the mission type from `$targetType` should work as intended. This keeps the Expedition-specific logic/nuances in one block. This also saves minor calculations in this `if` block: https://github.com/lanedirt/OGameX/pull/788/files#diff-e525b1313033870e6d6c040232e27b12e221f6deceea3230184c5c23aeb1e7b8R232

You are absolutely right and that is a much cleaner and easier to maintain approach. I have implemented it at once!
> 2. I preface by saying that I prefer the UI changes that you made for Deep Space, but lanedirt is adamant in a replication effect. As such, the way the official OGame does it is as such:
> 
> image

Here I am not too sure what the issue is. I'll let @lanedirt weigh in on this because I don't see what the difference is and I wouldn't know what to change. I currently have my version and the official version open side by side and I cannot make out a difference.

Kays:
> You are absolutely right and that is a much cleaner and easier to maintain approach. I have implemented it at once!
```php
        if (static::class === ExpeditionMission::class) {
            $mission->time_holding = $holdingHours * 3600;
            $targetType = PlanetType::DeepSpace;
        }

        $mission->type_to = $targetType->value;
```
Does this work? `PlanetType::DeepSpace` is just a number right? Can `$targetType->value` be used correctly after? Either that needs to be a `PlanetType` object like `PlanetType::from(PlanetType::DeepSpace)` or move the `if` block to below `$mission->type_to` line like so:

```php
        $mission->type_to = $targetType->value;

        if (static::class === ExpeditionMission::class) {
            $mission->time_holding = $holdingHours * 3600;
            $mission->type_to = PlanetType::DeepSpace;
        }
```

> Here I am not too sure what the issue is. I'll let @lanedirt weigh in on this because I don't see what the difference is and I wouldn't know what to change. I currently have my version and the official version open side by side and I cannot make out a difference.

Just the class name and title on the `<figure>` markdown. It's hidden so no one can see it anyway. Like I said, I do like your changes, but if we want a replication, then the class name should be "planet" and title should be "Planet". Just a nitpick. I don't see any faults with what you have done and I do like the change.

Geda173:
> > You are absolutely right and that is a much cleaner and easier to maintain approach. I have implemented it at once!
> 
> ```
>         if (static::class === ExpeditionMission::class) {
>             $mission->time_holding = $holdingHours * 3600;
>             $targetType = PlanetType::DeepSpace;
>         }
> 
>         $mission->type_to = $targetType->value;
> ```
> 
> Does this work? `PlanetType::DeepSpace` is just a number right? Can `$targetType->value` be used correctly after? Either that needs to be a `PlanetType` object like `PlanetType::from(PlanetType::DeepSpace)` or move the `if` block to below `$mission->type_to` line like so:
> 
> ```
>         $mission->type_to = $targetType->value;
> 
>         if (static::class === ExpeditionMission::class) {
>             $mission->time_holding = $holdingHours * 3600;
>             $mission->type_to = PlanetType::DeepSpace;
>         }
> ```

`PlanetType::DeepSpace` is a backed enum case (with integer value 4), so assigning it to `$targetType` keeps the type consistent as a `PlanetType` enum. When we later call `$targetType->value`, it properly returns the integer value `4` whether `$targetType` is the original parameter or the overridden `PlanetType::DeepSpace` enum case.

The flow is:

1. For non-expedition missions: `$targetType` stays as the passed enum → `->value` returns its int value

2. For expeditions: `$targetType` is reassigned to `PlanetType::DeepSpace` enum → `->value` returns `4`
3. 
Both paths work because we're always dealing with `PlanetType` enum cases, and all backed enums support the `->value` property.

> > Here I am not too sure what the issue is. I'll let @lanedirt weigh in on this because I don't see what the difference is and I wouldn't know what to change. I currently have my version and the official version open side by side and I cannot make out a difference.
> 
> Just the class name and title on the `<figure>` markdown. It's hidden so no one can see it anyway. Like I said, I do like your changes, but if we want a replication, then the class name should be "planet" and title should be "Planet". Just a nitpick. I don't see any faults with what you have done and I do like the change.

Oh wow, okay, I can see it now. Thanks for explaining, it would have taken me ages to see the difference.

Kays:
> `PlanetType::DeepSpace` is a backed enum case (with integer value 4), so assigning it to `$targetType` keeps the type consistent as a `PlanetType` enum. When we later call `$targetType->value`, it properly returns the integer value `4` whether `$targetType` is the original parameter or the overridden `PlanetType::DeepSpace` enum case.
> 
> The flow is:
> 
>     1. For non-expedition missions: `$targetType` stays as the passed enum → `->value` returns its int value
> 
>     2. For expeditions: `$targetType` is reassigned to `PlanetType::DeepSpace` enum → `->value` returns `4`
> 
> 
> Both paths work because we're always dealing with `PlanetType` enum cases, and all backed enums support the `->value` property.

Thanks for clarifying and educating me on this. I don't work with this language much so I'm not good at it; hence I asked just to be safe.

public/css/ingame.css
```diff
@@ -9477,6 +9477,9 @@ figure.planetIcon.tf {
 figure.planetIcon.planet {
     display: none;
 }
+figure.planetIcon.deep-space {
+    display: none;
+}
```
lanedirt: CSS changes should be made in one of the CSS source files in `resources/css/ingame/`, afterwards you should run the following command (in `ogamex-app` container):

```
npm run dev watch
```
https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md#6-run-css-and-js-build

This will recompile this `public/css/ingame.css` based on the source files and ensures no changes get lost.

lanedirt:
Thanks, looking good! Merging!

PR #797, opened 2025-11-04
fix fleets on missions are not counted towards highscore

### LINKED ISSUE #793: [BUG] Ships on missions are removed from point/highscore calculation
**Describe the bug**
When ships are sent on a mission, the highscore points they yield will be exempt from the highscore calculation for as long as the ships are not stationed on a planet.

**To Reproduce**
1. Send ships on a mission.
2. Wait for the highscore calculation to happen.
3. Watch the points highscore points drop.

**Expected behavior**
Ships should be counted towards the point / highscore calculation, whether they are in motion or not.

rbardtke:
## Description
fix fleets on missions are not counted towards highscore

### Type of Change:
- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #793 

## Checklist

- [ ] **Testing:**

## Additional Information

### DIFF
--- a/app/Services/HighscoreService.php
+++ b/app/Services/HighscoreService.php
@@ -71,11 +71,6 @@ class HighscoreService
                 continue;
             }
 
-            // Skip return missions - ships are already counted in the parent mission
-            if ($mission->parent_id !== null) {
-                continue;
-            }
-
             // Calculate score for all ships in this mission
             foreach (ObjectService::getShipObjects() as $ship) {
                 $amount = $mission->{$ship->machine_name} ?? 0;
@@ -109,11 +104,6 @@ class HighscoreService
         foreach ($activeMissions as $mission) {
             // Skip processed missions (already counted on planet)
             if ($mission->processed) {
-                continue;
-            }
-
-            // Skip return missions - ships are already counted in the parent mission
-            if ($mission->parent_id !== null) {
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


app/Services/HighscoreService.php
```diff
@@ -49,6 +50,135 @@ public function setHighscoreType(int $type): void
...
+        foreach ($activeMissions as $mission) {
+            // Skip processed missions (already counted on planet)
+            if ($mission->processed) {
+                continue;
+            }
+
+            // Skip return missions - ships are already counted in the parent mission
+            if ($mission->parent_id !== null) {
+                continue;
+            }
```
lanedirt: I'm not sure if this will work. Because the `getActiveFleetMissionsSentByCurrentPlayer()` method should only return active missions that are not processed yet.

And so if a parent mission has been completed and a return mission is started, only the return mission will be returned by the method above. So I think we shouldn't skip the return missions here, but process them as usual.

Can you double check this?

app/Services/HighscoreService.php
```diff
@@ -49,6 +50,135 @@ public function setHighscoreType(int $type): void
...
```
lanedirt: See the other comment

lanedirt (changes requested):
Hi @rbardtke,

Thanks for looking into this! Overall the approach looks solid to me, but I do have a comment regarding the parent mission check: can you take a look at this?

app/Services/HighscoreService.php
rbardtke: @lanedirt you are right this was wrong, i fixed and tested it on my local instance, works fine for me

lanedirt:
Awesome, thanks for the fix. LGTM, merging!

PR #827, opened 2025-11-24
Fix #806: Allow sending recyclers to invisible debris fields

### LINKED ISSUE #806: [BUG] No invisible debries field
**Describe the bug**
after completely mining a debris field, you are not able to send more recyclers to the debris field.

**To Reproduce**
1. mine debries field completely
2. Go to fleet, try to send a recycler to the already mined debries field
3. you will get a message there is no debries field

**Expected behavior**
It is possible to send ships to a debris field that does not show up in the Galaxy View, as any field with not more than than 300 total resources will not display on the galaxy screen (called an invisible DF).

https://ogame.fandom.com/wiki/Debris_Field

nguyenthanhan:
## Description

Fixes bug where players couldn't send recyclers to completely harvested debris fields. Implements proper "ghost debris field" mechanics matching official OGame behavior.

### Type of Change:

- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #806

## Checklist

- [x] **Testing:** All 11 recycle mission tests pass locally.
- [x] **CSS & JS Build:** No CSS/JS changes.
- [x] **Documentation:** No documentation changes needed.

### Behavior

- Can send recyclers to debris fields with resources (visible in Galaxy View)
- Can send recyclers to ghost debris fields with 0 resources (invisible)
- Cannot send recyclers to coordinates that never had debris fields
- Ghost fields are automatically cleaned up every Monday at 1:00 AM

### DIFF
--- a/app/Console/Commands/ResetDebrisFields.php
+++ b/app/Console/Commands/ResetDebrisFields.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace OGame\Console\Commands;
+
+use Illuminate\Console\Command;
+use OGame\Models\DebrisField;
+
+class ResetDebrisFields extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'ogamex:reset-debris-fields';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Resets (deletes) empty debris fields. In OGame, this happens weekly on Monday at 1:00 AM.';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): void
+    {
+        $this->info('Resetting empty debris fields...');
+
+        // Delete all debris fields that have no resources
+        $deletedCount = DebrisField::where('metal', 0)
+            ->where('crystal', 0)
+            ->where('deuterium', 0)
+            ->delete();
+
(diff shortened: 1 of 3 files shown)


Geda173:
Hi @nguyenthanhan,
thanks for taking the time to submit a PR!

Removing the check is certainly the simplest way to approximate the behavior, but it unfortunately doesn’t match how debris fields work in OGame. In the original game, debris fields remain as invisible “ghost” fields after being harvested, and they’re reset every Monday at 1:00 AM. Because of that, you can’t fly to just any coordinate’s debris field; a field must have existed at some point since the last reset.

This means that a fully accurate implementation is a bit more involved than it looks at first glance.

anngth:
Hi @Geda173, thank you for pointing that out!

I've updated the pull request (PR) to address the issue properly:

- The validation now checks whether a debris field exists in the database, rather than just verifying its resources.
- I added a command to clean up empty debris fields.
- The cleanup is scheduled to run every Monday at 1:00 AM.

Now, we can send recyclers to a harvested debris field until the weekly reset, but we won't be able to send them to random coordinates that never had a debris field.

Please let me know if this looks better!

lanedirt:
Hi @nguyenthanhan, thanks for the work on this. Nice detail to make it work like the official implementation. Change looks good to me, merging!

Thanks @Geda173 for the review!

PR #831, opened 2025-11-27
fix: Preserve resources sent with fleet missions (#754)

### LINKED ISSUE #754: [BUG] Fleet events delete carried resources
**Describe the bug**
When sending resources with a fleet to achieve a task, upon event generation the carried resources are deleted. Example: Sending an expedition with resources to the expedition slot or sending a recycler with resources to mine a debris field will override the carried resources once the result of that event is generated.

**To Reproduce**
1. Go to Fleet.
2. Select Recycler or other ships.
3. Enter coordinates for the expedition slot (x:x:16) or for a debris field. Dispatch fleet.
4. Once the event happens, the cargo will be overwritten, resulting in lost resources.

**Expected behavior**
The resources should stay with the fleet and be delivered upon return. Unless a black hole swallows them during the expedition. Saving to debris fields is a standard saving tactic in OGame.

nguyenthanhan:
## Description
Fixes bug where resources sent with fleet missions (expeditions, recyclers) were deleted when the mission event was processed. This affected fleetsave tactics, where players send resources with their fleets.

**Changes:**
- Modified `startReturn()` to preserve original resources by default
- Added `preserveParentResources` parameter for resource handling control
- Fixed Transport/Colonisation missions to clear delivered resources
- Added 5 comprehensive tests covering all affected mission types

### Type of Change:
- [x] Bug fix

## Related Issues
Fixes #754

## Checklist
- [x] **Code Standards:** PSR-12 verified with Laravel Pint
- [x] **Static Analysis:** PHPStan passed with no errors
- [x] **Testing:** 5 new tests added, all 111 fleet dispatch tests pass (1845 assertions)
- [x] **CSS & JS Build:** No frontend changes
- [x] **Documentation:** Code comments updated

## Test Coverage
New tests verify:
- Expedition preserves resources when nothing found
- Expedition preserves + adds found resources
- Recycle preserves + adds harvested resources
- Recycle to empty debris field (ghost field fleetsave)
- Attack recall preserves resources

All existing fleet dispatch tests pass without regression.

### DIFF
--- a/app/GameMissions/ColonisationMission.php
+++ b/app/GameMissions/ColonisationMission.php
@@ -114,12 +114,6 @@ class ColonisationMission extends GameMission
         $resources = $this->fleetMissionService->getResources($mission);
         $target_planet->addResources($resources);
 
-        // Clear resources from mission since they've been delivered to the new colony
-        // This prevents them from being returned in the return trip
-        $mission->metal = 0;
-        $mission->crystal = 0;
-        $mission->deuterium = 0;
-
         // Mark the arrival mission as processed
         $mission->processed = 1;
         $mission->save();
@@ -131,6 +125,7 @@ class ColonisationMission extends GameMission
         $units->removeUnit($colony_ship, 1);
 
         // Create and start the return mission (if the colonisation mission had ships other than the colony ship itself).
+        // Resources were delivered to the new colony, so return with empty cargo.
         $this->startReturn($mission, new Resources(0, 0, 0, 0), $units);
     }
 
--- a/tests/Feature/FleetDispatch/FleetDispatchExpeditionTest.php
+++ b/tests/Feature/FleetDispatch/FleetDispatchExpeditionTest.php
@@ -358,13 +358,13 @@ class FleetDispatchExpeditionTest extends FleetDispatchTestCase
         $highscore->delete();
 
         // Get the return trip mission for the original mission
-        // and assert that total resources is exactly 100k (limited by cargo capacity) + 2 (original resources sent)
+        // and assert that total resources is exactly 100k (limited by cargo capacity)
         $returnTripMission = $fleetMissionService->getFleetMissionByParentId($originalMission->id, false);
         $totalResources = $returnTripMission->metal + $returnTripMission->crystal + $returnTripMission->deuterium;
         $this->assertEquals(
-            100002,
+            100000,
             $totalResources,
-            'Total resources should be exactly 100k (found) + 2 (original resources sent) when sending 4 large cargos'
+            'Total resources should be exactly 100k when sending 4 large cargos (cargo capacity limit)'
         );
     }
 
--- a/app/GameMissions/Abstracts/GameMission.php
+++ b/app/GameMissions/Abstracts/GameMission.php
@@ -124,8 +124,8 @@ abstract class GameMission
         $mission->save();
 
         // Start the return mission with the resources and units of the original mission.
-        // Use preserveParentResources=false because getResources() already includes parent mission resources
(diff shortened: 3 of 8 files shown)


tests/Feature/FleetDispatch/FleetDispatchExpeditionTest.php
```diff
@@ -358,13 +358,13 @@ public function testExpeditionWithGainResourcesResultSmallFleet(): void
...
         $returnTripMission = $fleetMissionService->getFleetMissionByParentId($originalMission->id, false);
         $totalResources = $returnTripMission->metal + $returnTripMission->crystal + $returnTripMission->deuterium;
         $this->assertEquals(
-            100000,
+            100002,
             $totalResources,
-            'Total resources should be exactly 100k when sending 4 large cargos (cargo capacity limit)'
+            'Total resources should be exactly 100k (found) + 2 (original resources sent) when sending 4 large cargos'
         );
     }
```
lanedirt: Is this assertion correct? The total cargo capacity for the 4 large cargos should be 100k, but here we are asserting 100.002 because the parent mission included 2 resources.

If the parent mission sends resources, shouldn't we expect the total gained resources to be 99.998 instead, making the same max. cargo total of exactly 100.000? I haven't checked how this works in the official game, just wondering.

app/GameMissions/ColonisationMission.php
```diff
@@ -114,6 +114,12 @@ protected function processArrival(FleetMission $mission): void
         $resources = $this->fleetMissionService->getResources($mission);
         $target_planet->addResources($resources);
 
+        // Clear resources from mission since they've been delivered to the new colony
+        // This prevents them from being returned in the return trip
+        $mission->metal = 0;
+        $mission->crystal = 0;
+        $mission->deuterium = 0;
+
```
lanedirt: I'm running into an issue with clearing the parent mission's original resources during processing. I understand your reasoning, but from a logging and debugging standpoint, I believe we should retain the original resources on the parent mission rather than zeroing them out. Keeping them intact allows us to accurately trace how many resources were initially sent, which is extremely useful when troubleshooting.

Instead of having `$preserveParentResources` default to true, I think it would be cleaner to handle this logic explicitly within each mission type's `process()` method and manually append the parent resources to the $resources passed into startReturn() where appropriate. This gives us clearer intent and avoids surprising side effects.

From a long-term design perspective, this approach improves log clarity. For example, for attack missions, logs would show:

1. Attack mission launched with 100k resources onboard (cargo capacity 1M).
2. Battle occurs; max loot is calculated using carried resources + cargo capacity, so even if 100M is available, only 900k is gained (shown in battle report).
3. Return mission is created with the full 1M resources.

This way, logs clearly reflect the full flow:
- Parent mission launched with 100k resources (still visible).
- Battle report showing 900k gained (still visible).
- Return mission carrying 1M total (still visible).

Thoughts?

tests/Feature/FleetDispatch/FleetDispatchExpeditionTest.php
anngth: Ok, my bad! You're correct. If 2 resources are sent, only 99,998 space remains for found resources. Total should be 100,000, not 100,002. I'll check the expedition logic and fix this.

app/GameMissions/ColonisationMission.php
anngth: Ah yeah, that's a much cleaner approach! Keeping the parent mission resources intact makes debugging way easier 👍

lanedirt:
Thanks for the updates and your work on this, LGTM. Merging! 👍

PR #838, opened 2025-11-28
fix: Space Dock no longer counts towards planet fields (#713)

### LINKED ISSUE #713: [BUG] Space Dock is counted towards planet's fields
**Describe the bug**
Each level of Space Dock adds +1 to the planet's fields. This shouldn't be the case, per its description.

https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L703
Since Space Dock is a StationObject and this goes through all of them, perhaps there should be a condition to exempt it like so:

```php
if ($object->machine_name !== 'space_dock' && $this->planet->{$object->machine_name} > 0) {
```

nguyenthanhan:
## Description

Fixes bug where Space Dock incorrectly counted towards planet fields. Space Dock "floats in orbit" per its description and should not consume a field.

**Solution:** Added `consumesPlanetField` property to `GameObject` class (defaults to `true`), set to `false` for Space Dock. Updated field calculation logic to respect this property.

### Type of Change:

- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #713

## Checklist

- [x] **Code Standards:** Code adheres to PSR-12 coding standards.
- [x] **Testing:** Unit tests added and passing locally.
- [x] **CSS & JS Build:** N/A - no frontend changes.
- [x] **Documentation:** N/A - code is self-documenting.

## Additional Information

**Test Results:** All PlanetServiceTest tests pass (9/9). New test verifies Space Dock doesn't count towards fields.

**Example:** Planet with Metal Mine (10) + Space Dock (3) = 10 fields used (not 13).

### DIFF
--- a/app/GameObjects/Models/Abstracts/GameObject.php
+++ b/app/GameObjects/Models/Abstracts/GameObject.php
@@ -69,6 +69,14 @@ abstract class GameObject
     public GameObjectAssets $assets;
 
     /**
+     * Whether this object consumes a planet field when built. Most buildings consume a field,
+     * but some (like Space Dock which floats in orbit) do not.
+     *
+     * @var bool
+     */
+    public bool $consumesPlanetField = true;
+
+    /**
      * Custom calculation formulas for this object. These formulas can be used to calculate custom values for the object.
      * E.g. a formula to determine max amount of planets that can be colonized with astrophysics technology.
      *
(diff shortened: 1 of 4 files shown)


lanedirt:
Nice work, thanks for following the suggestion for implementation. Looks good! 😄 Merging!

PR #841, opened 2025-11-28
fix: add separate fuel capacity for ships to fix espionage probe carg…

### LINKED ISSUE #206: Add option to disable cargo space for espionage probes
Currently the `Espionage Probe` GameObject ship is defined with a cargo capacity of 5.

See:
- `app/GameObjects/CivilShipObjects.php`

However in the official game a restriction has been added where the cargo space of espionage probes is set to 0 when they are used to attack players or transport goods. This is to prevent a game tactic where espionage probes are used to rapidly attack players or transport goods because of their enormous speeds.

**The challenge:**
Currently the codebase only supports "cargo capacity". However, when we would set the "cargo capacity" of espionage probes to 0, it causes issues when dispatching fleets because of deuterium consumption. In a specific part of the fleet dispatch system a second option is mentioned: "fuel capacity". When checking what the official game uses for espionage probes it is seen that it has these values:

- baseFuelCapacity: 5
- baseCargoCapacity: 0

So basically it means we will need to introduce a second config option "fuel capacity" in addition to "cargo capacity".

See `FleetController::dispatchCheckTarget()` which includes a return value for `baseFuelCapacity`. The fleet dispatch screen requires this parameter.

nguyenthanhan:
## Description

Introduces a separate `fuel_capacity` property for ships to fix the espionage probe exploit. Espionage probes now have cargo capacity = 0 (cannot transport resources) but fuel capacity = 5 (can still fly). Fully backward compatible - existing ships default fuel capacity to cargo capacity.

### Type of Change:
- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #206

## Checklist

- [x] **Testing:** Relevant unit and feature tests are included or updated. Tests successfully run locally.
- [x] **CSS & JS Build:** N/A - No JS/CSS changes.
- [x] **Documentation:** N/A - Code is self-documenting with comprehensive tests.

## Additional Information

**Changes:**
- Added `FuelCapacityPropertyService` for fuel capacity calculations
- Updated `GameObjectProperties` with fuel_capacity property (defaults to capacity for backward compatibility)
- Espionage probe: cargo=0, fuel=5
- Fleet validation now uses fuel capacity for deuterium checks

### DIFF
--- a/app/GameMissions/Abstracts/GameMission.php
+++ b/app/GameMissions/Abstracts/GameMission.php
@@ -202,9 +202,10 @@ abstract class GameMission
         $this->startMissionSanityChecks($planet, $targetCoordinate, $targetType, $units, $deduct_resources);
 
         $totalCargoCapacity = $units->getTotalCargoCapacity($planet->getPlayer());
+        $totalFuelCapacity = $units->getTotalFuelCapacity($planet->getPlayer());
 
         // Check if the player has sufficient deuterium storage capacity for the fleet.
-        if ($totalCargoCapacity < $consumption) {
+        if ($totalFuelCapacity < $consumption) {
             throw new Exception(__('You don\'t have sufficient storage capacity!'));
         }
 
--- a/app/GameObjects/CivilShipObjects.php
+++ b/app/GameObjects/CivilShipObjects.php
@@ -150,7 +150,7 @@ As soon as Impulse Drive research has reached level 17, Recyclers are refitted w
             new GameObjectRequirement('espionage_technology', 2),
         ];
         $espionageProbe->price = new GameObjectPrice(0, 1000, 0, 0);
-        $espionageProbe->properties = new GameObjectProperties($espionageProbe, 1000, 0, 0, 100000000, 5, 1);
+        $espionageProbe->properties = new GameObjectProperties($espionageProbe, 1000, 0, 0, 100000000, 0, 1, 5);
 
         $espionageProbe->assets = new GameObjectAssets();
         $espionageProbe->assets->imgSmall = 'espionage_probe_small.jpg';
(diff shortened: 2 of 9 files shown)


Geda173:
Hi @nguyenthanhan,

an Espionage Probe travelling from Galaxy 1 to Galaxy 5 (as an example) needs 9 Deuterium to be dispatched. Your options are either: decresing the speed and thus the Deuterium consumption, or increase the cargo capacity by researching Hyperspace Technology.

Long story short, does Hyperspace Technology increase `fuel capacity`? How did you end up on the number 5? (I am assuming that is the standard `cargo capacity` for Espionage Probes but I haven't checked.)

anngth:
hi @Geda173,
Yes! In my implementation, `fuel_capacity` uses the same bonus calculation as `cargo_capacity` - it gets +5% per Hyperspace Technology level. So with Hyperspace Tech level 4, fuel capacity would be: `5 + (5 × 4 × 5%) = 6`.

See `FuelCapacityPropertyService.php`:
```php
protected function getBonusPercentage(PlayerService $player): int
{
    $hyperspace_technology_level = $player->getResearchLevel('hyperspace_technology');
    return 5 * $hyperspace_technology_level;
}
```

The number 5 comes directly from the issue description (#206), which states the official game uses:
- `baseFuelCapacity: 5`
- `baseCargoCapacity: 0`

This matches the original cargo capacity value (5) that espionage probes had before this change, ensuring they can still fly the same distances as before while preventing resource transport exploit.

Geda173:
Awesome, good thinking! It may be outside the scope of this PR, but there are OGame universes, in which Espionage Probes DO have cargo capacity and they are used for raiding and transporting. That is probably something for the server settings, in which an admin should be able to choose between these two options: "regular" Espionage Probes or the hyperaggressive ones.

anngth:
Thanks for the feedback! You raise a valid point about different universe configurations.

Adding a server setting to toggle between "regular" (cargo=0) and "hyper-aggressive" (cargo=5) espionage probes would be a great enhancement. The current `fuel_capacity` implementation helps facilitate this future feature.

I’ll keep this pull request focused on separating fuel and cargo capacity. If you want to add the server setting feature, feel free to open a new issue for it!

lanedirt:
Thanks for the work on this, LGTM. Merging!

PR #847, opened 2025-11-30
fix: optimize high score calculation using geometric series formula #732

### LINKED ISSUE #732: High score calculation optimization
**Is your feature request related to a problem? Please describe.**
https://github.com/lanedirt/OGameX/blob/main/app/Services/PlanetService.php#L1895

For score calculation, particularly for buildings, the above loop is used to calculate and sum up every level from 1 to the building's current level. That is the most accurate way, but maybe a bit slow, especially if there's a lot of players. There is a formula to approximate the sum very closely, well under 1000, which is good enough because you will divide by at least 1000 anyway.

`FLOOR(<base> * (1 - <factor>**<level>) / (1 - <factor>))`

You can test this on a spreadsheet. Most variance happen on the lower levels, and gets more accurate at very high levels due to significant digits. But the difference between the actual sum and the value from that formula should be well under 1000 points, probably within 100 or 200 points. If added up, amongst 3 values (metal, crystal, deuterium), it MAY be off by 1 total highscore point, but that's insignificant. I think it should be safe and for mass updates, and this should be way quicker than the current brute-force method. Potentially, can run the highscore update scheduler more frequent than every 5 minutes as well.

Just a suggestion!

nguyenthanhan:
## Description
Optimizes high score calculation by replacing iterative loops with a geometric series formula. This significantly improves performance when calculating scores for players with high-level buildings, especially beneficial for mass score updates.

### Type of Change:
- [x] Feature enhancement

## Related Issues
Fixes #732

## Checklist
- [x] **Testing:**
  - [x] Relevant unit and feature tests are included or updated.
  - [x] Tests successfully run locally (17 tests, 76 assertions passed).
- [x] **CSS & JS Build:** N/A - No changes to JS/CSS files.
- [x] **Documentation:** N/A - Internal optimization, no user-facing changes.

### DIFF
--- a/app/Services/ObjectService.php
+++ b/app/Services/ObjectService.php
@@ -582,6 +582,87 @@ class ObjectService
     }
 
     /**
+     * Gets the cumulative cost of all levels from 1 to the specified level for a building or research.
+     * Uses geometric series formula for O(1) performance instead of iterative O(n) calculation.
+     *
+     * @param string $machine_name
+     * @param int $level
+     * @return Resources (with energy excluded)
+     */
+    public static function getObjectCumulativeCost(string $machine_name, int $level): Resources
+    {
+        try {
+            $object = self::getObjectByMachineName($machine_name);
+        } catch (Exception $e) {
+            return new Resources(0, 0, 0, 0);
+        }
+
+        if ($object->type !== GameObjectType::Building &&
+            $object->type !== GameObjectType::Station &&
+            $object->type !== GameObjectType::Research) {
+            return new Resources(0, 0, 0, 0);
+        }
+
+        if ($level === 0) {
+            return new Resources(0, 0, 0, 0);
+        }
+
+        if ($level === 1) {
+            $base_price = $object->price;
+            $metal = floor($base_price->resources->metal->get());
+            $crystal = floor($base_price->resources->crystal->get());
+            $deuterium = floor($base_price->resources->deuterium->get());
+
+            return new Resources($metal, $crystal, $deuterium, 0);
+        }
+
+        $base_price = $object->price;
+        $factor = $base_price->factor;
+
+        $metal = self::calculateCumulativeCostForResource(
+            $base_price->resources->metal->get(),
+            $factor,
+            $level
+        );
+
+        $crystal = self::calculateCumulativeCostForResource(
+            $base_price->resources->crystal->get(),
+            $factor,
+            $level
+        );
+
+        $deuterium = self::calculateCumulativeCostForResource(
+            $base_price->resources->deuterium->get(),
+            $factor,
+            $level
+        );
+
+        return new Resources($metal, $crystal, $deuterium, 0);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


lanedirt:
Hi @nguyenthanhan,

Thanks for these changes. I have locally run it against a database with 2k test users with various levels to see what the performance difference is.

This is the original implementation (main branch):
```bash
root@aac716ee72e9:/var/www# php artisan ogamex:generate-highscores
Updating highscores...
 2160/2160 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
All highscores completed in 307.30 seconds.
```

This is the new implementation (this PR):
```bash
root@aac716ee72e9:/var/www# php artisan ogamex:generate-highscores
Updating highscores...
 2160/2160 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%
All highscores completed in 247.23 seconds.
```

So the difference is about 20% quicker. Does this 20% improvement match what you would expect?

I also see some minor differences in final scores. The biggest being: `561,636,979` (old) vs `561,636,763` (new). However since this is such a small difference and the calculation logic is the same for every player which makes it "fair", I think this is an acceptable issue.

anngth:
Hi @lanedirt,

Thank you for testing! The 20% improvement aligns with my expectations. Please note that the optimization only impacts the calculation portion of the command. There is still time spent on database I/O, sorting, and updates, which are not optimized in this case.

For pure calculations involving high-level buildings, the improvement could reach up to 50-80%. However, when averaged across all operations and various building levels, a 20% improvement is a solid result.

Regarding score differences: a difference of 216 points is well within the specified tolerance of less than 1000 in #732. As you mentioned, it's consistent and fair for all players.

psimoesbits:
Hi @nguyenthanhan 
Nice work on this PR! I wonder if there would be extra performance gains by caching the method's `ObjectService::getObjectCumulativeCost($object->machine_name, $level)`  return in memory. 
The result will always be the same for the pair `{ machine_name, level }` if I am not mistaken.

anngth:
Hi @psimoesbits,
Thanks for the suggestion! Caching could definitely be beneficial, especially if we repeatedly calculate the same building levels.
I focused this PR solely on replacing the loop with the formula to maintain simplicity. The formula itself is already quite efficient, so I'm uncertain about the additional benefits we might gain from caching, but it could be worth testing!
This could potentially be addressed in a new PR.

lanedirt:
> I focused this PR solely on replacing the loop with the formula to maintain simplicity. The formula itself is already quite efficient, so I'm uncertain about the additional benefits we might gain from caching, but it could be worth testing!
This could potentially be addressed in a new PR.

Agreed! As the current solution is more performant than the original implementation, this is a good start. Additional PR's to potentially improve performance even further are always welcome.

Thanks again @nguyenthanhan! Merging!

PR #860, opened 2025-12-04
fix: expedition return trip timing to include holding time #729

### LINKED ISSUE #729: [BUG] Expedition return trip arrival time is calculated incorrectly
**Describe the bug**
Expedition's return trip, after the hold, may complete instantly without return timing unless there's an expedition delay long enough to offset. The time is still wrong, but the mission will not complete instantly after hold.

https://github.com/lanedirt/OGameX/blob/main/app/GameMissions/Abstracts/GameMission.php#L291
- The start time is set to the parent's end.
- The end time is set to the travel duration +/- delay/haste time.

The issue with this logic for expedition is the lack of hold time inclusion. Expedition hold times are at least 1h long, and this value is stored separately. Setting the return mission start time to be the parent's end time result in its value being in the past by the hold time amount.

Reference for auto-completion:
https://github.com/lanedirt/OGameX/blob/main/app/GameMissions/Abstracts/GameMission.php#L352

Side note: the fleet event listing does show the expedition trips and timing properly.

**Solution Idea**
Add the expedition hold time to the parent missions's end time as the start time for the return trip.

nguyenthanhan:
## Description

This PR fixes a bug where expedition return trips would complete instantly after the expedition arrival, bypassing the required holding time. The issue occurred because the return mission's start time was calculated using only the parent mission's arrival time, without including the expedition's holding time (which is at least 1 hour).

### Type of Change:

- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #729

## Checklist

- [x] **Code Standards:** All 297 files pass PSR-12 check
- [x] **Static Analysis:** PHPStan analysis completed with no errors
- [x] **Testing:** Relevant unit and feature tests are included or updated.

### DIFF
--- a/app/GameMissions/Abstracts/GameMission.php
+++ b/app/GameMissions/Abstracts/GameMission.php
@@ -332,8 +332,9 @@ abstract class GameMission
 
         // No need to check for resources and units, as the return mission takes the units from the original
         // mission and the resources are already delivered. Nothing is deducted from the planet.
-        // Time this fleet mission will depart (arrival time of the parent mission)
-        $time_start = $parentMission->time_arrival;
+        // Time this fleet mission will depart (arrival time of the parent mission + holding time if applicable)
+        // For expeditions, the holding time must be included as the mission doesn't complete until after the hold.
+        $time_start = $parentMission->time_arrival + ($parentMission->time_holding ?? 0);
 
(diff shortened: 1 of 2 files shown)


lanedirt:
Thanks for the fix! Merging!

Issue #435, opened 2024-11-01
Decrease building & research levels button


alican22:
Original ogame has decrease buildings and research by tear down button. 

When you decrease the building or research you will get refund 62.5% resources amount.

![Screenshot 2567-11-01 at 23 21 33](https://github.com/user-attachments/assets/fd140cb7-ce68-438d-a9a7-c41f47c126ba)

lanedirt:
Hi @alican22,

Thanks for reporting this issue. There already is an issue created for this however, see:
- #110 

Also you mention that a refund of 62.5% resources is given. I actually thought that in the official game it costs resources to downgrade. I.e. you don't get anything back but instead you have to pay more. But I might have that wrong, haven't really used this functionality when I played the game.

Can you add any comments or extra context you might have to the original issue above?

Closing this as duplicate.

alican22:
> * Implement building downgrade/tear down functionality #110

Hi @lanedirt 

Absulately you are right. It's not refund and resource. It has to be pay more 62.5% more from the level that you upgraded.

PR #874, opened 2025-12-07
feat: implement building downgrade functionality including cost calcu…

### LINKED ISSUE #110: Implement building downgrade/tear down functionality

nguyenthanhan:
## Description

This PR implements the building downgrade/tear-down functionality, allowing players to reduce building levels by one at a cost. This feature matches OGame's traditional mechanics, where downgrading requires resources (62.5% of the upgrade cost) and takes time equal to the original construction time.

### Type of Change:

- [ ] Bug fix
- [x] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #110

## Checklist

- [x] **Testing:**

### Formula Details:
- Downgrade Cost: 62.5% of the upgrade cost (cost to build from level-1 to the current level)
- Ion Technology Bonus: 1% cost reduction per Ion technology level (e.g., level 24 = -24% cost)
- Downgrade Duration: 100% of construction time (same as building from level-1 to current level)

### DIFF
--- a/app/Http/Controllers/Abstracts/AbstractBuildingsController.php
+++ b/app/Http/Controllers/Abstracts/AbstractBuildingsController.php
@@ -193,7 +193,7 @@ abstract class AbstractBuildingsController extends OGameController
     {
         // Check if this is a downgrade request (mode 3)
         $mode = $request->input('mode');
-        if ($mode == 3) {
+        if ((int)$mode === 3) {
             return $this->downgradeBuildRequest($request, $player);
         }
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -65,7 +65,7 @@ Route::middleware(['auth', 'globalgame', 'locale'])->group(function () {
     Route::get('/facilities', [FacilitiesController::class, 'index'])->name('facilities.index');
     Route::get('/ajax/facilities', [FacilitiesController::class, 'ajax'])->name('facilities.ajax');
     Route::get('/facilities/add-buildrequest', [FacilitiesController::class, 'addBuildRequest'])->name('facilities.addbuildrequest');
-    Route::post('/facilities/add-buildrequest', [FacilitiesController::class, 'addBuildRequest'])->name('facilities.addbuildrequest.get');
+    Route::post('/facilities/add-buildrequest', [FacilitiesController::class, 'addBuildRequest'])->name('facilities.addbuildrequest.post');
     Route::post('/facilities/downgrade', [FacilitiesController::class, 'downgradeBuildRequest'])->name('facilities.downgrade');
     Route::post('/facilities/cancel-buildrequest', [FacilitiesController::class, 'cancelBuildRequest'])->name('facilities.cancelbuildrequest');
 
--- a/app/Services/BuildingQueueService.php
+++ b/app/Services/BuildingQueueService.php
@@ -47,13 +47,13 @@ class BuildingQueueService
     public function retrieveQueueItems(PlanetService $planet): Collection
     {
         // Fetch queue items from model
+        // For items not yet started (time_start = 0), order by ID to maintain queue order
         return BuildingQueue::where([
             ['planet_id', $planet->getPlanetId()],
             ['processed', 0],
             ['canceled', 0],
         ])
-            ->orderBy('time_start', 'asc')
-            ->orderBy('id', 'asc') // Secondary sort by ID for items not yet started
+            ->orderBy('id', 'asc') // Primary sort by ID (insertion order)
             ->get();
     }
 
@@ -95,13 +95,8 @@ class BuildingQueueService
         foreach ($queue_items as $item) {
             $item_object = ObjectService::getObjectById($item->object_id);
             if ($item_object->machine_name === $building->machine_name) {
-                if ($item->is_downgrade ?? false) {
-                    // Downgrade: level decreases
-                    $level_after_queue = $item->object_level_target;
-                } else {
-                    // Upgrade: level increases
-                    $level_after_queue = $item->object_level_target;
-                }
+                // Each item sets the level to its target (simulating queue execution)
+                $level_after_queue = $item->object_level_target;
             }
         }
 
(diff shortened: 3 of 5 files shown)


Geda173:
Hi @nguyenthanhan,

thanks for the PR. Where are you getting your information/formulas from?

- Ion Technology reduces teardown cost by 4% per level, not 1%
- The cost for teardown is not 65% of the original cost. The cost of the teardown, without any bonuses applied, is actually the construction cost of the level as if you were building it right now: Cost = Construction x - 1

anngth:
Hi @Geda173, thank you for the review and for pointing out the incorrect formulas. I will update the code in 2 points:

1. Downgrade Cost Formula:
* Before: 62.5% of construction cost (referenced from this issue: [https://github.com/lanedirt/OGameX/issues/435](https://github.com/lanedirt/OGameX/issues/435))
* After: 100% of construction cost (the cost to build from `level-1` to `current level`)

2. Ion Technology Bonus:
* Before: 1% reduction per level (I think so)
* After: 4% reduction per level

lanedirt:
Hi @nguyenthanhan, thanks for your work on implementing this feature. Structure looks good! Also thanks @Geda173 for the logic review! I just tested it locally and did find a couple of minor UI things:

1. When downgrading a building, it now shows `Improve to`:
image

But on the official game when downgrade is active, it shows `Downgrade to`:
image


2. When downgrading a building, it shows target level "3" inside the object icon (i.e., the next level), but it shows target level "1" in the queue. Both should say "1":
image

3. When downgrading, in the sidebar it currently shows the normal "yellow" build icon:
image

On the official game it actually shows a "red" build icon instead:
image

4. Downgrading "Facilities" is not yet possible in OGameX:
image


While in the official game it is:
image

anngth:
Hi @lanedirt, I have fixed the bugs. Please check again.

app/Http/Controllers/Abstracts/AbstractBuildingsController.php
```diff
@@ -187,6 +191,12 @@ public function indexPageParams(Request $request, PlayerService $player): array
      */
     public function addBuildRequest(Request $request, PlayerService $player): JsonResponse
     {
+        // Check if this is a downgrade request (mode 3)
+        $mode = $request->input('mode');
+        if ($mode == 3) {
```
Geda173: We could use `=== ` here for strict type checking to avoid unexpected behavior. `$mode == 3` currently matches `3`, `"3"`, `3.0`, and even `true`.

routes/web.php
```diff
@@ -58,13 +58,15 @@
...
     Route::get('/resources/add-buildrequest', [ResourcesController::class, 'addBuildRequest'])->name('resources.addbuildrequest');
     Route::post('/resources/add-buildrequest', [ResourcesController::class, 'addBuildRequest'])->name('resources.addbuildrequest.post');
+    Route::post('/resources/downgrade', [ResourcesController::class, 'downgradeBuildRequest'])->name('resources.downgrade');
     Route::post('/resources/cancel-buildrequest', [ResourcesController::class, 'cancelBuildRequest'])->name('resources.cancelbuildrequest');
 
     // Facilities
     Route::get('/facilities', [FacilitiesController::class, 'index'])->name('facilities.index');
     Route::get('/ajax/facilities', [FacilitiesController::class, 'ajax'])->name('facilities.ajax');
     Route::get('/facilities/add-buildrequest', [FacilitiesController::class, 'addBuildRequest'])->name('facilities.addbuildrequest');
     Route::post('/facilities/add-buildrequest', [FacilitiesController::class, 'addBuildRequest'])->name('facilities.addbuildrequest.get');
```
Geda173: In the Resources Controller you correctly labelled all the `POST` routes as such, but here you label them as `.get`.  While this technically works, we should make sure it is uniform for maintainbility.

app/Http/Controllers/Abstracts/AbstractBuildingsController.php
```diff
@@ -216,6 +226,41 @@ public function addBuildRequest(Request $request, PlayerService $player): JsonRe
...
+            $this->queue->addDowngrade($player->planets->current(), $building_id);
+
+            return response()->json([
+                'status' => 'success',
+                'message' => 'Building downgrade started.',
+            ]);
+        } catch (Exception $e) {
+            return response()->json([
+                'success' => false,
+                'errors' => [['message' => $e->getMessage()]],
```
Geda173, line 259: The success and error responses use different fields (`status` vs `success`). Using one consistent field, e.g.  `success: true/false`, would enable the frontend to handle all responses the same way. This would also align the downgrade endpoint with the existing build/upgrade endpoints elsewhere in the codebase, which already use a `success` flag in their JSON responses.

app/Http/Controllers/Abstracts/AbstractBuildingsController.php
```diff
@@ -211,11 +221,46 @@ public function addBuildRequest(Request $request, PlayerService $player): JsonRe
         $this->queue->add($player->planets->current(), $building_id);
 
         return response()->json([
-            'status' => 'success',
+            'success' => true,
             'message' => 'Building construction started.',
         ]);
```
lanedirt, line 226: I think this change has caused some regressions. Now in the frontend when trying to upgrade a building, nothing happens. The original OGame frontend JS requires the previous implementation with `'status' => 'success'`.

I do agree with the comments of @Geda173 before that streamlining response values would be nice, but this will require us changing all the original OGame JS too. So in this case I suggest changing this back to the original value. We can always clean this up in a follow-up PR if we want, which  allows us to better focus on potential regression bugs.

app/Http/Controllers/Abstracts/AbstractBuildingsController.php
lanedirt: Here I think my comment from above also applies: with this version when doing a downgrade, no confirm dialog is shown in the interface anymore.

lanedirt (approved):
Thanks for your work and for the adjustments, looks to be working good locally now, awesome! Merging!

PR #885, opened 2025-12-08
Add WithCachedConfig / WithCachedRoutes to speed up tests

jackbayliss:
This adds WithCachedConfig / WithCachedRoutes trait to the test case to speed up tests. 

This was added in [12.38.0](https://github.com/laravel/framework/releases/tag/v12.38.0) 

Essentially this just speeds up tests because we don't tap into the config / routes. Mainly noticeable locally, but should give us a bit of speed in CI too. 

P.S sorry I haven't been too active, I don't play the game- but I know Laravel so popping in to do some Laravel bits 🫡


lanedirt:
@jackbayliss Looks good! Always nice to have a separate pair of eyes for modern Laravel features! 😄

PR #888, opened 2025-12-08
Update run-phpstan-code-analysis.yml

jackbayliss:
This uses v2 setup-php, as there's no need to use a specific commit afaik, 

This also ensures we don't have coverage, as we don't need it. 

I've also run stan using our composer script- rather than repeat it :)


lanedirt:
> This uses v2 setup-php, as there's no need to use a specific commit afaik,

Yes I also don't really remember why a specific commit was used there, maybe I've taken that from an example somewhere. Thanks for the update, looks good!

PR #890, opened 2025-12-10
fix: prevent building/research/ship construction in vacation mode #884

### LINKED ISSUE #884: [BUG] When player is in vacation mode building can still be upgraded
**Describe the bug**
When player has enabled vacation mode, they are still able to upgrade buildings, start research and build ships. In the official game, all these actions are blocked, example:

Image

Image

A secondary bug that occurs because of this: when a player that has vacation mode enabled and tries to build something, once the build queue is finished the page wil automatically reload (which is normal), but because the actual processing of the build queue is blocked, the page will become stuck in a reload loop because of the if() check in this function:
```php
/**
 * Update this planet's buildings by checking the build queue.
 * This should happen on every users page load and every time the planet is touched.
 *
 * @param bool $save_planet
 *  Optional flag whether to save the planet in this method. This defaults to TRUE
 *  but can be set to FALSE when update happens in bulk and the caller method calls
 *  the save planet itself to prevent on unnecessary multiple updates.
 *
 * @throws Exception
 */
public function updateBuildingQueue(bool $save_planet = true): void
{
    // Skip building queue processing if player is in vacation mode
    if ($this->getPlayer()->isInVacationMode()) {
        return;
    }
    [...]
 }
```

nguyenthanhan:
## Description

This PR fixes a bug where players in vacation mode could still upgrade buildings, start research, and build ships. In the official OGame, all these actions should be blocked when a player is in vacation mode.

### Type of Change:

- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #884

## Checklist

- [x] **Testing:** All existing vacation mode tests still pass.
    - Tests successfully run locally
  - No changes to JS/CSS files in this PR
  - No documentation changes required for this bug fix

### DIFF
--- a/app/Http/Controllers/Abstracts/AbstractUnitsController.php
+++ b/app/Http/Controllers/Abstracts/AbstractUnitsController.php
@@ -124,6 +124,7 @@ abstract class AbstractUnitsController extends OGameController
             'build_queue' => $build_queue,
             'build_queue_countdown' => $queue_time_countdown,
             'open_tech_id' => $open_tech_id,
+            'is_in_vacation_mode' => $player->isInVacationMode(),
         ];
     }
 
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -1116,6 +1116,13 @@ class PlanetService
         // NOTE: this issue can be circumvented with a continious job runner which can
         // update all planets periodically...
 
+        // If time_last_update is 0 or null, initialize it to current time to prevent
+        // calculating resources from epoch (1970) which would cause massive resource gain.
+        if ($time_last_update <= 0) {
+            $time_last_update = $current_time;
+            $this->planet->time_last_update = $current_time;
+        }
+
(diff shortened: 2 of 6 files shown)


lanedirt:
Hi @nguyenthanhan, thanks for looking into this bug, I appreciate it!

I just tested the changes and have two comments for full coverage:

1. The shipyard should also show the disabled state in the icons when vacation mode is active:
image

Currently it shows like this in OGameX, so build button is disabled (which is good) but icons are shown as normal:
image

This applies to the `Defense` page as well.

2.  The `tear down` button should also be disabled. It's currently showing as active. Clicking it though does not actually start deconstruction, so that's good.
image

It should instead look like this:
image

anngth:
Hi @lanedirt, I’ve fixed the issues. Please check again.

lanedirt:
Looks good, thanks for the adjustments! Merging!

PR #932, opened 2025-12-23
Update README.md

jackbayliss:
Noticed some of the shields were not showing in the readme.

image

This changes the build / test shields.


jackbayliss:
Stan CI seems flaky, so updated that in a diff PR.

PR #933, opened 2025-12-23
Update run-phpstan-code-analysis.yml

jackbayliss:
I noticed this seems to be running weird couple things I noticed: 

1. We only need the dev composer installed, as its static analysis 
2. We dont need to generate anything,  or adjust any storage bits re keys - laravel etc.
3. Added cache to make it even quicker


lanedirt:
@jackbayliss thanks for looking into this and the improvements! Stan has indeed been flaky as of lately in CI. Strangely as it seemed to have worked stable in the beginning. I tried playing with the memory size but it kept throwing timeouts sometimes.

PR #950, opened 2025-12-29
fix: exclude nanite factory from self-benefit in build time

### LINKED ISSUE #940: [BUG] Nanite Factory build time incorrect
**Describe the bug**
When upgrading the nanite factory, the building time increases/doubles with every level.

**Expected behavior**
The doubling of the building time should be counteracted by the halving the nanite factory applies to build times. In practice this means that every level of nanite factory should actually always have the same build time.

nguyenthanhan:
## Description

Fixed the incorrect Nanite Factory build time calculation. The Nanite Factory was incorrectly benefiting from its own level when calculating construction time, which is not how it should work according to the OGame mechanics.

### Type of Change:

- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues

Fixes #940

## Checklist

- [x] **Testing:**

## Additional Information

**Changes Made:**

### DIFF
--- a/app/Services/PlanetService.php
+++ b/app/Services/PlanetService.php
@@ -776,13 +776,24 @@ class PlanetService
             $universe_speed = 1;
         }
 
-        // The actual formula which return time in seconds
-        $time_hours =
-            (
-                ($price->metal->get() + $price->crystal->get())
-                /
-                (2500 * max((4 - ($next_level / 2)), 1) * (1 + $robotfactory_level) * $universe_speed * (2 ** $nanitefactory_level))
-            );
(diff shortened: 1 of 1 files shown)


Geda173:
The Nanite Factory *should* be benefitting from itself! The build time for it shoudl be static, as the next level doubles the amount of time needed, but it also beneftis from itself, halving the build time! With this PR, the build time simply doubles, which is incorrect.

Vanilla build time Nanite Factory 6:
image

Vanilla build time Nanite Factory 7:
image

anngth:
I continue to fix in #951

PR #986, opened 2025-12-31
Fix flaky fleet dispatch tests

### LINKED ISSUE #976: [BUG] Flaky fleet dispatch tests: Test helper randomly selects admin users causing failures
**Describe the bug**
Fleet dispatch tests have been randomly failing recently and I think I have tracked down the cause.

The `getNearbyForeignPlanet() `helper in `AccountTestCase.php` is supposed to find a random nearby foreign planet for testing. The problem is it sometimes picks planets owned by admin users (Legor), which breaks tests in unexpected ways:

- Some tests try to send buddy requests to these planets → fails because you can't send buddy requests to admins
- Some tests expect certain missions to be available → fails because the target being a buddy/admin changes which missions are allowed

The failures are random because the query uses `inRandomOrder()`, so sometimes you get a regular user and sometimes you get an admin.

There's already a check to exclude "Legor" by username, but that's not enough because:
- Admin status is role-based (using `hasRole('admin')`)
- Any user can be assigned the admin role
- The tests don't reset the database between runs, so admin users created during testing stick around

The helper needs to actually check the roles tables, not just filter by username.

Alternative approaches:

1. Add `RefreshDatabase` trait to the test base classes - this would reset the DB between tests but might make tests slower
2. Create dedicated test planets instead of randomly selecting from existing ones
3. Keep a whitelist of "safe" test users instead of trying to exclude admins

Geda173:
## Description
The `getNearbyForeignPlanet()` and `getNearbyForeignMoon()` helpers were randomly selecting planets owned by admin users, causing test failures when missions like buddy requests or attacks were attempted against admin-protected planets.
Modified both helpers to check the `model_has_roles `table to properly exclude users with the 'admin' role, rather than relying on username filtering alone.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #976 

## Checklist

- [X] **Testing:**

## Additional Information

### DIFF
--- a/tests/AccountTestCase.php
+++ b/tests/AccountTestCase.php
@@ -5,7 +5,6 @@ namespace Tests;
 use Carbon\Carbon;
 use Exception;
 use Illuminate\Contracts\Container\BindingResolutionException;
-use Illuminate\Support\Facades\Cache;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 use Illuminate\Testing\TestResponse;


tests/AccountTestCase.php
```diff
@@ -216,13 +216,20 @@ protected function getNearbyForeignPlanet(): PlanetService
...
             ->where('galaxy', '<=', $maxGalaxies)
             ->where('planet_type', PlanetType::Planet)
             ->whereBetween('system', [$this->planetService->getPlanetCoordinates()->system - 15, $this->planetService->getPlanetCoordinates()->system + 15])
+            ->whereNotIn('user_id', function ($query) {
+                $query->select('model_id')
+                    ->from('model_has_roles')
+                    ->join('roles', 'roles.id', '=', 'model_has_roles.role_id')
+                    ->where('roles.name', 'admin')
+                    ->where('model_has_roles.model_type', 'OGame\\Models\\User');
+            })
```
jackbayliss:
```suggestion
            ->whereNotIn('user_id', function ($query) {
               Cache::memo()->rememberForever('admins-tests', function () {
                    return User::role('admin')->pluck('id');
               });
            })
```



We can change all these calls to something simpler - this means it won't be a query for all of them, it'll sit in memory as it's pointless requerying it multiple times 👍🏻  Just apply this everywhere you need it. Spatie permissions has a role scope so no need to join and figure it out ourselves.

Geda173: Thanks for the suggestion, I have applied it. Unfortunately we are now failing the PHP Static Analysis. I'll have to clean it up by hand later.

jackbayliss: I can adjust it, I think 🤔 I'll have a go

Geda173: I think I got it, thanks for the suggestions and the offer to adjust it.

The main problem:
`whereNotIn()` with a closure expects you to build a subquery using the `$query` parameter, but the code was trying to use `Cache::memo()` inside the closure instead. This caused a few issues:

  1. `Cache::memo()` doesn't exist in Laravel (it's either `Cache::remember()` or the `once()` helper)
  2. The `$query` parameter wasn't being used at all
  3. The closure didn't return anything, so `whereNotIn()` had nothing to work with
  4. PHPStan couldn't find the `Cache` class because it wasn't imported

Instead of trying to cache inside the `whereNotIn() `closure, we need to cache the admin IDs outside and then pass the array to `whereNotIn()`. Here's what I implemented:

```
// Added a helper method to cache admin IDs
protected function getAdminUserIds(): array
{
    return once(function () {
        return \DB::table('model_has_roles')
            ->join('roles', 'roles.id', '=', 'model_has_roles.role_id')
             ->where('roles.name', 'admin')
            ->where('model_has_roles.model_type', 'OGame\\Models\\User')
            ->pluck('model_id')
            ->toArray();
    });
}

// Then use it like this:
->whereNotIn('user_id', $this->getAdminUserIds())
```

I used `once() `instead of `Cache::rememberForever()` because:
- It caches for the current request only (no cache pollution between tests)
- No need to manage cache invalidation
- Still gives us the performance benefit you were after

Your suggestion is the right way to go: making 4 different calls to achieve the same thing just bloats everything. The implementation just needed a small tweak to work with Laravel's query builder!

jackbayliss: @Geda173  Awesome, [Cache memo is brand new](https://laravel.com/docs/12.x/cache#cache-memoization) - but happy with your once implementation. Nice work.

tests/AccountTestCase.php
```diff
@@ -204,6 +205,23 @@ protected function getSecondPlayerId(): int
...
 
+    /**
+     * Get admin user IDs. Cached for the current request to avoid repeated queries.
+     *
+     * @return array
+     */
+    protected function getAdminUserIds(): array
+    {
+        return once(function () {
+            return \DB::table('model_has_roles')
```
jackbayliss:
```suggestion
            return DB::table('model_has_roles')
```

Can we import the DB facade here? :) Same with elsewhere - could be worth an issue if its done everywhere.

tests/AccountTestCase.php
Geda173: If it is not too much trouble, I'd be interested to see your solution!

jackbayliss: I would of ended up with something like: 
Once is also fine though, I would just use the role scope as it reads easier (to me anyway) 

The memo method means its only cached per request or job

```php 

protected function getAdminUserIds(): array
    {
        return Cache::memo()->rememberForever('admin_tables', function () {
            return User::role('admin')->pluck('id')->toArray();
        });
    }
```

tests/AccountTestCase.php
```diff
@@ -5,6 +5,8 @@
 use Carbon\Carbon;
 use Exception;
 use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Support\Facades\Cache;
```
jackbayliss:
```suggestion

```
This is now unused I believe :)

Geda173: You're right, just ran pint and it flagged it as well.

jackbayliss (approved):
Looks good to me, and makes sense.

lanedirt:
Thanks @Geda173 for the work on this, and thanks @jackbayliss for the review. Merging!

PR #993, opened 2026-01-01
Introduce new pint rules

jackbayliss:
This introduces two new pint rules 

```
    "no_unused_imports": true,
    "ordered_imports": true,
```

This means we don't end up with any unused imports, and they'll be ordered. 

I've applied it to the repo too.


lanedirt:
Good idea I agree, makes it more readable!

PR #994, opened 2026-01-01
Pint rule - Always import classes

jackbayliss:
Noticed we were using the global class in many places, this just leans on pint as IMO it fits with consistency of Laravel. 

But, up to you, I don't mind heavily- just means pint sorts it.


lanedirt:
@jackbayliss Looks good to me, having a sane default for everywhere makes the code more readable. There are however a few merge conflicts now, probably because of some other PR's that just got merged.

PR #998, opened 2026-01-01
Fix Flaky FleetDispatchAttackTest

### LINKED ISSUE #997: [BUG] Flaky Test: DispatchFleetReturnShown fails when run in full suite
**Describe the bug**
The test `tests/Feature/FleetDispatch/FleetDispatchAttackTest::testDispatchFleetReturnShown()` fails when run as part of the full test suite but passes when run in isolation.

Previous tests in the suite (likely ACS Defend tests) are creating state (buddy relationships, test planets) that persists and affects subsequent tests. The `getNearbyForeignPlanet()` method may be selecting a planet with unexpected state.

We likely need to clean up `FleetDispatchAcsDefendTest` to ensure buddy relationships and test planets don't leak into subsequent tests.

Geda173:
## Description
`FleetDispatchAttackTest::testDispatchFleetReturnShown()` (and other attack tests) failed when run as part of the full test suite but passed in isolation.

- `FleetDispatchAcsDefendTest` created buddy users/planets and set vacation mode during tests
- This state persisted in the database after tests completed
- `getNearbyForeignPlanet()` randomly selected these test planets for attack tests

Attack missions failed because:
- Buddy planets had unexpected state from previous tests
- Vacation mode prevented attacks (can't attack vacation mode players)

**Solution**:
Added `tearDown()` method to `FleetDispatchAcsDefendTest` that cleans up test state:
- Removes buddy_requests entries (prevents buddy relationship leakage)
- Resets vacation_mode to false 
- Preserves test users/planets in database to avoid full database reset

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #997 

## Checklist

- [X] **Testing:**

### DIFF
--- a/tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php
+++ b/tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\Feature\FleetDispatch;
 
+use Illuminate\Support\Facades\DB;
 use OGame\GameObjects\Models\Units\UnitCollection;
 use OGame\Models\Enums\PlanetType;
 use OGame\Models\Message;
@@ -88,6 +89,9 @@ class FleetDispatchAcsDefendTest extends FleetDispatchTestCase
      * Clean up test data after each test to prevent state leakage.
      * Only removes buddy relationships and resets vacation mode - test users and planets
      * remain in database but won't have special state that affects subsequent tests.
+     *
+     * @todo Refactor test architecture to support DatabaseTransactions/RefreshDatabase
+     *       by removing dependency on reloadApplication() expecting persisted users.
      */
     protected function tearDown(): void
     {
@@ -97,7 +101,7 @@ class FleetDispatchAcsDefendTest extends FleetDispatchTestCase
             $buddyUserId = array_shift(self::$allCreatedBuddyUserIds);
 
             // Delete buddy requests involving this user (with proper SQL grouping)
-            \Illuminate\Support\Facades\DB::table('buddy_requests')
+            DB::table('buddy_requests')
                 ->where(function ($query) use ($buddyUserId) {
                     $query->where('sender_user_id', $buddyUserId)
                         ->orWhere('receiver_user_id', $buddyUserId);
@@ -106,7 +110,7 @@ class FleetDispatchAcsDefendTest extends FleetDispatchTestCase
 
             // Reset all vacation mode fields for buddy user
             // activateVacationMode() sets: vacation_mode, vacation_mode_activated_at, vacation_mode_until
-            \Illuminate\Support\Facades\DB::table('users')
+            DB::table('users')
                 ->where('id', $buddyUserId)
                 ->update([
                     'vacation_mode' => false,
@@ -118,7 +122,7 @@ class FleetDispatchAcsDefendTest extends FleetDispatchTestCase
         // Also reset all vacation mode fields for the current test user
         // (testDispatchFleetFromVacationModeError sets this)
         if (isset($this->currentUserId)) {
-            \Illuminate\Support\Facades\DB::table('users')
+            DB::table('users')
                 ->where('id', $this->currentUserId)
                 ->update([
                     'vacation_mode' => false,


tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php
```diff
@@ -69,13 +69,45 @@ protected function messageCheckMissionReturn(PlanetService $destinationPlanet):
...
+        if (!empty($this->createdBuddyUserIds)) {
+            foreach ($this->createdBuddyUserIds as $buddyUserId) {
+                // Delete buddy requests involving this user
+                \Illuminate\Support\Facades\DB::table('buddy_requests')
+                    ->where('sender_user_id', $buddyUserId)
+                    ->orWhere('receiver_user_id', $buddyUserId)
+                    ->delete();
+
+                // Reset vacation mode to false
+                \Illuminate\Support\Facades\DB::table('users')
```
jackbayliss: Rather than handle this ourselves in the teardown we can use the RefreshDatabase trait- which means before and after each test it'll be refreshed  see https://laravel.com/docs/12.x/database-testing#resetting-the-database-after-each-test

Geda173: Thanks for the suggestion, but if I am not mistaken, this completely resets the database between each test, right? If so, we want to avoid that, as testing features locally becomes tedious as you have to completely set up shop again after running the tests. That's why I opted to "manually" clean up after the test to preserve the database.

jackbayliss: Aha, I see! Apologies I didn't realise that's what we were doing. Usually you'd setup all the models etc in the test itself using factories. Perhaps it's worth looking into further.

Geda173: I think you are right though, `AccountTestCase` already does `createAndLoginUser()`, so tests don't depend on pre-existing accounts. I'll give it a go and report back!

Geda173: Looks like we will have to stick with manual cleanup for now. The current test architecture uses `reloadApplication()`, which expects users to persist in the database. `RefreshDatabase` wipes everything between tests, causing `User::find($this->currentUserId)` to return null.

`DatabaseTransactions` could work since it only rolls back changes after each test (users persist during the test). However, each test class would need the trait added, and we'd need to verify it doesn't break tests that rely on committed data or background processes.

jackbayliss: DatabaseTransactions is a good shout, we could add it to the test cases, but one for another day agreed. 

Nice.

tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php
```diff
@@ -78,6 +84,52 @@ protected function setUp(): void
         $this->buddyPlanet = null;
     }
 
+    /**
+     * Clean up test data after each test to prevent state leakage.
+     * Only removes buddy relationships and resets vacation mode - test users and planets
+     * remain in database but won't have special state that affects subsequent tests.
+     */
+    protected function tearDown(): void
```
jackbayliss, line 96: Would be worth adding a `@todo` comment so we know to come back and refactor it?

tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php
```diff
@@ -78,6 +84,52 @@ protected function setUp(): void
...
+    {
+        // Clean up buddy relationships and vacation mode created during this test run
+        // Process and remove each ID to avoid accumulation
+        while (!empty(self::$allCreatedBuddyUserIds)) {
+            $buddyUserId = array_shift(self::$allCreatedBuddyUserIds);
+
+            // Delete buddy requests involving this user (with proper SQL grouping)
+            \Illuminate\Support\Facades\DB::table('buddy_requests')
```
jackbayliss: Can we import the DB class? Seeing as we're using it 3x why not :)

Geda173: Yeah, makes sense. Good catch!

lanedirt:
Thanks @Geda173 and @jackbayliss! I agree that the current testing setup can be a bit tricky with the random planet targeting, and test planets/users having different states because each test sets up the user and planet differently. And so this can interfere with expected outcomes.

Something to look at in the future if we can refactor the test framework to work more based on specific input and outcomes with e.g. the proposed transactions.

Merging this one, thanks again for the work on this 👍 !

PR #1018, opened 2026-01-04
HamillManoeuvreTest missing construct values

jackbayliss:
See https://github.com/lanedirt/OGameX/actions/runs/20679354929 

This simply adds in the construct values.. which - I hope are right :D


jackbayliss:
Merging this for now as it's a test.. and seems to pass.

Issue #1020, opened 2026-01-04
Consider using rector


jackbayliss:
I've noticed a few things re 

https://github.com/lanedirt/OGameX/issues/1019 
https://github.com/lanedirt/OGameX/issues/963

that https://github.com/rectorphp/rector may help with?

lanedirt:
@jackbayliss Sounds good to me!

Having an automated way to refactor the whole codebase would be very cool to change the existing code, and then allow us to have all new code enforced via linting. Would save a lot of time compared to doing it manually.

Issue #1019, opened 2026-01-04
Use property promotion


jackbayliss:
```
   protected SettingsService $settings;

    /**
     * PlayerServiceFactory.
     *
     * @var PlayerServiceFactory
     */
    protected PlayerServiceFactory $playerServiceFactory;

    /**
     * PlanetServiceFactory constructor.
     */
    public function __construct(SettingsService $settingsService, PlayerServiceFactory $playerServiceFactory)
    {
        $this->settings = $settingsService;
        $this->playerServiceFactory = $playerServiceFactory;
    }

```

Can just become 
```
    /**
     * PlanetServiceFactory constructor.
     */
    public function __construct(protected  SettingsService $settingsService, protected  PlayerServiceFactory $playerServiceFactory)
    {
        
    }


```

lanedirt:
@jackbayliss Looks like a nice improvement!

But in your second example, I'm assuming we don't need the explicit setters anymore right? So it would become like this with an empty body, correct?

```php
public function __construct(
    protected SettingsService $settings,
    protected PlayerServiceFactory $playerServiceFactory
) {}
```

jackbayliss:
@lanedirt correct we can lose the body! I'll adjust the main body  ☺️

Issue #963, opened 2025-12-29
Replace inline fully-qualified class references with the imported class name


jackbayliss:
I see a lot of usages of the FQCN ie: 

https://github.com/lanedirt/OGameX/blob/d736a6362476de8aaaa61b899eb6d0f3712a39f7/tests/Unit/HalvingServiceTest.php#L403

Rather than importing the class, this is inconsistent common Laravel conventions  (At least IMO) - I would perhaps refactor them to just import the class.

Not sure if you agree, but stuck out to me.

lanedirt:
Yep sounds good to me in general. Sometimes though when there are many one-off references, having them inline in the code as fully qualified can reduce import clutter.

But in general I’m in favour to following Laravel best practices, so +1 from me.

PR #1023, opened 2026-01-04
Feat: Add Rector

jackbayliss:
## Description
This closes https://github.com/lanedirt/OGameX/issues/1020,  https://github.com/lanedirt/OGameX/issues/1019 and https://github.com/lanedirt/OGameX/issues/963

### Type of Change:

- [X] Feature enhancement
- [X] Documentation update

### Changes:
- Adds Laravel rector to our dev requirements
- Adds a rector.php with a bunch of rules (for now I've just added some fairly simple ones, but we can build on it) 
- Added a new rector script in the composer.json  (so all weneed to do is run `composer run rector`)
- Adds a rector workflow, so it will fail if a contributor hasn't ran it
- Applied rector to the repo as is!  (I've split out the commits so should be super clear)

TLDR - Should save a lot of reviewing and keeps everything consistent

### DIFF
--- a/.github/workflows/run-rector-code-analysis.yml
+++ b/.github/workflows/run-rector-code-analysis.yml
@@ -23,7 +23,7 @@ jobs:
         coverage: none
 
     - name: Cache Composer packages
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: vendor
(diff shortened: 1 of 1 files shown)


lanedirt:
Looks good, thanks for getting this integrated! Will indeed save a lot of time and make everything more consistent :)!

Issue #855, opened 2025-12-02
Add Alliance Combat System


Geda173:
The Alliance Combat System (ACS) enables players to attack or defend together. The rules surrounding it are vast and there are a number of pitfalls.

ACS Attack:

- [ ] Maximum of 16 fleets / five players
- [ ] Players need to be Buddies or in the same Alliance
- [ ] ACS Attack missions cannot circumvent new player protections
- [ ] Loot maximum of attacked planet is still 50% (or server side configured value); loot is distributed among attackers' cargo space
- [ ] Invitations need to be registered as a new Message Class and be sorted into the appropriate tab in the Messaging System
- [ ] An ACS attack group can only ever be delayed by 30% of the remaining time until impact
- [ ] ACS Attack missions always use the combat techs of the main attacker / initiator
- [ ] Regular attacks can be converted to ACS Attack missions
- [ ] ACS Attack missions can generate moons
- [ ] ACS Attack missions can generate wreckage for the defender and any attacking player with the General class
- [ ] Codebase does currently not contain the ACS Attack .gif for the fleet widget, only the fleet dispatch
- [ ] ACS Attacks generate one combat report per participating player, regardless of the amount of fleets added (one player joining with two fleets will still only receive one combat report)
- [ ] (NEEDS VERIFICATION) Fleet widget should display each arriving fleet individually
- [ ] ACS Attack groups display the total amount of ships in the fleet widget, hovering over them should show a granular view, i.e. participants and amount of ships sent
- [ ] Combat Reports need to support granular views, i.e. providing drop down menus to allow viewing individual participating fleets and their losses
- [ ] Players can withdraw their fleets from the attack group at any point prior to impact
- [ ] (NEEDS VERIFICATION) If the initiating fleet is recalled, the remaining fleets in the attack group continue their approach
- [ ] (NEEDS VERIFICATION) Return missions of an ACS Attack use the same, synchronized speed of the parent mission (making slowed down sheeps also slow on their return mission)

ACS Defend / Hold

- [ ] Maximum of five players can aid another, foreign planet
- [ ] Players need to be Buddies or in the same Alliance
- [ ] You cannot defend the planet of a player that is still in new player protection
- [ ] Hold Times: 0, 1, 2, 4, 8, 16 or 32 hours
- [ ] Alliance Depot can supply deuterium to defending fleets
- [ ] Hold time can be extended by supplying more deuterium to defending fleet
[...]

PR #1061, opened 2026-01-09
Feat: Add Alliance Depot Functionality and Refactor ACS Defend

Geda173:
## Description
This PR implements the functionalities of the Alliance Depot.

 **Alliance Depot Building Interface**
- Alliance Depot Popup overlay
- Supply capacity (10,000 deuterium per hour per level)
- List of all fleets currently holding at the planet and its moon
- Fleet owner names in properly styled dropdown
- Real-time countdown timers showing remaining hold time
-  Ship counts for each fleet
- Deuterium cost calculator for supply rockets
- Imported missing assets for popup window

**Supply Rocket Functionality**
- **Send supply rockets** to extend friendly fleet hold times
- **Dynamic cost calculation** based on fleet compositio
- **Validation**: Only fleets holding for ≥1 hour can be extended
- No page reload** after supplying fleet - instant fadebox confirmation

Mission Integration:
- **Automatic hold time extension** for ACS Defend missions
- Handles fleets **before return mission creation** (extends `time_holding `on outbound mission)
- **Real-time updates** to fleet widget showing new hold duration

Feature Tests have been added as well. It also includes 2 critical bug fixes for ACS Defend missions:

- ACS Defend fleets can be recalled during hold times
- Handling of resources sent along with an ACS Defend mission (return + destruction of ships)

### Type of Change:
- [X] Feature enhancement

## Related Issues
Continues work on #855 

## Checklist

- [X] **Testing:**

## Additional Information
What an utterly, useless feature. My god. Extending the hold time is just completely pointless, as in regular play players would have their fleets arrive seconds before an incoming attack. Furthermore, the Alliance Depot only ever consumes Deuterium of the host planet, if the host extends the holding time. Therefore, there is no way to take some of the Deuterium cost away from the fleet owner upfront or in any other meaningful way.

### DIFF
--- a/app/Services/FleetMissionService.php
+++ b/app/Services/FleetMissionService.php
@@ -7,7 +7,10 @@ use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Support\Facades\Date;
 use OGame\Enums\FleetSpeedType;
 use OGame\Factories\GameMissionFactory;
+use OGame\Factories\PlanetServiceFactory;
 use OGame\GameConstants\UniverseConstants;
+use OGame\GameMessages\AcsDefendArrivalHost;
+use OGame\GameMessages\AcsDefendArrivalSender;
 use OGame\GameMissions\Abstracts\GameMission;
 use OGame\GameObjects\Models\Units\UnitCollection;
 use OGame\Models\Enums\PlanetType;
@@ -600,18 +603,18 @@ class FleetMissionService
      */
     private function sendAcsDefendArrivalMessages(FleetMission $mission): void
     {
-        $planetServiceFactory = app(\OGame\Factories\PlanetServiceFactory::class);
+        $planetServiceFactory = app(PlanetServiceFactory::class);
 
         $origin_planet = $planetServiceFactory->make($mission->planet_id_from, true);
         $target_planet = $planetServiceFactory->make($mission->planet_id_to, true);
 
         // Send message to sender (Fleet Command)
-        $this->messageService->sendSystemMessageToPlayer($origin_planet->getPlayer(), \OGame\GameMessages\AcsDefendArrivalSender::class, [
+        $this->messageService->sendSystemMessageToPlayer($origin_planet->getPlayer(), AcsDefendArrivalSender::class, [
             'to' => '[planet]' . $mission->planet_id_to . '[/planet]',
         ]);
 
         // Send message to host/target (Space Monitoring)
-        $this->messageService->sendSystemMessageToPlayer($target_planet->getPlayer(), \OGame\GameMessages\AcsDefendArrivalHost::class, [
+        $this->messageService->sendSystemMessageToPlayer($target_planet->getPlayer(), AcsDefendArrivalHost::class, [
             'to' => '[planet]' . $mission->planet_id_to . '[/planet]',
         ]);
     }
--- a/app/Http/Controllers/ResourcesController.php
+++ b/app/Http/Controllers/ResourcesController.php
@@ -2,7 +2,6 @@
 
 namespace OGame\Http\Controllers;
 
-use OGame\Services\SettingsService;
 use Exception;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
@@ -18,6 +17,7 @@ use OGame\Services\BuildingQueueService;
 use OGame\Services\CharacterClassService;
 use OGame\Services\ObjectService;
 use OGame\Services\PlayerService;
+use OGame\Services\SettingsService;
 use OGame\Services\UnitQueueService;
 
 class ResourcesController extends AbstractBuildingsController
--- a/tests/Feature/FleetDispatch/FleetDispatchMissionResourceHandlingTest.php
+++ b/tests/Feature/FleetDispatch/FleetDispatchMissionResourceHandlingTest.php
@@ -2,15 +2,15 @@
 
 namespace Tests\Feature\FleetDispatch;
 
-use OGame\Services\DebrisFieldService;
-use OGame\Models\Enums\PlanetType;
 use Illuminate\Support\Facades\DB;
 use OGame\Factories\PlanetServiceFactory;
 use OGame\GameObjects\Models\Units\UnitCollection;
+use OGame\Models\Enums\PlanetType;
 use OGame\Models\Planet;
 use OGame\Models\Resources;
 use OGame\Models\User;
 use OGame\Services\BuddyService;
+use OGame\Services\DebrisFieldService;
 use OGame\Services\FleetMissionService;
 use OGame\Services\ObjectService;
 use OGame\Services\PlayerService;


Geda173:
There is one thing about the implementation that also needs to be discussed: In universes where the ACS is disabled, the Alliance Depot can't be built. Should we implement it like that as well and tie the presence of the Depot in /facilities to the ACS server settings?

lanedirt:
> There is one thing about the implementation that also needs to be discussed: In universes where the ACS is disabled, the Alliance Depot can't be built. Should we implement it like that as well and tie the presence of the Depot in /facilities to the ACS server settings?

I didn't know that the official game implemented this limitation. But adding this same behavior would be a nice detail IMO. But not a dealbreaker for this PR, so we can also add it later.

Btw I had one more bug with the preview. Could you merge main into this branch once more? Then the preview environment should work again. 😄

Geda173:
> > There is one thing about the implementation that also needs to be discussed: In universes where the ACS is disabled, the Alliance Depot can't be built. Should we implement it like that as well and tie the presence of the Depot in /facilities to the ACS server settings?
> 
> I didn't know that the official game implemented this limitation. But adding this same behavior would be a nice detail IMO. But not a dealbreaker for this PR, so we can also add it later.
> 
> Btw I had one more bug with the preview. Could you merge main into this branch once more? Then the preview environment should work again. 😄

I'll merge it momentarily. 

Not having the Alliance Depot available is not a set rule. Universe 20 did not have the ACS enabled and the Alliance Depot was thus disabled alongside it. The universe Jupiter also had ACS disabled, but the Alliance Depot was available.

lanedirt:
Hi @Geda173,

Looking good! I've never really used the alliance depot feature myself, so I can't really properly tell how it should work. However I did notice the following:

1. Made buddies between test2 and test3 players.
2. Send fleet from test3 > test2 with ACS defend.
3. As `test2` extend duration with 3 hours.
4. `test3` player tries to recall fleet: it says its successful (HTTP 200) but the fleet is not actually recalled, see recording here:

https://github.com/user-attachments/assets/0ae8d961-ba55-400b-ae2d-81523c7a3f34

You can login with the `test2` and `test3` players in the preview environment to see it for yourself.

Geda173:
> Hi @Geda173,
> 
> Looking good! I've never really used the alliance depot feature myself, so I can't really properly tell how it should work. However I did notice the following:
> 
> 1. Made buddies between test2 and test3 players.
> 2. Send fleet from test3 > test2 with ACS defend.
> 3. As `test2` extend duration with 3 hours.
> 4. `test3` player tries to recall fleet: it says its successful (HTTP 200) but the fleet is not actually recalled, see recording here:
> You can login with the `test2` and `test3` players in the preview environment to see it for yourself. 

That's a good catch! Sending a supply rocket modifies `time_holding`. However, the `cancelMission` method has a blanket check that prevents any recall after `time_arrival`, which doesn't account for missions with `time_holding`. I'll have to check  the Fleet Mission Service and make adjustments that don't interfere with Expeditions, as meddling with how holding periods are treated could interfere with them as well.

Other than that, the Alliance Depot is sadly a waste of everyone's time. You send an ACS Defend mission to your buddy/alliance member and pay all the Deuterium it needs for travel + set holding time upfront. If your buddy wishes, they can send supply rockets up and extend the holding time by x hours. That's the entire feature set.

Even Gameforge have not invested much time. For example, when one player sends multiple fleets to the same host planet with a depot, the host can select all available fleets in the drop down menu but the number of ships is always static and displays the first fleet to have arrived. The display is thus not dynamic at all and you can never display the other fleets or ship counts of the same player. Sending supply rockets and seeing the calculations works though. It's wonky in all kinds of ways.

Geda173:
I am still working on the return during holding time issue. It is now properly working but then it started setting the return time to include the time already spent holding and from there the issues cascaded. I'll request a re-review once I have those kinks ironed out.

app/GameMissions/Abstracts/GameMission.php
```diff
@@ -158,6 +158,20 @@ public function cancel(FleetMission $mission): void
         $mission->processed = 1;
         $mission->save();
 
+        // If the mission had already arrived (ACS Defend during hold time), find and cancel
```
lanedirt: For my information: could you explain what would cause this edge case?

I think logically, "canceling" a return mission is a bit strange and I think not possible in the normal game flow. So having a return mission carry the status "canceled" in the db seems a bit off to me (although technically it should work).

Do we have any alternative options here?

app/Services/FleetMissionService.php
```diff
@@ -239,7 +239,9 @@ public function getActiveFleetMissionsSentByCurrentPlayer(): Collection
     {
         // Note: this only includes missions that the current player has sent themselves
         // so it does not include any incoming missions by other players (e.g. hostile attacks, espionage, transports etc.)
-        $query = $this->model->where('user_id', $this->player->getId())->where('processed', 0);
+        $query = $this->model->where('user_id', $this->player->getId())
+            ->where('processed', 0)
+            ->where('canceled', 0); // Exclude canceled missions from fleet slot count
```
lanedirt: If a mission is "canceled", shouldn't it also have the status "processed = 1", and therefore this check could stay as-is?

Not sure though how the other missions handle this, haven't checked the main code.

app/GameMissions/Abstracts/GameMission.php
Geda173: The Edge Case Scenario:
1. You send an ACS Defend fleet with 2 hour hold time
2. Fleet arrives at destination → Game automatically creates a return mission scheduled to leave in 2 hours
3. Someone uses supply rocket → Return mission departure time gets pushed back to 4 hours from now
4. You decide to recall the fleet → Game needs to create a NEW return mission that leaves immediately

The Problem:
- You now have TWO return missions in the database:
- Old return mission: scheduled to leave in 4 hours
- New return mission: leaves immediately
- Both contain the same ships → Ship duplication!

Having them marked as cancelled is weird, instead we should probably delete them altogether.

app/Services/FleetMissionService.php
Geda173: True, I think we can remove the redundant check. I have been fighting this ship duplication bug for many hours and it all started with ACS Defend not being able to be recalled during the hold time, which lead to incorrect return times, ship duplication and ultimately fleet slot occupation with incorrectly tracked return missions.

app/GameMissions/Abstracts/GameMission.php
lanedirt: Gotcha, thanks for the explanation.

One possible option to prevent this issue might be to refactor how ACS defend works. Instead of
`Arrive at destination` → `create return mission` → `start holding`, we could follow the expedition mission model, where the holding time is added on top of the arrival time and the mission is not marked as arrived until both have elapsed.

That way, no return mission would exist while the fleet is holding. And then both usecases: 1) fleet hold expires , mission is marked `arrived` and `processed`, and then automatically returns by creating return trip, and 2) fleet hold is manually recalled by owner, would both work without conflicting with each other. Doing it similar to this approach would likely avoid this situation.

Geda173: That would be the right thing to do. I'll get on it. It will take time, as I pretty much need to restart everything I already have. The Alliance Depot currently checks for fleet holding and it would need to be changed to find fleets by checking `time_arrival + time_holding > now`. 

So what I am going to do is refactor ACS Defend to be more expedition like and deal with all the fallout that stems from that. I'll mark this as a draft and mark it as ready once I have finished the work on it. Thanks for the feedback!

lanedirt:
Hi @Geda173,

I just tested the latest version. Now when a ACS fleet is in the "hold" state, the alliance hangar shows "there are no holding fleets". See in the screenshot below, there is an holding fleet active (counting down from 1 hour):

image

Geda173:
> Hi @Geda173,
> 
> I just tested the latest version. Now when a ACS fleet is in the "hold" state, the alliance hangar shows "there are no holding fleets". See in the screenshot below, there is an holding fleet active (counting down from 1 hour):

One of these days I will just move into the woods and call it quits. I had the exact same issue, solved it and of course only then pushed the commits. I'll check it out and try to figure out what the problem is.

Geda173:
I just checked again and it works on my end. The only changes I applied was a small bug I found. The new expedition-like logic broke the arrival message for the holding fleet but other than that, the Alliance Depot is recognizing the holding fleet and I can extend it. Can you maybe walk me through what you did so we can troubleshoot the issue?

lanedirt:
> I just checked again and it works on my end.

I checked it again just now and indeed it works. Extending holding time works fine, and recalling fleet as well. Perhaps I selected a different planet with the alliance depot as where the ACS defend fleet was holding the first time around.

So Llooks good, thanks for the work on this! Merging!

PR #1102, opened 2026-01-14
Fix: Handle float precision in resource deduction for large amounts of resources

### LINKED ISSUE #1014: [BUG] "Not enough resources" error when sending large amounts of available resources
When a large amount of resources is to be sent, for example 75 million each, using a fleet with more than enough capacity, the following error message appears:
Image

This also happens if the "all resources" button is pressed to automatically select the resources.
It is possible to send the fleet if a certain amount of resources is removed from the cargo, in the example above I was able to send the fleet when I removed 100k each, for larger amounts of resources more needs to be removed.

Geda173:
## Description
This PR fixes an issue where attempting to send large amounts of resources (75M+) would fail with "Not enough resources or units on the planet" error even when sufficient capacity was available.

Root cause: Resource columns use FLOAT type which loses precision with large values (20M+). Displayed value can differ from actual stored value by small amounts (~8-12 units), causing WHERE clause comparison to fail.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #1014 

## Checklist

- [X] **Testing:**


app/Services/PlanetService.php
```diff
@@ -1663,27 +1663,33 @@ public function deductResourcesAtomic(Resources $resources): bool
...
-            $query->where('crystal', '>=', $crystalCost);
+            $conditions[] = "crystal + 1000 > {$crystalCost}";
         }
         if ($deuteriumCost > 0) {
-            $query->where('deuterium', '>=', $deuteriumCost);
+            $conditions[] = "deuterium + 1000 > {$deuteriumCost}";
+        }
+
+        if (!empty($conditions)) {
+            $query->whereRaw(implode(' AND ', $conditions));
```
lanedirt: I just checked the docs and indeed the `FLOAT` column types we're using now for resource columns are limited in their precision (~7 digits).

While adding an arbitrary buffer like this would work, it is more of a workaround instead of solving the actual problem. So in this case I would propose that we change the db structure instead and make all resource columns a `DOUBLE` instead which has ~15 digits of precision (1.000.000B), which I think should be more than enough for most game purposes.

When we change the db column types, the query logic can stay the same as how it was originally (i.e.: revert changes), and then the tests you added should still pass.

Geda173: Absolutely. I have reverted the changes to `PlanetService.php` and instead created a database migration to change the database to accomodate much larger digit precision. The tests are still passing.

PR #1134, opened 2026-01-21
Implement planet and moon field restriction for building construction

### LINKED ISSUE #393: When all planet fields are used no buildings should be able to be built anymore
PR #390 added logic to calculate the amount of used fields on a planet.

When all fields on a planet are used up the building or upgrading of buildings should be disabled. I am not sure yet what it looks like in the official game when a planet is actually full. But we could start by disallowing the upgrade via a server-side exception and improve it with the proper client-side state/warnings later.

**Help wanted:**
If someone knows what the official game shows when a planet is full feel free to comment.

Todo:
- [ ] Add check if all planet fields are used. If so, disallow the building or upgrading of buildings.

Further notes:
- When a planet is full, the only way to upgrade a building is by demolishing/downgrading an existing building. This logic is not built yet, see issue #110.
- The way to continue expanding a planet that is (almost) full is by upgrading the `Terraformer` building which gives additional bonus planet fields per level. This terraformer logic however is not implemented yet. A separate issue is created for this, see #392.

Geda173:
## Description
This PR implements planet and moon field restrictions as well as a host of feature tests for both planets and moons. I have added server-side validations in `BuildingQueueService` as well as `ViewModel` properites to ensure the correct UI state. Special buildings like Space Dock, Solar Satellites, and Crawlers don't consume fields and aren't restricted.

### Type of Change:
- [X] Feature enhancement

## Related Issues
Fixes #393 

## Checklist

- [X] **Testing:**

### DIFF
--- a/app/Http/Traits/ObjectAjaxTrait.php
+++ b/app/Http/Traits/ObjectAjaxTrait.php
@@ -187,6 +187,7 @@ trait ObjectAjaxTrait
             $maxFields = $planet->getPlanetFieldMax();
 
             // Calculate the projected building count after all queued items complete
+            // Only apply this for building queues (BuildingQueueViewModel with level_target property)
             $queuedFieldChange = 0;
             $build_active = $build_queue->getCurrentlyBuildingFromQueue();
             $build_queued = $build_queue->getQueuedFromQueue();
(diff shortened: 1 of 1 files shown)


lanedirt:
Hi @Geda173, thanks for the work on this! I just played around with it a bit on the testing environment. 

One thing I found is that it looks like its possible to build more buildings than fields because of the queue system.

E.g. if I am at `170/171` fields and I try to build something: I get the warning that I'm about to use up my last field. Then I press `OK` and it's added to the queue.

However once that queued item is not finished yet, I can keep adding more buildings to the queue (and every time get the same `you're using up your last field` warning again). Then if I l let the queue complete, I can end up with e.g. `173/171` fields instead of the max.

Geda173:
> Hi @Geda173, thanks for the work on this! I just played around with it a bit on the testing environment.
> 
> One thing I found is that it looks like its possible to build more buildings than fields because of the queue system.
> 
> E.g. if I am at `170/171` fields and I try to build something: I get the warning that I'm about to use up my last field. Then I press `OK` and it's added to the queue.
> 
> However once that queued item is not finished yet, I can keep adding more buildings to the queue (and every time get the same `you're using up your last field` warning again). Then if I l let the queue complete, I can end up with e.g. `173/171` fields instead of the max.

Ah, so basically the same issue that the missile silo had. That should be an easy fix.

lanedirt:
Awesome, looking good! Merging!

PR #1180, opened 2026-02-02
fix: trader max value not possible

### LINKED ISSUE #1053: [BUG] Resource Market max trade value impossible
**Describe the bug**
When attempting to trade at the Resource Market, it is currently impossible to trade for the maximum amount of a resource, because resource production continues in the background, making the calculated resources in the UI outdated. This leads to an attempt to gain more resources than the storage value allows for.

**Expected behavior**
There should be a buffer of x% over storage budget to allow transactions to still reasonably conclude without the game throwing errors.

**Screenshots**

Image

193 crystal should not make or break this transaction. It functionally makes the "gain max resources" button impossible to use.

Geda173:
## Description
This PR implements fixes an issue in which the max trade functionality in the trader section was not properly working because it would fill the storage over capacity, making the trade fail. Instead, MerchantService now fills the storage exactly to its limit, making the trade successful.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #1053 

## Checklist

- [X] **Testing:**

## Additional Information

### DIFF
--- a/app/Services/MerchantService.php
+++ b/app/Services/MerchantService.php
@@ -23,7 +23,6 @@ class MerchantService
      */
     public const DARK_MATTER_COST = 3500;
 
-
     /**
      * Call a merchant.
      *
@@ -256,7 +255,7 @@ class MerchantService
 
                 // Verify player still has enough after adjustment
                 if ($currentAmount < $giveAmount) {
-                    $giveAmount = $currentAmount;
+                    $giveAmount = (int)floor($currentAmount);
                     $receiveAmount = (int)floor($giveAmount * $exchangeRate);
                 }
             } else {


lanedirt:
Hi @Geda173,

Thanks for looking into this. I was wondering though: did you get an idea how the official game does this? Do they also allow a buffer?

Because I was thinking, wouldn't it be simpler if in the `MerchantService.php` where the resources are appended, we simply do a check there what the current resources are, and in case where we would go over the storage cap, we trim off a part of the requested resources so it matches up to exactly the storage cap in that transaction?

Like for following scenario:
1. User has `50.000` metal and a storage cap of `100.000`
2. User opens trader page, selects max metal to add: `50.000`
3. - In the meanwhile resource production has continued, making it now `51.000` -
4. User submits trade from step 2: server calculates the actual max resources to give, and given the cap of 100k, it actually gives out `49.000` instead of the requested 50k.

So instead of allowing to exceed the storage, we cap it (in potentially negative way for the player, however most scenarios we would only be capping small percentages representing seconds or maybe minutes of resources production, so that would not need to cause worry I think).

Just my two cents, I think this feels like a safer server-side guard to me in general for these kind of usecases, but open to discussion.

Geda173:
> Hi @Geda173,
> 
> Thanks for looking into this. I was wondering though: did you get an idea how the official game does this? Do they also allow a buffer?
> 
> Because I was thinking, wouldn't it be simpler if in the `MerchantService.php` where the resources are appended, we simply do a check there what the current resources are, and in case where we would go over the storage cap, we trim off a part of the requested resources so it matches up to exactly the storage cap in that transaction?
> 
> Like for following scenario:
> 
> 1. User has `50.000` metal and a storage cap of `100.000`
> 2. User opens trader page, selects max metal to add: `50.000`
> 3. * In the meanwhile resource production has continued, making it now `51.000` -
> 4. User submits trade from step 2: server calculates the actual max resources to give, and given the cap of 100k, it actually gives out `49.000` instead of the requested 50k.
> 
> So instead of allowing to exceed the storage, we cap it (in potentially negative way for the player, however most scenarios we would only be capping small percentages representing seconds or maybe minutes of resources production, so that would not need to cause worry I think).
> 
> Just my two cents, I think this feels like a safer server-side guard to me in general for these kind of usecases, but open to discussion.

Hi @lanedirt,

I implemented it the way I did because I was certain that trading for max value would fill the storage slightly above cap but I just called the merchant multiple times in vanilla and it always filled up the storage exactly to the maximum amount. I have pushed some changes that revert my earlier changes and instead attempt to implement it the way OGame does.

lanedirt:
Looks to be working good, thanks, merging!

PR #1214, opened 2026-02-21
fix: player defenses incorrectly added to NPC fleet on expeditions

### LINKED ISSUE #1100: [BUG] Expedition pirates much stronger than intended
**Describe the bug**
Pirates encountered during expeditions are significantly stronger than they should be. This is a recent change and was not the case previously. For example, in a recent expedition, the pirates’ attacks are sending an abnormally high number of shots, resulting in damage far beyond what is expected.

**Expected behavior**
Pirates during expeditions should have normal strength and deal damage within expected limits. Their attacks should not exceed the normal maximum damage values.

**Screenshots**

Image

Screenshot from January 9 expedition shows normal pirate strength.

Image

Screenshot from January 12 expedition shows pirates dealing extreme damage.

As shown in the January 12 screenshot, the pirate Shooters fired 167,943 shots with a total damage of 48,955,252, which is much higher than should be possible.

**Additional context**
I’m not sure what specifically caused this issue, but it seems to have been introduced recently.

Geda173:
## Description
 Fixes a bug where expedition battles against pirates or aliens could result in wildly incorrect shot counts,  for example, an NPC fleet of ~3,700 ships reporting over 500,000 shots fired in a single round.

As mentioned in https://github.com/lanedirt/OGameX/issues/1100#issuecomment-3936990260, the player's expedition fleet always fights against the player's own defenses on the origin planet. Players with heavy fortifications will always wipe out their own fleets, while players without defenses will have a regular expedition experience.

A secondary bug was also fixed: NPC owner IDs are negative (-1 for pirates, -2 for aliens) but get clamped to 0 when serialised to the Rust engine (which uses u32). The matching logic in `convertBattleOutput()` was comparing the original negative ID against the clamped value returned by Rust, so NPC per-fleet results were silently dropped after every battle.

 Changes

- `app/Services/NPCPlanetService.php`: override `getObjectAmount()`to return 0, so no real planet data leaks into the NPC fleet
- `app/GameMissions/BattleEngine/RustBattleEngine.php`: apply the same `max(0, ownerId)` clamping during result matching in `convertBattleOutput()` so NPC fleet results are correctly populated
- `tests/Feature/FleetDispatch/FleetDispatchExpeditionTest.php`: two regression tests: one asserting defense units are excluded from the NPC fleet, one asserting hit counts stay within a reasonable bound for the fleet composition that originally triggered the bug

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #1100 

## Checklist

- [ ] **Automated Refactoring:** Rector has been run and no outstanding issues remain.
- [ ] **Testing:**

## Additional Information

### DIFF
--- a/tests/Feature/MessagesTest.php
+++ b/tests/Feature/MessagesTest.php
@@ -4,6 +4,7 @@ namespace Tests\Feature;
 
 use Lang;
 use OGame\Factories\GameMessageFactory;
+use OGame\GameMessages\Abstracts\ExpeditionGameMessage;
 use OGame\Models\BattleReport;
 use OGame\Models\EspionageReport;
 use OGame\Models\Message;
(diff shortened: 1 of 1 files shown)


lanedirt:
Looks good, thanks for investigating this issue and applying the fix!

PR #1268, opened 2026-03-10
FIX Incorrect fleet dispatch text

piciolo:
Description
This Pull Request replaces the hardcoded English string "Send espionage probe to:" with the appropriate translation key t_ingame.fleet.fleet_dispatch in the FleetController (specifically within case 6 handling espionage missions).

The problem: Previously, sending an espionage probe from the Galaxy view resulted in a response message that was either hardcoded in English or displayed as a raw translation key if the localization helper was missing. This was particularly visible to users using non-English locales (e.g., Italian).

The fix:

Utilized Laravel's __() helper to wrap the translation key.

Ensured the existing logic for probe count and coordinate concatenation remains intact.

Verified that the key t_ingame.fleet.fleet_dispatch is correctly defined in language files (EN/IT/NL).

Confirmed functionality via Laravel Tinker: app()->setLocale('it'); echo __('t_ingame.fleet.fleet_dispatch'); successfully returned "Invio flotta".

Type of Change:
[x] Bug fix

[ ] Feature enhancement

[ ] Documentation update

[ ] Other (please describe):

Related Issues
Fixes # [1168]

Checklist

[x] Testing:

[x] Relevant unit and feature tests are included or updated.

[x] Tests successfully run locally (Verified via Tinker and manual browser testing).

[ ] CSS & JS Build: No changes made to JS/CSS files.

[ ] Documentation: Documentation has been updated to reflect any changes made.

image

image

### DIFF
--- a/app/Http/Controllers/FleetController.php
+++ b/app/Http/Controllers/FleetController.php
@@ -544,10 +544,10 @@ class FleetController extends OGameController
                 break;
             case 8: // Recycle
                 if ($position === UniverseConstants::EXPEDITION_POSITION) {
-                    $responseMessage = __('Send pathfinder to:');
+                    $responseMessage = __('t_ingame.fleet.fleet_dispatch');
                     $units->addUnit(ObjectService::getUnitObjectByMachineName('pathfinder'), $shipCount);
                 } else {
-                    $responseMessage = __('Send recycler to:');
+                    $responseMessage = __('t_ingame.fleet.fleet_dispatch');
                     $units->addUnit(ObjectService::getUnitObjectByMachineName('recycler'), $shipCount);
                 }
                 break;


Geda173:
Hi @piciolo,

this change is not necessary, as espionage probes do not follow the same convention as debris recycling ships. The current text for espionage probes is correct.

piciolo:
> Hi @piciolo,
> 
> this change is not necessary, as espionage probes do not follow the same convention as debris recycling ships. The current text for espionage probes is correct.

Hi @Geda173, thanks for the feedback on #1268. To make sure I get the logic 100% right this time, could you clarify exactly which missions should follow the new convention?

For example:

Recycle/Pathfinder mining: Should show Fleet dispatch [coord] (n) Ok.

Espionage: Should keep the original text (ship names).

Other missions (Transport, Attack, etc.): Should these also be updated to the generic format or stay as they are?

I want to ensure I only modify the strictly necessary parts. Thanks for your help!

Geda173:
> > Hi @piciolo,
> > this change is not necessary, as espionage probes do not follow the same convention as debris recycling ships. The current text for espionage probes is correct.
> 
> Hi @Geda173, thanks for the feedback on #1268. To make sure I get the logic 100% right this time, could you clarify exactly which missions should follow the new convention?
> 
> For example:
> 
> Recycle/Pathfinder mining: Should show Fleet dispatch [coord] (n) Ok.
> 
> Espionage: Should keep the original text (ship names).
> 
> Other missions (Transport, Attack, etc.): Should these also be updated to the generic format or stay as they are?
> 
> I want to ensure I only modify the strictly necessary parts. Thanks for your help!

Hi @piciolo,

Correct, mining shows the above message, while espionage shows the specific ship names. Other missions do not have a quick send button, therefore they do not need adjustment.

app/Http/Controllers/FleetController.php
Geda173, line 1: This diff shows all 914 lines changed but the before/after content looks identical. It seems like a line-ending or encoding change snuck in?

piciolo: I'll do a clean PR here too, something in my GIT that causes problems.

PR #1276, opened 2026-03-12
Fix adjust missile attack behavior when there are 0 missiles available #1054

### LINKED ISSUE #1054: [BUG]  Adjust missile attack behavior when there are 0 missiles available
**Describe the bug**
OGameX currently features an error screen when wanting to send a missile attack to a planet with the currently selected planet not having any missiles.

**Expected behavior**
The missile attack button should be available and open up the missile attack dialog, but display 0 rockets and have the launch button be greyed out.

piciolo:
Description
This PR improves the UX and prevents errors when a player has no Interplanetary Missiles available on their active planet. Following community feedback, the fix now operates on two levels:

Galaxy View: The missile attack icon is now hidden from the planet's mouse-over menu if the player's active planet has 0 missiles.

Overlay Fallback: As a safeguard for players who haven't refreshed their page after firing all missiles, the "Fire" button in the missile attack overlay is now visually and functionally disabled if the missile count is 0.

Type of Change:
[x] Bug fix

[ ] Feature enhancement

[ ] Documentation update

[ ] Other (please describe):

Related Issues
Fixes #1054

Checklist

[x] Testing:

Verified that the missile icon disappears from the Galaxy view when the ammo count is 0.

Verified that the "Fire" button is disabled in the overlay if accessed directly with 0 missiles.

Tests successfully run locally.

[x] CSS & JS Build: N/A.

[x] Documentation: UI behavior now matches the expected game flow described in the issue feedback.

Additional Information
This dual approach ensures a clean UI by not showing impossible actions, while also preventing LogicException errors in edge cases where the overlay is already open or accessed via a stale page state.
image

image

### DIFF
--- a/app/Http/Controllers/GalaxyController.php
+++ b/app/Http/Controllers/GalaxyController.php
@@ -68,7 +68,11 @@ class GalaxyController extends OGameController
             $galaxy = (int)$galaxy_qs;
             $system = (int)$system_qs;
         }
+<<<<<<< Updated upstream
 
+=======
+// Fix #1054: Hide - missile icon if no missiles are available on active planet
+>>>>>>> Stashed changes
         return view('ingame.galaxy.index')->with([
             'current_galaxy' => $galaxy,
             'current_system' => $system,
--- a/app/Http/Controllers/GalaxyController.php
+++ b/app/Http/Controllers/GalaxyController.php
@@ -423,7 +423,7 @@ class GalaxyController extends OGameController
             $targetCoordinate = new Coordinate($galaxy, $system, $position);
             $distance = $this->calculateSystemDistance($currentPlanet->getPlanetCoordinates(), $targetCoordinate);
 
-            if ($distance <= $missileRange && $currentPlanet->getObjectAmount('interplanetary_missile') > 0) {
+            if ($distance <= $missileRange) {
                 $canMissileAttack = true;
                 $missileAttackLink = route('galaxy.missile-attack.overlay', [
                     'galaxy' => $galaxy,
@@ -932,7 +932,7 @@ class GalaxyController extends OGameController
             'system' => 'required|integer|min:1',
             'position' => 'required|integer|min:1|max:15',
             'type' => 'required|integer',
-            'missile_count' => 'required|integer|min:1',
+            'missile_count' => 'required|integer|min:0',
             'target_priority' => 'required|integer|min:0|max:7',
         ]);
 
@@ -948,10 +948,17 @@ class GalaxyController extends OGameController
 
         // Check if player has enough missiles
         $availableMissiles = $currentPlanet->getObjectAmount('interplanetary_missile');
+        if ($missileCount === 0 || $availableMissiles === 0) {
+            return response()->json([
+                'success' => false,
+                'error' => __('t_ingame.galaxy.insufficient_range'),
+                'close_overlay' => true,
+            ], 400);
+        }
         if ($missileCount > $availableMissiles) {
             return response()->json([
                 'success' => false,
-                'error' => __('Not enough missiles available'),
+                'error' => __('t_ingame.galaxy.not_enough_missiles'),
             ], 400);
         }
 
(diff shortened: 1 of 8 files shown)


piciolo:
@Geda173 
Let me know if that's okay now, thank you.

Geda173:
> @Geda173 Let me know if that's okay now, thank you.

Hi @piciolo,

I just checked again in OGame and the flow is a bit different there. I have attached a video. Disregard all the colored options, those are part of an extension and not part of the original game. The main difference here is that after depelting all Interplanetary Missiles and then opening the missile dialogue again, the button is not greyed out. Instead the dialogue remains fully functional. Upon firing with 0 missiles, the window closes and an error fadebox displays an error about insuffiecient range.

https://github.com/user-attachments/assets/565d48d1-c93c-472e-89ea-7f1d1597c2bc

piciolo:
@Geda173 
Thank you for the video and the feedback! I have updated the PR to match the exact flow shown:

UI Changes: >   - The missile attack modal now opens even if the player has 0 missiles (range check is now performed only on the system distance).

The "Fire" button is never disabled, even with zero missiles.

The input field defaults to 0 when no missiles are available.

Backend & Logic:

Updated GalaxyController.php to allow the modal to open with 0 missiles and refined the validation logic.

If an attack is launched with 0 missiles, the backend returns a 400 error with the specific message: "Insufficient range (research level impulse drive) of your interplanetary missiles!".

Added the insufficient_range translation key to English, Italian, and Dutch.

JavaScript Flow: >   - The modal now closes automatically before displaying the red error fadebox when a "close_overlay" flag is received from the server (matching the video behavior).

This should now be 100% faithful to the original game. Could you please check this in the preview deployment? Thanks!

image

Geda173:
Hi @piciolo,

I just tested the implementation and we have a regression now. When there are 0 missiles available or the planet is out of range, the missile attack option in /galaxy should not be visible in the mouse over of a target planet. The only instance in which the button should be visible while at 0 missiles is when you just fired all of your missiles and there hasn't been a page refresh yet.

piciolo:
@Geda173 
I've just pushed a fix for the regression you pointed out in.

The issue was a discrepancy between the backend flags and the frontend rendering logic in the Galaxy view. I've now added an explicit check for availableMissiles directly in the JS rendering function.

Locally, the "Missile Attack" link is now correctly hidden when missiles are at 0 or targets are out of range. Could you please re-test the latest PR preview whenever you have a moment?

Thanks!

Geda173 (approved):
Frontend changes are looking good, the flow is currently the way it should be. I'll let @lanedirt take another look and then it should be good to merge.

app/Http/Controllers/GalaxyController.php
```diff
@@ -109,7 +109,7 @@ public function getGalaxyArray(int $galaxy, int $system, PlayerService $player,
 
         $planets = [];
         foreach ($planet_list as $planet) {
-            $planetService = $planetServiceFactory->makeFromModel($planet);
+            $planetService = $planetServiceFactory->makeFromModel($planet instanceof Planet ? $planet : new Planet((array)$planet));
```
lanedirt: Is this newly added sanity check actually necessary here? 

I have personally not seen errors regarding the way the `makeFromModel()` works previously.

lanedirt (commented):
Hi both, thanks for the work and review on this! A small comment from my side. But not blocking for the merge. Other than this looks good!

app/Http/Controllers/GalaxyController.php
piciolo: You're right, this check isn't necessary and I'm removing it.

I added it in commit aa3235d ("Allow missile overlay with zero missiles") as a defensive precaution, but it was tangential to that commit's actual purpose and not fixing any observed bug. In this context $planet_list comes from Planet::where(...)->get(), so Eloquent guarantees each $planet is a Planet instance; the else branch (new Planet((array)$planet)) is unreachable. On top of that, makeFromModel() is type-hinted as Planet $planet, so any non-Planet input would already fail at the signature. If the fallback ever did trigger it would silently pass a non-persisted model to PlanetService, which would be worse than a clean TypeError. Reverted to the original line.

PR #1296, opened 2026-03-16
Use translation key for wrong_class tooltip

### LINKED ISSUE #1142: [BUG] Wrong mouse-over text for non-buildable ships (player classes)
**Describe the bug**
When trying to build something that can only be built by a different player class, the mouse over text on "build" or "improve" is incorrect, as it states: "Wrong character class!"

**Expected behavior**
It should say: "You need to activate the matching class to build this."

piciolo:
## Description
This PR addresses an issue where the error message for building units restricted by player classes was hardcoded and poorly phrased. 

Changes include:
- **Centralization:** Moved the "Wrong character class!" message from a hardcoded string in the Blade view to a proper namespaced translation key: `t_ingame.buildings.wrong_class`.
- **UX Improvement:** Rephrased the message to "You need to activate the matching class to build this." to provide clearer instructions to the user.
- **Multilingual Support:** Updated the translation files for English (EN), Italian (IT), and Dutch (NL).

### Type of Change:
- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #1142

## Checklist

- [x] **Testing:**

## Additional Information
Updating the AJAX view (`object.blade.php`) was critical as it was using the raw string as a translation key, which prevented localization updates from taking effect. The new implementation follows the project's standards for namespaced translations.

image

image


Geda173:
Hi @piciolo,

thanks for the fix and making it immediately translatable. I do have to admit though, I made a mistake when submitting the issue. I just checked again in OGame and the text is slightly different from what I reported:
image
image

And for the Pathfinder the text is accordingly. Could you please update the texts? You don't have to close this PR, just update the texts, push the commit and we can merge it then.

piciolo:
Hi! No worries at all, thanks for providing the exact vanilla OGame texts.

I will update the translation keys to be dynamic so they specifically mention the required class (General, Collector, or Discoverer) depending on the unit you are inspecting, matching vanilla 1:1.

Also, just a quick heads-up: while working on this, I noticed several other hardcoded English strings scattered across the idle queues (Shipyard, Research, Facilities, Overview, etc.). I am currently running a global sweep to fix all of those and namespace them properly for EN, IT, and NL.

I will include both the updated class texts and the global translation fixes in my next commit. I'll let you know as soon as I push!

Geda173:
> Hi! No worries at all, thanks for providing the exact vanilla OGame texts.
> 
> I will update the translation keys to be dynamic so they specifically mention the required class (General, Collector, or Discoverer) depending on the unit you are inspecting, matching vanilla 1:1.
> 
> Also, just a quick heads-up: while working on this, I noticed several other hardcoded English strings scattered across the idle queues (Shipyard, Research, Facilities, Overview, etc.). I am currently running a global sweep to fix all of those and namespace them properly for EN, IT, and NL.
> 
> I will include both the updated class texts and the global translation fixes in my next commit. I'll let you know as soon as I push!

Please open a new PR for the other translations and keep this one focused on the player class tooltips. That way we can keep the merges separate from one another.

piciolo:
sure

resources/views/ingame/ajax/object.blade.php
```diff
@@ -247,7 +247,14 @@
                                 if ($is_in_vacation_mode) {
                                     $tooltip = __('You are not able to build while in vacation mode');
                                 } elseif (!$character_class_met) {
-                                    $tooltip = __('Wrong character class!');
+                                    $tooltip = __('t_ingame.buildings.wrong_class');
+                                    $wrongClassKey = match ($object->machine_name) {
+                                        'reaper'     => 't_ingame.buildings.wrong_class_general',
+                                        'crawler'    => 't_ingame.buildings.wrong_class_collector',
+                                        'pathfinder' => 't_ingame.buildings.wrong_class_discoverer',
+                                        default      => 't_ingame.buildings.wrong_class_general',
```
Geda173: If everything goes well, players would never see the default message. From a code readibility point this can become confusing: why is the general class error message the default? I would propose to add `'default' => 't_ingame.buildings.wrong_class'` and have it be something like `You do not have the required character class for this building` or something like that.

resources/views/ingame/ajax/object.blade.php
```diff
@@ -247,7 +247,14 @@
                                 if ($is_in_vacation_mode) {
                                     $tooltip = __('You are not able to build while in vacation mode');
                                 } elseif (!$character_class_met) {
-                                    $tooltip = __('Wrong character class!');
+                                    $tooltip = __('t_ingame.buildings.wrong_class');
```
Geda173: Something here isn't properly working. When I have no class selected, it shows me the raw translation string: 

Image

With the wrong class selected, I still get the raw translation string. So this error is not confined to having no player class selected.

Geda173 (changes requested):
Hi @piciolo,

I have added some comments below.

resources/views/ingame/ajax/object.blade.php
piciolo: Just pushed a quick fix:

Restored the generic wrong_class key as the default for the match block.

Added strtolower($object->machine_name) so it catches everything (like 'crawler' vs 'Crawler') without case sensitivity getting in the way.

I couldn't trigger the Crawler specifically on my local setup to test it, but with this new logic, it should be all good. Let me know what you think! Thanks.

Geda173 (approved):

resources/views/ingame/shipyard/unit-item.blade.php
```diff
@@ -44,7 +47,8 @@
         sprite_medium medium
         @else
         sprite_small small
-        @endif{{ $building->object->class_name }}">
+        @endif{{ $building->object->class_name }}"
+        title="{{ $building->object->title }}">
```
Geda173: This title attribute on the sprite `<span>` is unnecessary. The parent `<li>` already handles the tooltip. The custom tooltip system intercepts at the `<li>` level so this attribute has no effect, but it's dead markup that shouldn't be here. Please remove it.

resources/views/ingame/shipyard/unit-item.blade.php
```diff
@@ -22,7 +22,10 @@
     title="{{ $building->object->title }}<br/>{{ __('t_ingame.buildings.requirements_not_met') }}"
     @elseif (!$building->character_class_met)
         data-status="disabled"
-    title="{{ $building->object->title }}<br/>{{ __('t_ingame.buildings.wrong_class') }}"
+    @php
+        $wrongClassKey = ['reaper' => 'wrong_class_general', 'crawler' => 'wrong_class_collector', 'pathfinder' => 'wrong_class_discoverer'][strtolower($building->object->machine_name)] ?? 'wrong_class';
```
Geda173: `object.blade.php` uses a match statement for the same mapping logic. Please use the same approach here for consistency.

Geda173 (changes requested):

resources/views/ingame/shipyard/unit-item.blade.php
```diff
@@ -1,5 +1,13 @@
 @props(['building'])
 @php /** @var OGame\ViewModels\UnitViewModel $building */ @endphp
+@php
+    $wrongClassKey = match (strtolower($building->object->machine_name)) {
+        'reaper' => 'wrong_class_general',
+        'crawler' => 'wrong_class_collector',
+        'pathfinder' => 'wrong_class_discoverer',
+        default => 'wrong_class',
```
Geda173, line 8: These use short keys + prefix concatenation, whereas `object.blade.php:251–254` stores full keys and passes them directly to `__()`. Both work, but worth aligning next time these files are touched.

Geda173 (approved):

lanedirt (approved):

PR #1323, opened 2026-03-26
Laravel 13 Upgrade

jackbayliss:
This upgrades to 13.x 

- Caught up the config / session 
- Upgraded composer packages
- Added Coordinate::class to serialize_classes as its used in getHighscorePlayers
- Added some stan neon bits, cause looks like stan doesnt like some DB raw bits, but seems to run fine.

### DIFF
--- a/config/session.php
+++ b/config/session.php
@@ -32,7 +32,7 @@ return [
     |
     */
 
-    'lifetime' => (int) env('SESSION_LIFETIME', 120),
+    'lifetime' => (int) env('SESSION_LIFETIME', 720),
 
     'expire_on_close' => env('SESSION_EXPIRE_ON_CLOSE', false),
 


jackbayliss:
@Geda173 Would appreciate some 👀 here when you get time, feel free to make any adjustments you see fit as I might not be around 👍🏻  

Seems to work locally for me.

Geda173:
> @Geda173 Would appreciate some 👀 here when you get time, feel free to make any adjustments you see fit as I might not be around 👍🏻
> 
> Seems to work locally for me.

I think you are the expert here. The only things I noticed are the changes from `_session` to `-session` as well as `_cache_` to `-cache-`, which would orphan all the cached data and log users out in existing deployments, but that can hardly be called a breaking change 😂 

I quickly searched for more `serializable_classes`, which might need to be added. `Coordinate` is already listed and from my cursory search the list seems to be complete, as all other objects stored are either a scalar or a plain array of scalars.

So yeah, looks good to me!

PR #1326, opened 2026-03-27
Fix FK constraint violation when deleting a player with espionage reports

### LINKED ISSUE #1325: [BUG] PlayerService::delete() missing cleanup for espionage_reports and battle_reports causes FK constraint violation
**Describe the bug**
#1301 exposed an issue. When `PlayerService::delete()` is called for a user who has been the target of espionage or attack missions, the deletion fails with:

> SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row:                                                                                                                  
> a foreign key constraint fails (`laravel`.`espionage_reports`,                                                                                                                                                CONSTRAINT `espionage_reports_planet_user_id_foreign`                                                                                                                                                        
> FOREIGN KEY (`planet_user_id`) REFERENCES `users` (`id`))

Both `espionage_reports` and `battle_reports` store the target player's user ID in a `planet_user_id` column with a foreign key to `users.id`. Neither table defines `ON DELETE CASCADE`, so deleting a user that has  incoming espionage or battle reports fails at the database level. `PlayerService::delete()` never cleans up these records.

The `ogamex:dev:seed-users` command calls `PlayerService::delete()` on existing test users before re-creating them. This matters because the PR preview re-deployment workflow stops the stack without the `-v` flag. The database volume survives between deployments. When a new commit is pushed to a PR, the stack is stopped and restarted, migrations run, and then `ogamex:dev:seed-users` tries to delete the users left over from the previous deployment. If the preview has been actively used (espionage probes, attacks), residual `espionage_reports` or `battle_reports` rows block the deletion.

The final teardown on PR close uses `-v`, which wipes the volume entirely, so that path is unaffected. This only breaks on re-deployments of an already-used preview.

To prevent this in the future, we should add the following to `PlayerService::delete()` in `app/Services/PlayerService.php`, before the user is deleted:

```
// Delete espionage reports where this player was the target.                                                                                                                                                
EspionageReport::where('planet_user_id', $this->getId())->delete();
                                                                                                                                                                                                               
// Delete battle reports where this player was the target.
BattleReport::where('planet_user_id', $this->getId())->delete();    
```

Geda173:
## Description
`espionage_reports.planet_user_id` had a plain non-cascading FK to `users.id`. Deleting a user who had been the target of espionage missions would therefore fail with an integrity constraint violation. This migration drops that constraint and re-adds it as `ON DELETE SET NULL`, consistent with how `battle_reports` already behaves.
                                                                                                                                                                                                               
The result: when a player is deleted, any espionage reports collected against them are preserved for the spy. The full snapshot (resources, ships, defenses) remains intact. This matches real OGame behaviour where reports survive a player's departure.
No changes to `PlayerService::delete()`, as initially proposed, were required. The database handles it automatically.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #1325 

## Checklist

- [X] **Testing:**

## Additional Information
The FK violation currently surfaces when the PR preview re-seeds test users after active testing. It will become a much more serious problem once automatic deletion of inactive accounts is introduced. Inactive players are precisely the ones most likely to have been farmed, making the overlap near-universal.

The branch is unfortunately incorrectly named, since I initially thought modifying `PlayerService::delete()` was the way to go.

### DIFF
--- a/tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php
+++ b/tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php
@@ -3,7 +3,8 @@
 namespace Tests\Feature\FleetDispatch;
 
 use Exception;
-use Illuminate\Support\Carbon;
+use Illuminate\Support\Facades\Date;
+use OGame\Factories\PlayerServiceFactory;
 use OGame\GameMissions\EspionageMission;
 use OGame\GameObjects\Models\Units\UnitCollection;
 use OGame\Models\EspionageReport;
@@ -348,7 +349,7 @@ class FleetDispatchEspionageTest extends FleetDispatchTestCase
         $fleetMissionId = $fleetMission->id;
 
         // Advance time by 5 seconds.
-        $fleetParentTime = Carbon::getTestNow()->addSeconds(5);
+        $fleetParentTime = Date::getTestNow()->addSeconds(5);
         $this->travelTo($fleetParentTime);
 
         // Cancel the mission
@@ -382,7 +383,7 @@ class FleetDispatchEspionageTest extends FleetDispatchTestCase
         $this->playerSetAllMessagesRead();
 
         // Advance time by amount of minutes it takes for the return trip to arrive.
-        $this->travelTo(Carbon::createFromTimestamp($fleetMission->time_arrival));
+        $this->travelTo(Date::createFromTimestamp($fleetMission->time_arrival));
 
         // Do a request to trigger the update logic.
         $response = $this->get('/overview');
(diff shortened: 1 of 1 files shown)


tests/Feature/FleetDispatch/FleetDispatchEspionageTest.php
```diff
@@ -382,7 +383,7 @@ public function testDispatchFleetRecallMission(): void
         $this->playerSetAllMessagesRead();
 
         // Advance time by amount of minutes it takes for the return trip to arrive.
-        $this->travelTo(Carbon::createFromTimestamp($fleetMission->time_arrival));
+        $this->travelTo(Date::createFromTimestamp($fleetMission->time_arrival));
```
lanedirt, line 386: I see you changed `Carbon` to `Date` here. I just checked the codebase and I see we're already using both `Carbon` and `Date` interchangeably in multiple places. `travelTo` accepts both so it works, but I think it would be good if we decide on one convention for maintainability purposes and refactor other tests too to use the one convention we decide on. Do you have an opinion on this?

@jackbayliss pinging you here as well: do you by any chance know the latest best practices in Laravel regarding use of `Date` vs `Carbon`? I vaguely remember that `Carbon` was the default a few years ago but I could be wrong, as I couldn't find anything in the Laravel docs at this moment regarding best practices.

Geda173: The `Carbon` class change was made by Rector automatically; it's enforcing a project-wide rule that's already configured in `rector.php`.  If we ran rector across the full test suite, it would be flagged and exchanged everywhere.

lanedirt: > The Carbon class change was made by Rector automatically; it's enforcing a project-wide rule that's already configured in rector.php

Ah gotcha, wasn't aware of that. Then LGTM! We can indeed later run rector across the full test suite to ensure consistency.

PR #1375, opened 2026-04-12
Implement password reset and email flows with SMTP integration

### LINKED ISSUE #1327: Add Email system: SMTP configuration, password reset UI, and notification emails
OGameX currently has no functioning email system. Several user-facing workflows (password reset, forgotten email) depend on being able to send email, but the SMTP configuration is incomplete and the UI entry points are non-functional stubs.

Current State

- `.env.example` has most mail keys (`MAIL_DRIVER`, `MAIL_HOST`, `MAIL_PORT`, `MAIL_USERNAME`, `MAIL_PASSWORD`, `MAIL_ENCRYPTION`) but is missing `MAIL_FROM_ADDRESS` and `MAIL_FROM_NAME`
- The login modal already has "Forgot your password?" and "Forgot your email address?" links, but both point to `href="#"`
- Laravel Fortify has password reset routes registered (`/forgot-password`, `/reset-password/{token}`), but no custom views exist for these routes that match the game's visual style, and no Mailable classes exist to actually send the email   

What Needs to Be Done

1. SMTP configuration: Add `MAIL_FROM_ADDRESS` and `MAIL_FROM_NAME` to `.env.example`
2. Password reset views: Create styled views for the forgot-password form and the reset-password form under resources/views/outgame/, matching the existing login page design
3. Wire up the links: Update `href="#"` on the "Forgot your password?" link in the login modal to point to `/forgot-password` as well as "Forgot your email address?" pointing to `/forgot-email`
4. Mailable class: Implement a `ResetPasswordMail`/ `RetrieveEmailMail` and hook it into Fortify's password reset flow
5. Documentation: Add a note to the README or `.env.example` about how to configure an external SMTP provider
6. Optionally, a `TicketReplyMail` for support ticket notifications (TODO)

Notes

Any SMTP provider supported by Laravel's mail system works (Gmail via `smtp.gmail.com:587`, Mailgun, Postmark, Resend, etc.). The infrastructure is already in place via Laravel. Only configuration and UI are missing.

piciolo:
---

## Description

This PR implements a complete SMTP email system for OGameX, covering two core account-recovery flows: **password reset** and **email retrieval**.

### Changes by file:

**`app/Mail/ResetPasswordMail.php`** *(new)*
Mailable class for the password reset flow. Uses `Envelope`/`Content` API (Laravel 10+), reads the reset URL and username from constructor arguments, and renders the branded `outgame.mail.reset-password` Blade view.

**`app/Mail/RetrieveEmailMail.php`** *(new)*
Mailable class for the "forgot email" flow. Sends a masked version of the user's email address (e.g. `pi***@gmail.com`) along with a login link, using the branded `outgame.mail.retrieve-email` Blade view.

**`app/Http/Controllers/ForgotEmailController.php`** *(new)*
Controller for the `/forgot-email` route. Accepts a username, looks up the associated account, and sends the masked email hint via `RetrieveEmailMail`. Always returns the same success message regardless of whether the username exists, to prevent user enumeration.

**`app/Models/User.php`**
Added `sendPasswordResetNotification($token)` override to replace Laravel's default plain-text notification with the branded `ResetPasswordMail`. Builds the full reset URL (including `email` parameter) before passing it to the Mailable.

**`app/Providers/FortifyServiceProvider.php`**
Registered custom Blade views for Fortify's password reset link request page (`outgame.forgot-password`) and reset password page (`outgame.reset-password`), replacing the default Fortify responses which were previously throwing `BindingResolutionException`.

**`resources/views/outgame/forgot-password.blade.php`** *(new)*
Branded outgame page for requesting a password reset link. Matches the login page visual style (space background, OGameX card design).

**`resources/views/outgame/reset-password.blade.php`** *(new)*
Branded outgame page for submitting a new password after clicking the reset link from email.

**`resources/views/outgame/forgot-email.blade.php`** *(new)*
Branded outgame page for the "forgot email" flow — accepts a username and shows a confirmation message after submission.

**`resources/views/outgame/mail/reset-password.blade.php`** *(new)*
HTML email template for password reset. Table-based layout for email client compatibility. Matches the login page color palette (`#22303f` card, `#619fc8` blue headings, `#1a2530` header). OGameX logo embedded as base64 data URI so it renders correctly in all mail clients (Gmail, Outlook, etc.) without depending on `APP_URL` being publicly accessible.

**`resources/views/outgame/mail/retrieve-email.blade.php`** *(new)*
HTML email template for the email retrieval flow. Same visual style as the reset-password template. Displays the masked email address in a highlighted inner box.

**`resources/views/outgame/layouts/main.blade.php`**
Replaced `href="#"` placeholder links in the login modal with proper `route()` calls: `route('password.request')` for forgot password and `route('password.email-lookup')` for forgot email.

**`resources/js/outgame/b55eb79922e157d28e811c7452ab10.js`**
Removed legacy JS handlers that intercepted clicks on `#pwLost` / `#emailLost` and redirected them to `https://{server}/game/reg/mail.php` and `https://{server}/game/index.php?page=standalone&component=retrieveEmail` (original OGame live server URLs). Replaced with internal route navigation.

**`routes/web.php`**
Added two routes for the forgot-email flow:
- `GET /forgot-email` → `ForgotEmailController@show` (named `password.email-lookup`)
- `POST /forgot-email` → `ForgotEmailController@send` (with `throttle:5,1` middleware)

**`config/auth.php`**
Made the password reset throttle configurable via `AUTH_PASSWORD_RESET_THROTTLE` env variable (default: 60 seconds) instead of hardcoded value.

**`database/migrations/2026_04_05_193349_create_password_reset_tokens_table.php`** *(new)*
Creates the `password_reset_tokens` table required by Laravel Fortify's password reset flow (`email` PK, `token`, `created_at`). This table was missing from the project's migration history.

**`.env.example`**
Added SMTP-related environment variables:
- `MAIL_MAILER=smtp` — critical: without this key Laravel defaults to the `log` driver, silently dropping all emails
- `MAIL_FROM_ADDRESS`, `MAIL_FROM_NAME`
- `AUTH_PASSWORD_RESET_THROTTLE=60`
- Comments for common SMTP providers (Gmail, Mailgun, SES)

---

### Type of Change:
- [x] Bug fix *(missing `password_reset_tokens` table, broken forgot-password link, logo not rendering in emails)*
- [x] Feature enhancement *(complete SMTP email system with branded templates)*
- [ ] Documentation update
- [ ] Other

---

## Related Issues
N/A

---

## Checklist

- [ ] **Automated Refactoring:** Rector has been run and no outstanding issues remain.
- [ ] **Testing:**
- [x] **CSS & JS Build:** CSS and JS assets compiled via `npm run dev` after changes to `b55eb79922e157d28e811c7452ab10.js`.

---

## Additional Information

**Email delivery verified:** both `ResetPasswordMail` and `RetrieveEmailMail` confirmed delivered via Gmail SMTP (`smtp.gmail.com:587`, TLS).

**Logo rendering fix:** the OGameX icon in email templates is now embedded as a base64 data URI. This avoids the issue where external mail servers (Gmail, Outlook) cannot load images from `http://localhost` during development, or from servers with self-signed certificates.

**No breaking changes.** The `password_reset_tokens` migration must be run on existing installations: `php artisan migrate`.

image
image

image
image


resources/views/ingame/chat/index.blade.php
Geda173, line 1: Both changes in `resources/views/ingame/chat/index.blade.php` and `resources/views/ingame/layouts/main.blade.php` are unrelated to your PR and should not be in the diff.

.env.example
```diff
@@ -37,12 +37,21 @@ REDIS_HOST=127.0.0.1
 REDIS_PASSWORD=null
 REDIS_PORT=6379
 
+# Set to 0 to disable throttle during local development (default: 60 seconds)
+AUTH_PASSWORD_RESET_THROTTLE=60
+
+MAIL_MAILER=smtp
 MAIL_DRIVER=smtp
```
Geda173: The PR adds `MAIL_MAILER=smtp` but leaves the pre-existing `MAIL_DRIVER=smtp` directly below it. `MAIL_DRIVER` has been deprecated since Laravel 7; `MAIL_MAILER` is the correct key. The old one should be removed to avoid confusion.

resources/views/outgame/mail/reset-password.blade.php
```diff
@@ -0,0 +1,124 @@
...
+        img { border: 0; outline: none; text-decoration: none; -ms-interpolation-mode: bicubic; }
+        a { color: #619fc8; text-decoration: none; font-weight: bold; }
+        a:hover { color: #91b0c4; }
+    </style>
+</head>
+<body style="margin:0;padding:0;background-color:#000000;">
+
+<!-- Outer wrapper -->
+<table width="100%" border="0" cellpadding="0" cellspacing="0"
+       style="background-color:#000000;background-image:url('{{ config('app.url') }}/img/outgame/1867da5b5f8769b547bb91d88bb4f8.jpg');background-repeat:no-repeat;background-position:center top;">
```
Geda173: CSS `background-image` with an external URL is ignored by virtually all major email clients - Gmail strips it entirely, Outlook has never supported it, and Apple Mail blocks it unless the user explicitly loads remote content. This background will not render for most recipients. Since the logo was already embedded as base64 in a follow-up commit, the same approach should be used here, or the background mage should simply be dropped from the email layout.

resources/views/outgame/mail/retrieve-email.blade.php
```diff
@@ -0,0 +1,119 @@
...
+        table { border-collapse: collapse; mso-table-lspace: 0pt; mso-table-rspace: 0pt; }
+        img { border: 0; outline: none; text-decoration: none; -ms-interpolation-mode: bicubic; }
+        a { color: #619fc8; text-decoration: none; font-weight: bold; }
+        a:hover { color: #91b0c4; }
+    </style>
+</head>
+<body style="margin:0;padding:0;background-color:#000000;">
+
+<table width="100%" border="0" cellpadding="0" cellspacing="0"
+       style="background-color:#000000;background-image:url('{{ config('app.url') }}/img/outgame/1867da5b5f8769b547bb91d88bb4f8.jpg');background-repeat:no-repeat;background-position:center top;">
```
Geda173: CSS `background-image` with an external URL is ignored by virtually all major email clients - Gmail strips it entirely, Outlook has never supported it, and Apple Mail blocks it unless the user explicitly loads remote content. This background will not render for most recipients. Since the logo was already embedded as base64 in a follow-up commit, the same approach should be used here, or the background mage should simply be dropped from the email layout.

Geda173 (changes requested):
Hi @piciolo,

thank you for your contribution. Having this in the project will make the user experience better! The core implementation is well-structured: username enumeration protection in `ForgotEmailController`, correct `sendPasswordResetNotification` override on the User model, clean Mailable classes, and the throttle configurable via `AUTH_PASSWORD_RESET_THROTTLE` env is nice to have.

I have added a couple change requests below. Furthermore I would ask that you add some form of test validation for your PR. The forgot-password and forgot-email flows have no feature test coverage. Given the project has thorough tests for other auth-adjacent flows, these should have equivalent coverage. At minimum: happy path (email sent), unknown username/email (silent success, no leak), throttle enforcement, and token expiry/invalid token on reset.

piciolo:
---

## Summary

Full implementation of SMTP email system for account recovery flows: **password reset** (via Laravel Fortify) and **email retrieval** (custom flow by username lookup).

### New features
- **Password reset flow** — integrated with Fortify: custom branded views for `/forgot-password` and `/reset-password`, `ResetPasswordMail` Mailable class, `User::sendPasswordResetNotification()` override
- **Forgot email flow** — custom `ForgotEmailController`: accepts username, sends masked email hint (e.g. `pi***@gmail.com`) via `RetrieveEmailMail`, no username enumeration (always returns success)
- **Branded HTML email templates** — table-based layout for email client compatibility, OGameX card design with dark header, white outer background, black copyright text
- **i18n support** — all UI and email strings added in EN, IT, NL (`t_external.php`)

### Bug fixes
- **Legacy JS intercept removed** — `#pwLost` / `#emailLost` click handlers no longer redirect to `https://{server}/game/reg/mail.php` (original OGame URLs); now navigate to internal `/forgot-password` and `/forgot-email` routes
- **Missing `password_reset_tokens` table** — added migration required by Fortify's reset flow
- **`MAIL_DRIVER` deprecated** — removed from `.env.example`, replaced with `MAIL_MAILER=smtp` (correct key since Laravel 7+)
- **Email background-image removed** — Gmail strips CSS `background-image`, Outlook ignores it; replaced with solid `#ffffff` background
- **Logo removed from emails** — base64 embedded image was not rendering in all clients; removed entirely to avoid broken icon
- **Copyright text fixed** — removed duplicate `© OGameX —` prefix (translation key already contains full text), changed color to `#000000` for readability on white background

### Cleanup
- Removed unrelated changes to `resources/views/ingame/chat/index.blade.php` and `resources/views/ingame/layouts/main.blade.php` that were accidentally included in a previous version of the branch

### Configuration
- `AUTH_PASSWORD_RESET_THROTTLE` env variable added to `config/auth.php` (default: 60s, configurable)
- `.env.example` updated with `MAIL_MAILER`, `MAIL_FROM_ADDRESS`, `MAIL_FROM_NAME`, SMTP provider hints

### Tests (11 tests, 77 assertions — all passing)
- `ForgotPasswordTest` (5 tests): page load, reset link sent for existing email, no mail for unknown email, valid token reset, invalid token rejection
- `ForgotEmailTest` (6 tests): page load, email sent for existing username, no mail for unknown username, masked email validation, empty username validation

Geda173:
Thanks for addressing the previous feedback. The tests, background image fix, and rate limiter cleanup all look good.

However, before this can be merged there are two issues that need to be resolved per the [CONTRIBUTING.md](https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md):

 1. Branch must be based on `main`

> "Always create your feature branch from the latest main. Do not base a branch on another feature branch or an outdated commit."

This branch was not created from `lanedirt/OGameX `main. It was brought up to date by merging from a fork remote (`piciolo/OGame`) rather than rebasing onto upstream main. Please rebase onto the current main before resubmitting.

2. Unrelated changes must be in a separate PR

> "Do not bundle unrelated changes — even small cleanups or fixes to other areas should be submitted separately."

Commit 0b5ec53a (fix: restore chat title styling and fix sidebar player list) is included in this branch but has nothing to do with SMTP. It modifies `chat.js`, `ingame.css`, `chat/index.blade.php`, and i`ngame/layouts/main.blade.php`. It also causes ~1,473 lines of noise in the compiled `ingame.js` / `ingame.min.js`bundles, which makes the diff significantly harder to review. Note that the chat changes this commit addresses are already merged upstream via #1371  and #1372. Dropping this commit and rebasing onto main should make it disappear entirely with no loss of functionality.                           
          
Once those two things are done the diff should be clean and the PR ready to merge. Please note that repeated violations of the CONTRIBUTING.md guidelines will cause PRs, which do not meet the criteria, to be closed without comment.

piciolo:
> Thanks for addressing the previous feedback. The tests, background image fix, and rate limiter cleanup all look good.
> 
> However, before this can be merged there are two issues that need to be resolved per the [CONTRIBUTING.md](https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md):
> 
> 1. Branch must be based on `main`
> 
> > "Always create your feature branch from the latest main. Do not base a branch on another feature branch or an outdated commit."
> 
> This branch was not created from `lanedirt/OGameX `main. It was brought up to date by merging from a fork remote (`piciolo/OGame`) rather than rebasing onto upstream main. Please rebase onto the current main before resubmitting.
> 
> 2. Unrelated changes must be in a separate PR
> 
> > "Do not bundle unrelated changes — even small cleanups or fixes to other areas should be submitted separately."
> 
> Commit [0b5ec53](https://github.com/lanedirt/OGameX/commit/0b5ec53a961d98f5ffa07f58906ec591d5f2637c) (fix: restore chat title styling and fix sidebar player list) is included in this branch but has nothing to do with SMTP. It modifies `chat.js`, `ingame.css`, `chat/index.blade.php`, and i`ngame/layouts/main.blade.php`. It also causes ~1,473 lines of noise in the compiled `ingame.js` / `ingame.min.js`bundles, which makes the diff significantly harder to review. Note that the chat changes this commit addresses are already merged upstream via #1371 and #1372. Dropping this commit and rebasing onto main should make it disappear entirely with no loss of functionality.
> 
> Once those two things are done the diff should be clean and the PR ready to merge. Please note that repeated violations of the CONTRIBUTING.md guidelines will cause PRs, which do not meet the criteria, to be closed without comment.

Done

Geda173:
Thanks for the rebase and for addressing all previous feedback. The diff is now clean and all concerns are resolved.                                                                                      
Two small nits before I approve, neither blocking:                                                                                                                                                         
    
1. `routes/web.php` line 36: `ForgotEmailController` is imported between `RewardsController` and `RulesController` but alphabetically it belongs in the F-section (near `FleetController`).                         
2. `forgot-password.blade.php` and `reset-password.blade.php`: the email input fields use `type="text"`. Changing to `type="email"` gives mobile users the right keyboard and enables browser format validation for free. 

Fix those and this is good to go.

piciolo:
> Thanks for the rebase and for addressing all previous feedback. The diff is now clean and all concerns are resolved. Two small nits before I approve, neither blocking:
> 
> 1. `routes/web.php` line 36: `ForgotEmailController` is imported between `RewardsController` and `RulesController` but alphabetically it belongs in the F-section (near `FleetController`).
> 2. `forgot-password.blade.php` and `reset-password.blade.php`: the email input fields use `type="text"`. Changing to `type="email"` gives mobile users the right keyboard and enables browser format validation for free.
> 
> Fix those and this is good to go.
Fix

tiweb442:
## Production smoke test (tiweb442 / space.tiweb-host.com)

Tested on a live OGameX deployment (Jul 2026):

- [x] `/forgot-password` and `/forgot-email` render (200)
- [x] `password_reset_tokens` migration applied
- [x] SMTP (STARTTLS port 587, self-signed cert allowed) — auth succeeds; password reset flow wired to panel mail host
- [x] Login modal links point to Fortify / forgot-email routes (not legacy `mail.php`)

Related local fixes in the same batch: [#1491](https://github.com/lanedirt/OGameX/pull/1491)–[#1494](https://github.com/lanedirt/OGameX/pull/1494) (tiweb442).

Happy to help rebase or re-test if this branch needs an update against current `main`.

tiweb442:
## Clean rebase onto current `main` (CONTRIBUTING)

[#1375](https://github.com/lanedirt/OGameX/pull/1375) is currently **conflicting** with `main`. Per [CONTRIBUTING.md](https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md) (branch from latest `main`, one issue per PR), I rebased the email-only commits onto current `main` with no unrelated changes:

**Replacement PR:** https://github.com/lanedirt/OGameX/pull/1500

Includes the prior review follow-ups (tests, rate limiter, alphabetized import, `type="email"`, template cleanup). Full credit remains with @piciolo for the original implementation — happy for this to be closed in favor of #1500, or for @piciolo to take the same rebase on `SMTP` if preferred.

piciolo:
Hi @Geda173, rebased and green — ready for another look.

The branch had gone stale (main advanced ~3 months and it was CONFLICTING), so I rebased it onto current `main` and squashed the iteration commits into one. All previously agreed changes and the two final nits (import alphabetization in `routes/web.php`, `type="email"` on the reset/forgot inputs) are preserved.

Two adaptations were needed for current `main`:

- **Vite migration** — `main` moved the outgame bundle from Laravel Mix (`public/js/*`) to Vite (`public/build/assets/*`). The forgot-password / forgot-email login-modal links were reapplied to the outgame source (`resources/js/outgame/…`) and the built bundle updated accordingly.
- **phpstan level 8** — the feature tests used `User::find()` (which is `User|null`); switched to `User::findOrFail()` so the reset/lookup assertions are type-safe at level 8.

piciolo:
Hi @Geda173, rebased and green — ready for another look.

The branch had gone stale (main advanced ~3 months and it was CONFLICTING), so I rebased it onto current `main` and squashed the iteration commits into one. All previously agreed changes and the two final nits (import alphabetization in `routes/web.php`, `type="email"` on the reset/forgot inputs) are preserved.

Two adaptations were needed for current `main`:

- **Vite migration** — `main` moved the outgame bundle from Laravel Mix (`public/js/*`) to Vite (`public/build/assets/*`). The forgot-password / forgot-email login-modal links were reapplied to the outgame source (`resources/js/outgame/…`) and the built bundle updated accordingly.
- **phpstan level 8** — the feature tests used `User::find()` (which is `User|null`); switched to `User::findOrFail()` so the reset/lookup assertions are type-safe at level 8.

CI is green: `test-docker`, `phpstan` (level 8), Pint, Rector and the preview all pass. Let me know if anything else is needed.

Geda173 (commented):
Thanks for the rebase. The branch is clean now. A few things left before I can approve: 

1. The built frontend bundle needs regenerating.The file `public/build/assets/outgame-b0a7e11e.js` was edited directly instead of being rebuilt. The filename contains a hash of the file's contents, so when the contents change the filename has to change too and `public/build/manifest.json` needs to update alongside it. Running `npm run build` and committing the result should sort it out.

2. A couple of test cases from the original request are still missing.

- Rate limiting isn't tested. The limiter is currently switched off in the test environment (`FortifyServiceProvider.php`), so it can't be exercised at all. I confirmed that repeated rapid requests are never blocked during tests. Could you remove that exemption and add a test that the limit actually kicks in? As it stands, if the rate limiting broke or got removed, nothing would notice.

- Expired reset tokens aren't tested. The invalid-token case is covered, but not expiry. The behaviour works correctly, it just isn't protected by a test.

3. One small thing on `testPasswordCanBeResetWithValidToken`: it only checks that the request redirects, not that the password actually changed. The reset does work correctly, but as written the test would still pass if it silently did nothing ,which makes it worth tightening, since it's the most important test in the PR. Asserting the new password works afterwards would cover it.

4. Minor style: `ResetPasswordMail` and `RetrieveEmailMail` both declare their traits on one line (`use Queueable, SerializesModels;`). Everywhere else in `app/` uses one trait per line, so please split them to match.

Nothing here is a big lift, the implementation itself is in good shape. Happy to take another look once these are addressed.

piciolo:
Thanks for the detailed review @Geda173 — all four points are addressed in the latest push.

**1. Rebuilt the outgame bundle**
Regenerated the `outgame` bundle so its filename is a real content hash of the new contents (`outgame-b0a7e11e.js` → `outgame-8f625bec.js`) and updated `public/build/manifest.json` to point at it; the hand-edited file is gone. I verified the regeneration reproduces the existing hashes byte-for-byte for unchanged input, so only the bundle whose source actually changed gets a new hash.

**2. Rate limiting is now exercised**
Removed the `testing`-environment `Limit::none()` exemption on the `forgot-email` limiter, so the throttle is live in tests, and added `testForgotEmailRequestsAreRateLimited` — five requests pass, the sixth gets a `429`.

One wrinkle worth flagging: with the exemption gone, the limiter state leaked across test methods, because `phpunit.xml` sets `CACHE_DRIVER` (the pre-Laravel-11 name) while the framework now reads `CACHE_STORE`, so tests fall back to the persistent `database` cache instead of the intended `array` store. I kept the change local by flushing the cache in `ForgotEmailTest::setUp()` so each test starts from a clean counter, but the `phpunit.xml` env key might be worth fixing separately.

**3. Expired-token coverage**
Added `testPasswordResetFailsWithExpiredToken` — creates a token, travels past the 60-minute reset lifetime, and asserts the reset is rejected with errors while the password stays unchanged.

**4. `testPasswordCanBeResetWithValidToken` now proves the reset works**
It asserts the new password validates against the stored hash (and the old one no longer does) and that the account can log in with the new password afterwards.

**5. Trait style**
Split `use Queueable, SerializesModels;` into one trait per line in `ResetPasswordMail` and `RetrieveEmailMail`.

CI is green: both docker test jobs, phpstan level 8, Pint, Rector, and the preview deploy.

tiweb442:
#1500 has been closed — this PR is the one to review.

Geda173 (approved):
Thanks for addressing the change requests. Looks good now. I would ask for @lanedirt and/or @jackbayliss to also take a look, but this should be good to go.

nagi1:
@piciolo FYI this PR merged on the active ogamex fork https://github.com/nagi1/ogamex-next/commit/38acced3814df847f37a001e8677e99228f49eb4

piciolo:
Rebased onto current `main` — the branch was 115 commits behind. All checks are green again.

**Conflicts and how they were resolved**

- `app/Models/User.php` — `main` added `casts()` at the end of the class, this branch added `sendPasswordResetNotification()`. Both kept; imports reordered for Pint's `ordered_imports`.
- `resources/js/outgame/b55eb…js` — the minified single-line file. `main` had added the `flashTrailer` pause helper (#1578); this branch only rewrites the `#pwLost` / `#emailLost` click handlers. Resolved by taking `main`'s text and re-applying just those two substitutions.
- `tests/Feature/ForgotPasswordTest.php` / `ForgotEmailTest.php` — no git conflict, but they extended `Tests\AccountTestCase`, which `main` retired in #1607. Migrated to `IsolatedAccountTestCase`, following the `BanTest` pattern (the base class does `actingAs`, so the guest-facing tests start with `$this->post('/logout')`).
- `app/Mail/*.php` — Pint 1.30.5 now rejects `) {}` on promoted constructors (`braces_position`), so the empty bodies were split.

**Vite bundle — and a heads-up about `main`**

The bundle was regenerated from the merged sources rather than hand-patched: `outgame-1a0852d6.js` → `outgame-f5d07e49.js`, with `manifest.json` updated.

While doing that I noticed that `main`'s current bundle name does not match its own content: the committed blob of `public/build/assets/outgame-1a0852d6.js` hashes to `10de045b…`, not `1a0852d6…`. `1a0852d6` is the sha256 of the same content **with CRLF line endings**, i.e. #1578 ran `npm run build` on a Windows checkout and git then normalised the file to LF on commit (`* text=auto`). At the merge-base the name was consistent (`b0a7e11e`, built on Linux). Harmless at runtime since the manifest just points at the file, but the hash is no longer a content hash. I generated mine from the LF content, which is what CI and a Linux build see:

```
git cat-file blob :public/build/assets/outgame-f5d07e49.js | sha256sum   # -> f5d07e49…
```

**Verification**

Pint `--parallel --test` (491 files), PHPStan level 8 (`No errors`) and Rector `--dry-run` all clean against `main`'s `composer.lock`. The 13 tests in this PR pass, plus 459 Unit and 716 Feature/Console. Two failures in my local environment (`CoordinateDistanceCalculatorTest > empty systems enabled`, `FleetDispatchAcsAttackTest > combat forces dropdown hidden for non invited player`) reproduce identically on a pristine `main` checkout, so they are local DB state, not regressions. Also checked the three pages in a browser: `/forgot-password`, `/forgot-email` and `/reset-password/<token>` render in the outgame layout, the regenerated bundle's handlers attach and navigate, and the forgot-email POST produces the anti-enumeration message with a correctly masked address in the mail.

tests/Feature/ForgotEmailTest.php
```diff
@@ -0,0 +1,141 @@
...
+class ForgotEmailTest extends IsolatedAccountTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // The 'forgot-email' throttle is active during tests. Because the test
+        // cache store is shared across test methods, reset it here so every test
+        // starts from a clean rate-limit counter and stays hermetic.
+        Cache::flush();
```
jackbayliss, line 23: Not sure if we need this as it's the array driver now, so shouldnt be shared

PR #1411, opened 2026-04-28
fix(1395): unbanning lifts the 48h minimum-vacation restriction

### LINKED ISSUE #1395: [BUG] Unbanning player does not lift vacation mode restrictions
**Describe the bug**
When a player is banned, their account is put in vacation mode. When unbanning them immediately after banning them, their account cannot leave vacation mode after the minimum of 48h have passed. This can forcefully extend the ban by up to 48 hours. When a ban is lifted, the restriction to leave vacation mode should also be lifted.

piciolo:
Closes #1395

## Problem
When a player is banned, their account is put into vacation mode (which carries a hard 48h minimum). When unbanning them shortly after banning, the 48h lock remains and the player still cannot leave vacation mode — effectively extending the ban by up to 48 hours.

## Fix
In `ServerAdministrationController::unban()`, after canceling the ban, release the `vacation_mode_until` time-lock so the player can manually leave vacation mode immediately.

The vacation mode flag itself is intentionally **not** auto-disabled — the player chooses when to disable it; we only release the time-lock that the ban introduced.

```php
if ($user->vacation_mode && $user->vacation_mode_until !== null && now()->lessThan($user->vacation_mode_until)) {
    $user->vacation_mode_until = now();
    $user->save();
}
```

### DIFF
--- a/app/Http/Controllers/Admin/ServerAdministrationController.php
+++ b/app/Http/Controllers/Admin/ServerAdministrationController.php
@@ -640,6 +640,15 @@ class ServerAdministrationController extends OGameController
                 'canceled'   => true,
                 'canceled_at' => now(),
             ]);
+
+            // Lift the 48h minimum-vacation restriction that was imposed by the ban
+            // so the player can leave vacation mode immediately after being unbanned.
+            // The vacation_mode flag itself is left ON — the player chooses when to
+            // disable it; we only release the time-lock that the ban introduced.
+            if ($user->vacation_mode && $user->vacation_mode_until !== null && now()->lessThan($user->vacation_mode_until)) {
+                $user->vacation_mode_until = now();
+                $user->save();
+            }
         }
 
         return redirect()->route('admin.server-administration.index')
--- a/app/Http/Controllers/Admin/ServerAdministrationController.php
+++ b/app/Http/Controllers/Admin/ServerAdministrationController.php
@@ -720,6 +720,15 @@ class ServerAdministrationController extends OGameController
                 'canceled'   => true,
                 'canceled_at' => now(),
             ]);
+
+            // Lift the 48h minimum-vacation restriction that was imposed by the ban
+            // so the player can leave vacation mode immediately after being unbanned.
+            // The vacation_mode flag itself is left ON — the player chooses when to
+            // disable it; we only release the time-lock that the ban introduced.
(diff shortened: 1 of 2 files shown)


piciolo:
Closing — opened on the wrong fork by mistake. Will reopen on piciolo/OGame.

tiweb442:
## Non-draft alternative with test coverage

[@tiweb442](https://github.com/tiweb442) opened [#1493](https://github.com/lanedirt/OGameX/pull/1493) with the same unban logic plus `testUnbanLiftsVacationModeUntilRestriction` in `BanTest.php`.

If this draft is waiting on tests or rebase, #1493 may be easier to merge as-is. Happy to close #1493 if this PR is preferred — the controller change is identical.

tiweb442 (commented):
Looks good — setting `vacation_mode_until = now()` (not null) is the right unlock given `canDeactivateVacationMode()`, and leaving `vacation_mode` on for the player matches the ban/unban intent.

One gap: please add a regression test (e.g. the `testUnbanLiftsVacationModeUntilRestriction` from #1493) so early unban → until ≤ now stays covered.

piciolo:
Thanks @tiweb442 — I've now added the regression test directly to this PR, so #1411 carries both the controller fix and the coverage. It's also out of draft.

`tests/Feature/BanTest.php` now includes `testUnbanLiftsVacationModeUntilRestriction` (ported from #1493): after an early unban of a player still in vacation mode, `vacation_mode` stays ON while `vacation_mode_until` is released back to `now()`, so the player can leave vacation immediately — it asserts `now() >= vacation_mode_until`.

One small adaptation vs. the #1493 snippet: it uses the existing `currentUsername()` helper instead of `auth()->user()->username`, since `main` is now on PHPStan level 8 and the direct call would flag the nullable `auth()->user()` — plus an extra `assertNotNull` guard.

Verified locally on PHP 8.5 (`php artisan test tests/Feature/BanTest.php`): the full `BanTest` suite is green — **8 passed, 75 assertions**. PHPStan level 8 is clean as well.

Since the controller change is identical, this PR should now be a complete drop-in, so #1493 can be closed in favor of it — but I'll of course defer to the maintainers on which they'd rather merge. Thanks again for flagging the coverage gap!

Geda173:
Thanks for the fix!

PR #1427, opened 2026-05-12
feat: Add server-wide attack block

Geda173:
## Description
Adds a configurable attack block that prevents hostile fleet missions (attack, ACS attack, espionage, moon destruction) from being dispatched while active. Friendly missions (transport, deployment, ACS defend, etc.) are unaffected.

### Type of Change:
- [ ] Bug fix
- [X] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #[issue-number]

## Checklist

- [X] **Testing:**

## Additional Information
One thing that still needs checking is how the fleet dispatch menu behaves in OGame when an attack block is in effect. As they are fairly rare, it may take a while to get the implementation 100% correct. With the recently announceed version 13.0 of OGame we may be able to see an attack block in effect soon(TM).

### DIFF
--- a/app/Services/SettingsService.php
+++ b/app/Services/SettingsService.php
@@ -2,6 +2,7 @@
 
 namespace OGame\Services;
 
+use Illuminate\Support\Facades\Date;
 use OGame\Models\Setting;
 
 /**
@@ -13,6 +14,11 @@ use OGame\Models\Setting;
  */
 class SettingsService
 {
+    /**
+     * @var array<int>
+     */
+    public const ATTACK_BLOCKED_MISSION_TYPES = [1, 2, 6, 9];
+
     /**
      * Array of setting objects.
      *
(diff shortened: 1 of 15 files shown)


app/Services/SettingsService.php
```diff
@@ -13,6 +14,11 @@
  */
 class SettingsService
 {
+    /**
+     * @var array<int>
+     */
+    public const ATTACK_BLOCKED_MISSION_TYPES = [1, 2, 6, 9];
+
```
lanedirt: Could we instead of defining an array in this `SettingService.php`, instead add a new property to the abstract `GameMission.php` like `$blockedByServerAttackBlock = true/false;`. This way we can set it to `true` for the affected missions, keeping all flags related to the missions in the mission files itself. 

So whenever new missions are added later, it's easier to see what is configurable.

lanedirt (changes requested):
Looks good, thanks for the work on this! I have one suggestion regarding maintainability, see below. Happy to hear your thoughts.

app/Services/SettingsService.php
Geda173: I can do that but the mission set is complete. No new missions will be added to this project in the future. Given that, I think the array in `SettingsService` is actually the simpler approach here. Happy to discuss if you feel strongly about it though!

lanedirt: Ah yes you're right, we're already complete in terms of missions. So in practice this probably won’t become a big issue.

As a personal preference though I do like keeping this kind of logic closer to the mission class itself, since it makes it a bit easier later on to see all mission-related behavior/configuration in one place. Otherwise someone has to know to also check `SettingsService` to understand whether a mission is affected by the server attack block or not.

That said, I'm completely fine with leaving it as-is for now. We could always revisit it later if we ever do a broader cleanup/refactor around mission/game object logic if there's time for it.

So feel free to merge it as-is 😄

Geda173: I'll get it sorted, no need to have multiple PRs to review and then merge if we are already working on it!

lanedirt: Nice work on the refactor, looks clean! Changes look good to me!

lanedirt (approved):

Issue #146, opened 2024-05-05
Add "Destroyed Planet" logic when abandoning a planet


lanedirt:
Currently when abandoning a planet it is immediately deleted from the database. However in the official version, after abandoning a planet, it shows up as a "Destroyed Planet" in the galaxy overview.

**AFAIK the following happens around destroyed planets:**

- When planet is abandoned, it gets the "destroyed" status. A destroyed planet still shows up until 48 hours after abandonment.
  - An abandoned planet can be espionaged.
  - An abandoned planet can be attacked. Any remaining fleet or defense on the planet will still fight back.
  - Resources on the abandoned planet can be stolen via attack.
- After 48 hours, the destroyed planet gets permanently deleted.
- TODO: what happens when fleets are still attacking or spying the destroyed planet when the 48 hour timer is up?
  - Check how fleet GameMission logic is handled: does it check on arrival whether the target planet (still) actually exists?

---

Here is a screenshot for what it should look like in the galaxy overview:

<img width="715" alt="Screenshot 2024-05-05 at 14 52 47" src="https://github.com/lanedirt/OGameX/assets/6917405/6b7e6a0e-bc2c-4f79-aeda-67c20861b62e">

Geda173:
To add to that, the logic behind Destroyed Planets is a bit convoluted. The server checks planets flagged for destruction at 3:00am every day. The planet needs to have that flag for at least 24 hours. If a planet is abandoned at 2:59am, the destroyed planet will persist for 24 hours and one minute. If the planet is destroyed at 3:01am, it will persist for 47 hours and 59 minutes. A destroyed planet will thus always go through two deletion cycles.

Structures on a destroyed planet persist until its ultimate deletion. The production (even the base production!) is turned off / to zero. So pillaging is possible but it will not turn into a resource pinata for the next 24/48 hours.

While a destroyed planet can be interacted with, it cannot be done through the galaxy view, only via fleet dispatch and inputting the coordinates manually. Doing so will display the owner as "space".

Moons need to be factored in as well and there are some peculiarities to keep in mind:

- A moon belonging to a destroyed planet will be destroyed alongside the planet
- Such a moon is displayed with a red border
- They cannot be interacted with at all; fleets may not target or approach it

A fleet en route to a destroyed planet will be teleported back once the planet is deleted and the ships are still traveling **to** or **from** the planet. Sending ships to a destroyed planet thus carries significant risk.

[...]

PR #1490, opened 2026-07-16
Feature/922 delete inactive players

### LINKED ISSUE #922: Add deletion of inactive players after 35 days
**Is your feature request related to a problem? Please describe.**
#703 added the inactivity markers to player accounts with the appropriate inactivity time. After a total of 35 days of inactivity, the player account should be permanently deleted and their planets abandoned, showing up as a destroyed planet (#146).

Additionally, players who are in vacation mode are **not** exempt from inactivity markers or deletion after 35 days.

In OGame, this can be prevented by having purchased Dark Matter. As there is no intention to include paid features, this is not relevant for an implementation.

SujalRatnaTamrakar:
## Description

This PR implements automatic deletion of players who have been inactive beyond a configurable number of days (issue #922).

OGameX already *measures* inactivity (`users.time`, surfaced through `PlayerService::isInactive()` / `isLongInactive()`), but nothing ever acts on it — abandoned accounts persist forever, inflating the database and leaving dead planets cluttering the galaxy. This change closes that lifecycle gap.

**What it does:**

- **Single server setting `inactive_player_deletion_days`** (integer). `0` disables the feature and is the **default**; any value `> 0` is the inactivity threshold in days. Editable in the admin **Server Settings** panel, backed by a typed `SettingsService` getter and seeded by a migration.
- **Daily scheduled command** `ogamex:scheduler:delete-inactive-players` (registered in `routes/console.php`). It no-ops when the setting is `0`, otherwise selects users whose last activity predates the threshold and deletes them.
- **Abandon-based deletion** via a new `PlayerService::deleteInactiveAccount()`: it clears the player's own fleet missions, **abandons every planet** through `PlanetService::abandonPlanet()` (freeing galaxy slots, cascading to moons, and nulling other players' incoming missions), then permanently removes the account. Runs inside a transaction.
- **Vacation mode does NOT exempt** a player from the deletion clock (per the issue requirements).
- **Admin accounts are excluded** from automatic deletion, mirroring the existing "Administrators cannot be banned" protection, to avoid irreversibly deleting the server operator.
- A `bool $force = false` parameter was added to `abandonPlanet()` so full-account deletion can bypass the user-facing guards (last remaining planet / active fleet missions). The default keeps existing behavior unchanged.
- **`TODO (#146)` markers** are left in the planet-abandonment path where the not-yet-implemented "Destroyed Planet" logic (#146) will hook in — no attempt is made to implement #146 itself.

### Type of Change:
- [ ] Bug fix
- [x] Feature enhancement
- [ ] Documentation update

## Related Issues
Fixes #922

## Checklist

- [x] **Automated Refactoring:** Rector has been run and no outstanding issues remain. (`composer run rector` — full-tree dry-run reports no changes needed.)
- [x] **Code Standards:** Code adheres to PSR-12 coding standards. Verified with Laravel Pint. (All files touched by this PR pass `composer run cs -- --test`; see note on pre-existing unrelated failures below.)
- [x] **Static Analysis:** Code passes PHPStan static code analysis. (`composer run stan` — 651 files, no errors.)
- [x] **Testing:**
    - Relevant unit and feature tests are included or updated. (New `tests/Feature/DeleteInactivePlayersCommandTest.php` with 5 tests.)
    - Tests successfully run locally. (`php artisan test` — 1065 passed, 11,744 assertions.)
- [x] **CSS & JS Build:** CSS and JS assets are compiled using Laravel Mix if any changes are made to JS/CSS files. (N/A — only a Blade template changed; no JS/CSS source was touched, so no asset rebuild is required.)
- [x] **Documentation:** Documentation has been updated to reflect any changes made. (The new setting is self-documented via an inline help description in the admin Server Settings panel; no other docs are affected.)

## Additional Information

**Design decisions worth reviewer attention:**

- **Battle/espionage reports are intentionally preserved.** Their `planet_user_id` foreign key is already `ON DELETE SET NULL` (added in #1326 / the battle-report fix), so deleting a player nulls the reference rather than orphaning or destroying the report — the other party keeps their copy. The deletion logic deliberately does **not** delete these rows.
- **Admin exclusion** and **vacation-not-exempt** are explicit behaviors (see Description).
- The new `deleteInactiveAccount()` reuses `abandonPlanet()` rather than the existing `PlayerService::delete()` (which raw-deletes planet rows), so galaxy slots are freed and in-flight missions are handled correctly.

**Files changed:**
- `app/Services/SettingsService.php`, `database/migrations/..._add_inactive_player_deletion_setting.php`, `app/Http/Controllers/Admin/ServerSettingsController.php`, `resources/views/ingame/admin/serversettings.blade.php` — the setting.
- `app/Services/PlayerService.php` (`deleteInactiveAccount()`), `app/Services/PlanetService.php` (`$force` param + `TODO (#146)` markers).
- `app/Console/Commands/Scheduler/DeleteInactivePlayers.php`, `routes/console.php` — the scheduled command.
- `tests/Feature/DeleteInactivePlayersCommandTest.php` — tests.

**Breaking changes:** None. The feature ships disabled by default (`inactive_player_deletion_days = 0`), and the added `abandonPlanet($force = false)` parameter is optional, so existing callers are unaffected.

**Note on Pint:** Running `composer run cs -- --test` across the whole tree reports pre-existing style issues in ~19 files that this PR does not modify (e.g. `app/Models/Setting.php`, `app/Models/User.php`, several `Scheduler`/`Test` commands). Per the "one issue, one PR" guideline these were left untouched; all files changed by this PR pass Pint cleanly.

### DIFF
--- a/app/Console/Commands/Scheduler/DeleteInactivePlayers.php
+++ b/app/Console/Commands/Scheduler/DeleteInactivePlayers.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace OGame\Console\Commands\Scheduler;
+
+use Illuminate\Console\Attributes\Description;
+use Illuminate\Console\Attributes\Signature;
+use Illuminate\Console\Command;
+use OGame\Factories\PlayerServiceFactory;
+use OGame\Models\User;
+use OGame\Services\SettingsService;
+use Throwable;
+
+#[Description('Permanently deletes players who have been inactive for the configured number of days.')]
+#[Signature('ogamex:scheduler:delete-inactive-players')]
+class DeleteInactivePlayers extends Command
+{
+    /**
+     * Execute the console command.
+     */
+    public function handle(SettingsService $settings, PlayerServiceFactory $playerServiceFactory): int
+    {
+        $days = $settings->inactivePlayerDeletionDays();
+
+        // 0 = feature disabled (this is the default).
+        if ($days <= 0) {
+            $this->info('Inactive player deletion is disabled (inactive_player_deletion_days = 0).');
+
+            return Command::SUCCESS;
+        }
+
+        $cutoff = now()->subDays($days)->timestamp;
+        $deletedCount = 0;
+        $failedCount = 0;
+
+        // Admins are excluded to protect the server operator and system accounts, mirroring the
+        // existing "administrators cannot be banned" rule. Vacation mode does NOT exempt a player.
+        // The users.time column holds the last-activity UNIX timestamp.
+        User::withoutRole('admin')
+            ->whereNotNull('time')
+            ->where('time', '<', $cutoff)
+            ->chunkById(200, function ($users) use ($playerServiceFactory, &$deletedCount, &$failedCount): void {
+                foreach ($users as $user) {
+                    try {
+                        // Load with a fresh cache so the deletion acts on the player's current
+                        // planets, moons and missions rather than any stale cached state.
+                        $playerServiceFactory->make($user->id, true)->deleteInactiveAccount();
+                        $deletedCount++;
+                    } catch (Throwable $e) {
+                        $failedCount++;
+                        $this->error("Failed to delete inactive player #{$user->id}: " . $e->getMessage());
+                    }
+                }
+            });
+
+        $this->info("Deleted {$deletedCount} inactive player(s) with no activity in the last {$days} day(s).");
+
+        if ($failedCount > 0) {
+            $this->warn("{$failedCount} player(s) could not be deleted; see errors above.");
+        }
+
+        return $failedCount > 0 ? Command::FAILURE : Command::SUCCESS;
+    }
+}
--- a/app/Console/Commands/Scheduler/DeleteInactivePlayers.php
+++ b/app/Console/Commands/Scheduler/DeleteInactivePlayers.php
@@ -5,6 +5,8 @@ namespace OGame\Console\Commands\Scheduler;
 use Illuminate\Console\Attributes\Description;
 use Illuminate\Console\Attributes\Signature;
 use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Log;
+use OGame\Factories\PlanetServiceFactory;
 use OGame\Factories\PlayerServiceFactory;
 use OGame\Models\User;
 use OGame\Services\SettingsService;
@@ -17,7 +19,7 @@ class DeleteInactivePlayers extends Command
     /**
      * Execute the console command.
      */
-    public function handle(SettingsService $settings, PlayerServiceFactory $playerServiceFactory): int
+    public function handle(SettingsService $settings, PlayerServiceFactory $playerServiceFactory, PlanetServiceFactory $planetServiceFactory): int
     {
         $days = $settings->inactivePlayerDeletionDays();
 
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/Admin/ServerSettingsController.php
Geda173, line 1: Since this setting permanently deletes accounts with no undo, could we put a floor on it? Right now 1 is accepted and would wipe everyone inactive for a day on the next run and it's an easy slip. I'd be much more comfortable if there was an dialogue box prompting the user to confirm the setting.

app/Services/PlayerService.php
```diff
@@ -916,6 +916,56 @@ public function delete(): void
...
+     * the account with all associated records.
+     *
+     * Unlike delete(), planets are removed through the abandonment flow (PlanetService::abandonPlanet())
+     * so galaxy slots are freed and incoming missions from other players are handled gracefully. The
+     * abandonment guards (last remaining planet / active fleet missions) are bypassed because the whole
+     * account is being removed. Careful: this action is irreversible!
+     *
+     * @return void
+     */
+    public function deleteInactiveAccount(): void
```
Geda173: This overlaps `delete() `above it quite heavily. Five of the statements are identical. My worry is drift: whoever adds the next player-related table updates one method and not the other. Since `delete()` is only called from dev commands now, could this become the single implementation with a flag for how planets get handled? The transaction wrapper here is also something `delete()` lacks, which is another argument for consolidating on this version.

app/Console/Commands/Scheduler/DeleteInactivePlayers.php
```diff
@@ -0,0 +1,63 @@
...
+        // The users.time column holds the last-activity UNIX timestamp.
+        User::withoutRole('admin')
+            ->whereNotNull('time')
+            ->where('time', '<', $cutoff)
+            ->chunkById(200, function ($users) use ($playerServiceFactory, &$deletedCount, &$failedCount): void {
+                foreach ($users as $user) {
+                    try {
+                        // Load with a fresh cache so the deletion acts on the player's current
+                        // planets, moons and missions rather than any stale cached state.
+                        $playerServiceFactory->make($user->id, true)->deleteInactiveAccount();
```
Geda173: The player and planet factories are shared singletons with no way to clear them, so everything built for each deleted account stays in memory for the whole run. The batching limits query size but not memory. Likely a non-issue for nightly runs; I'd mainly want to know it's been tried against a large backlog, since that first run could be thousands of accounts. Most of the deployments will have only small groups of people, but something main.ogamex.dev with its many inactive accounts could be seriously in trouble.

app/Services/PlayerService.php
```diff
@@ -916,6 +916,56 @@ public function delete(): void
...
+            // Abandon every planet; each call cascades to its moon. Force bypasses the user-facing
+            // guards which do not apply during full account deletion. Abandoning frees the galaxy slot
+            // and nulls out any remaining incoming missions from other players.
+            foreach ($this->planets->allPlanets() as $planet) {
+                $planet->abandonPlanet(true);
+            }
+
+            // Delete remaining user-scoped records that no database cascade covers.
+            Message::where('user_id', $this->getId())->delete();
+            Highscore::where('player_id', $this->getId())->delete();
```
Geda173: The highscores table already deletes these rows automatically when the user goes, so this line does nothing.

app/Console/Commands/Scheduler/DeleteInactivePlayers.php
```diff
@@ -0,0 +1,63 @@
...
+                        $deletedCount++;
+                    } catch (Throwable $e) {
+                        $failedCount++;
+                        $this->error("Failed to delete inactive player #{$user->id}: " . $e->getMessage());
```
Geda173: Same pattern as `CleanupWreckFields`, so take this as optional, but since scheduled output isn't captured anywhere, an account that fails to delete every night would be completely invisible. This being the one job that deletes accounts irreversibly, a `Log::error()` alongside the console message seems worth it.

app/Console/Commands/Scheduler/DeleteInactivePlayers.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        $cutoff = now()->subDays($days)->timestamp;
+        $deletedCount = 0;
+        $failedCount = 0;
+
+        // Admins are excluded to protect the server operator and system accounts, mirroring the
+        // existing "administrators cannot be banned" rule. Vacation mode does NOT exempt a player.
+        // The users.time column holds the last-activity UNIX timestamp.
+        User::withoutRole('admin')
```
Geda173: This one is more for @lanedirt: Should moderators be excluded alongside admin? It looks like a non-issue today since that role doesn't grant any permissions yet, but if it ever becomes a real staff role those accounts are exactly the infrequent-login users this targets. Cheap to add now: withoutRole(['admin', 'moderator']). There is only a handful of mentions of the role in the codebase and it currently doesn't do anything.

Geda173 (approved):
Hi @SujalRatnaTamrakar,

thanks for your work. I have attached some comments. There isn't anything big that needs changing, it's mostly preference.

Ideally I'd like a test for the case where another player has a fleet in flight toward the inactive player's planet. That's the scenario `$force` specifically unblocks by skipping the active-missions guard, and what keeps it safe is the `planet_id_to === null` check in `GameMission::process()`, which lives outside this feature and has no idea it's load-bearing for it. If that guard ever changes, nothing here would catch it. A test asserting the attacker's fleet turns around and returns home would pin the behaviour down.

app/Console/Commands/Scheduler/DeleteInactivePlayers.php
lanedirt: @Geda173 Yes I agree, since this is being touched already, adding the moderator role exclusion too here is a good idea (even if the role itself is not actually implemented yet as of this moment).

Geda173:
PHPStan is failing on this PR, please clean that up so I can take another look.

PR #1493, opened 2026-07-19
Lift vacation time-lock when unbanning a player

tiweb442:
## Summary
When a timed ban activates vacation mode, the player gets a 48h minimum before they can leave vacation. If an admin unbans them early, the ban is lifted but `vacation_mode_until` stayed in the future — effectively extending the restriction up to 48 hours after the ban ended.

This sets `vacation_mode_until = now()` on unban when vacation is still time-locked. `vacation_mode` stays enabled until the player disables it manually.

Same approach as [#1411](https://github.com/lanedirt/OGameX/pull/1411); includes a regression test.

Fixes #1395

## Test plan
- [ ] `./vendor/bin/phpunit tests/Feature/BanTest.php`

Made with [Cursor](https://cursor.com)

PR #1492, opened 2026-07-19
Fix 500 on fleet movement during moon destruction missions

tiweb442:
## Summary
- Add missing `$friendlyStatus = Hostile` to `MoonDestructionMission`
- `/fleet/movement` calls `GameMission::getFriendlyStatus()` for every active mission; type 9 crashed with an uninitialized static property (PHP 8.2+)
- The fleet event widget was unaffected because it uses ownership-based `determineFriendly()` instead

Fixes #1446

## Test plan
- [ ] `./vendor/bin/phpunit tests/Unit/MoonDestructionMissionTest.php --filter testFriendlyStatusIsHostile`
- [ ] `./vendor/bin/phpunit tests/Feature/FleetMovementTest.php --filter testFleetMovementShowsMoonDestructionMission`
- [ ] With an active moon destruction fleet, `/fleet/movement` returns 200 instead of 500

Made with [Cursor](https://cursor.com)

PR #1491, opened 2026-07-19
Fix phalanx range check to use donut system wrap-around

tiweb442:
## Summary
- Add `CoordinateDistanceCalculator::getSystemDistance()` for shortest system distance with donut wrap-around
- Use it in `PhalanxService::canScanTarget()` instead of plain `abs()`
- Add unit tests for wrap-around scanning (490 → 5) and the calculator helper

Fixes #1472

## Test plan
- [ ] `./vendor/bin/phpunit tests/Unit/PhalanxServiceTest.php`
- [ ] `./vendor/bin/phpunit tests/Unit/CoordinateDistanceCalculatorTest.php`
- [ ] Moon at system 490 with phalanx level 4 can scan system 5 in galaxy view

Made with [Cursor](https://cursor.com)

Issue #1472, opened 2026-06-23
[BUG] Phalanx range check ignores galaxy wrap-around


Geda173:
**Describe the bug**
The sensor phalanx range check in `PhalanxService::canScanTarget()` calculates system distance using a plain abs() subtraction:
  
 ` $system_distance = abs($moon_system - $target_coordinate->system);`
      
This does not account for the galaxy's wrap-around behaviour. `CoordinateDistanceCalculator` already handles this correctly for fleet travel. If the "short way round" crosses the boundary (e.g. system 490 to system 5), it takes the wrap into account and treats the distance as 14 instead of 485.

The phalanx service should apply the same logic, otherwise a moon at system 490 with a high-level phalanx cannot scan system 5 even though it is only 14 systems away.

tiweb442:
Fix in https://github.com/lanedirt/OGameX/pull/1491 (galaxy wrap-around phalanx range + PhalanxTest fixes from #1473). Tested on production.

PR #1494, opened 2026-07-19
Fix fleet and battle messages when planets or moons are destroyed

### LINKED ISSUE #1341: [BUG] Destroyed moons break messages
**Describe the bug**
When a moon is destroyed which has an active fleet mission as the origin point, upon arrival at the destination, the sent message to the player breaks:

Image

Image

In general, destroyed or abandoned moons and planets are causing loads of issues with the messaging system. Because the entries are deleted from the database, especially combat reports cannot be generated correctly, leaving them in a corrupted state.

tiweb442:
## Summary
- Add `FleetMissionPlanetFormatter` to emit `[planet]`, `[coordinates]`, or `[debrisfield]` tags so fleet messages never store broken `[planet][/planet]` placeholders when a body was deleted
- Redirect active return/inbound missions to the parent planet before a moon is removed (not only outbound launches)
- Snapshot attacker origin name/coords/type in battle reports and use stored data when the origin planet no longer exists

Fixes #1341

## Test plan
- [ ] `./vendor/bin/phpunit tests/Unit/FleetMissionPlanetFormatterTest.php`
- [ ] Transport from moon → destroy moon mid-flight → arrival/return messages render valid text
- [ ] Attack report still shows attacker origin after attacker moon is destroyed


tiweb442:
## Related PRs from tiweb442 (same deployment / review batch)

| Issue | PR |
|-------|-----|
| #1341 Destroyed moons/planets break messages | **#1494** (this) |
| #1472 Phalanx wrap-around | [#1491](https://github.com/lanedirt/OGameX/pull/1491) |
| #1446 Fleet movement during moon destruction | [#1492](https://github.com/lanedirt/OGameX/pull/1492) |
| #1395 Unban + vacation lock | [#1493](https://github.com/lanedirt/OGameX/pull/1493) |
| #1327 Email system | [#1375](https://github.com/lanedirt/OGameX/pull/1375) |

**What this fixes:** Fleet messages no longer emit broken `[planet][/planet]` tags when `planet_id_*` is null after moon/planet deletion; active return missions are redirected to the parent planet before moon removal; battle reports snapshot attacker origin data at creation time.

**Merge order:** Independent of #1491–#1493. Touches `MoonDestructionMission.php` like #1492 but different lines (battle report snapshot vs `$friendlyStatus`).

tiweb442:
**Ready for maintainer review** — preview is live and mergeable with `main`.

@lanedirt @jackbayliss Could you review when you have a moment? Fork PRs may also need **“Approve and run workflows”** so Pint / PHPStan / Docker tests run (currently only PR Preview has completed).

Test on preview: https://pr-1494.preview.ogamex.dev

tiweb442:
CI fixes pushed:
1. PHPStan: `?int` → `int|null` on `buildAttackerPlanetSnapshot`
2. Unit tests: assign `FleetMission` attributes directly instead of constructor mass-assignment

Geda173 (changes requested):
Good fix! The FleetMissionPlanetFormatter is well-structured and the incoming-mission redirect matches real OGame behavior. One required addition:

1. **`buildAttackerPlanetSnapshot()` is untested.** I mutation-tested every behavioral change; the snapshot is the only one no test catches. Removing it entirely (returning `[]`) still passes all tests. Add a feature test that attacks from a moon, destroys the moon, then asserts the battle report still contains the attacker's original planet name and coordinates. Without it, the snapshot could be removed in a future refactor and nobody would notice.

PR #1495, opened 2026-07-19
Fix sidebar wreck field icon using wrong planet Space Dock level

### LINKED ISSUE #1397: [BUG] Sidebar wreck field icon checks the wrong planet's Space Dock level
**Describe the bug**
In `resources/views/ingame/layouts/main.blade.php`, the sidebar planet list iterates over all of a player's planets to decide whether to show a wreck field icon next to each one. The Space Dock level check at line 1788 reads: 

`$hasSpaceDock = $currentPlayer->planets->current()->getObjectLevel('space_dock') > 0;`

This checks the currently selected planet's Space Dock level instead of the planet being iterated ($`planet`). It should be:

`$hasSpaceDock = $planet->getObjectLevel('space_dock') > 0;`

The wrong planet being evaluated causes two opposite failure modes for players with multiple planets:
 
- False positive: Planet A is selected (Space Dock ≥ 1), Planet B has wreckage but no Space Dock. The icon appears next to Planet B even though the player cannot inspect or repair the wreckage from there.
- False negative: Planet B is selected (no Space Dock), Planet A has wreckage and a Space Dock. The icon is hidden next to Planet A even though the player can inspect and repair the wreckage.

The icons should only appear when the planet has a wreck field AND has a Space Dock with level >0.

tiweb442:
## Summary
In the sidebar planet list, the wreck field icon used the **currently selected** planet's Space Dock level instead of the **iterated** planet's.

Fixes #1397

## Test plan
- [ ] Two planets: wreck on A with Space Dock → icon on A when B is selected
- [ ] Wreck on A without Space Dock → no icon on A even if B has Space Dock

Related: tiweb442 #1491, #1492, #1494

### DIFF
--- a/resources/views/ingame/layouts/main.blade.php
+++ b/resources/views/ingame/layouts/main.blade.php
@@ -1838,7 +1838,7 @@ However, the Space Dock's engineers think that some of the remains can be salvag
 
                                     @if ($wreckField && in_array($wreckField->status, ['active', 'blocked']) && !$planet->isMoon())
                                         @php
-                                            $hasSpaceDock = $currentPlayer->planets->current()->getObjectLevel('space_dock') > 0;
+                                            $hasSpaceDock = $planet->getObjectLevel('space_dock') > 0;
                                             $isOwner = $wreckField->owner_player_id === $currentPlayer->getId();
                                         @endphp
                                         @if ($isOwner && $hasSpaceDock)
--- a/tests/Feature/SidebarWreckFieldIconTest.php
+++ b/tests/Feature/SidebarWreckFieldIconTest.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace Tests\Feature;
+
+use DOMDocument;
+use DOMXPath;
+use Illuminate\Support\Facades\DB;
+use OGame\Models\WreckField;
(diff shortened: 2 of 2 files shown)


Geda173 (commented):
Fix looks right. `$planet` is the `@foreach` variable from line 1730, and the two lines either side of this one (`getPlanetCoordinates()` on 1828, `isMoon()` on 1839) already use it, so this was the odd one out.

Could we get a feature test with it though? This bug is invisible to any single-planet test, because with one planet `$planet` and `planets->current()` are the same object. That's why the existing wreck field coverage didn't catch it: 32 tests across `FacilitiesWreckFieldTest` and `WreckFieldTest` all pass both with and without this fix. Nothing currently asserts on the sidebar either.

tiweb442:
@Geda173 Good call — added `SidebarWreckFieldIconTest` with two multi-planet cases:

1. Wreck + Space Dock on A, B selected without dock → icon on A
2. Wreck on A without dock, B selected with dock → no icon on A

Confirmed both fail against the old `planets->current()` check and pass with this fix.

Geda173 (approved):

Geda173:
Thank you, merging.

PR #1497, opened 2026-07-19
Block expedition missions targeting debris fields (#1355)

tiweb442:
## Summary
- Reject expedition when target type is debris field

Fixes #1355

## Test plan
- [ ] Expedition to slot 16 still works
- [ ] Expedition to debris field is rejected

### DIFF
--- a/app/GameMissions/ExpeditionMission.php
+++ b/app/GameMissions/ExpeditionMission.php
@@ -119,6 +119,10 @@ class ExpeditionMission extends GameMission
             return $parentCheck;
         }
 
+        if ($targetType === PlanetType::DebrisField) {
+            return new MissionPossibleStatus(false);
+        }
+
         // Expedition mission is only possible for position 16.
         if ($targetCoordinate->position !== 16) {
             return new MissionPossibleStatus(false);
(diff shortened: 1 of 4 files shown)


Geda173 (changes requested):
The fix itself looks correct.

The test doesn't verify anything though. `checkTargetFleet(..., false)` only asserts HTTP 200, and the helper's own comment notes the endpoint always returns 200 whether or not the mission is possible. The assertJson is inside if (`$assertSuccess)` with no else branch. I confirmed by removing the guard: the test still passes.

I did notice issues being referenced in the inline comments. By no means a blocker but the instances in which this has happened are in the single digits and may warrant a change when refactoring at some point.

tiweb442:
Thanks for the review — you were right that the failure path was a no-op.

Addressed in aa4b1d1f:
- `checkTargetFleet(..., false)` now asserts `orders[missionType] === false`
- Confirmed the debris-field expedition test fails if the guard is removed
- Fixed `FleetDispatchTransportTest::testFleetCheckToForeignPlanet*` which incorrectly expected transport-to-other-player to fail (transport is allowed; same file already has a dispatch success coverage for that)
- Dropped the `#1355` reference from the inline test comment

Geda173:
Thanks for the fix, merging.

PR #1498, opened 2026-07-19
Show energy requirement for level-0 resource buildings (#1320)

### LINKED ISSUE #1320: [BUG] Energy needed not shown on resource buildings level 0
**Describe the bug**
Resource buildings at level 0 currently do not show their energy demand in the building overlay. The energy needed information is only added after constructing the first level.

tiweb442:
## Summary
- Always compute `energy_difference` in building overlays

Fixes #1320

## Test plan
- [ ] Level-0 metal mine shows energy needed for level 1


tiweb442:
Tested on production at https://space.tiweb-host.com — deployed locally for validation. Cross-ref tiweb442/OGameX branches.

Geda173 (changes requested):
The production fix is correct: `$energy_difference` is already initialized to 0 at line 107, so dropping the if `(!empty($production_current->energy->get())) `guard just lets it compute for level-0 buildings (where current consumption is 0). Confirmed working in-game.

Requesting changes on the test, though. `MineEnergyAtLevelZeroTest` doesn't actually exercise the fix. It never touches `ObjectAjaxTrait` (the only file changed); it re-implements the formula inline and asserts on that:

```
$energyDifference = ($productionNext->energy->get() - $productionCurrent->energy->get()) * -1;
$this->assertGreaterThan(0, $energyDifference);
```

That only asserts a property of `getObjectProduction` (a level-1 mine needs energy), which was never broken. The "bug" lived in the trait's guard, and the test skips it entirely.

Could we route the assertion through the actual overlay path instead? Rendering the building overlay for a level-0 mine and asserting the response reports a non-zero energy requirement would exercise the trait and fail against the old guard. 

Fix itself is good to go once the test covers the changed code.

tiweb442:
Addressed in 8fbec8c4 — replaced the unit test with a feature test that hits `ajax/resources` for a level-0 metal mine and asserts the overlay HTML includes a non-zero `additional_energy_consumption` value (exercising `ObjectAjaxTrait`). Confirmed the new test fails against the old energy guard and passes with the fix.

PR #1290, opened 2026-03-15
[Bug] cannot send all resources #1131

piciolo:
## Description
Include `CharacterClassService` in `dispatchCheckTarget` and apply the player's deuterium consumption multiplier to each ship's `fuelConsumption` so the backend accounts for character class modifiers. Update fleet dispatcher JS to treat fuel differently for return missions (double one-way consumption) and adjust the in-page override to use `Math.floor()` / `Math.ceil()` math so the client-side reservation perfectly matches server-side rounding/truncation. Rebuild/minify assets and update `mix-manifest` accordingly. 

This directly addresses the mismatches between PHP and JS fuel calculations, ensuring that when a player clicks "All Resources", the exact required fuel is handled without triggering a "Not enough resources" error, perfectly mimicking vanilla OGame behavior.

### Type of Change:
- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update
- [ ] Other (please describe):

## Related Issues
Fixes #1102

## Checklist
Before submitting this pull request, ensure all following requirements as outlined in [CONTRIBUTING.md](https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md) are met:

- [x] **Automated Refactoring:** Rector has been run and no outstanding issues remain.
- [x] **Code Standards:** Code adheres to PSR-12 coding standards. Verified with Laravel Pint.
- [x] **Static Analysis:** Code passes PHPStan static code analysis.
- [x] **Testing:**
    - Relevant unit and feature tests are included or updated.
    - Tests successfully run locally.
- [x] **CSS & JS Build:** CSS and JS assets are compiled using Laravel Mix if any changes are made to JS/CSS files.
- [ ] **Documentation:** Documentation has been updated to reflect any changes made.

## Additional Information
Extensive client-side console logging and server-side testing were performed alongside this fix to ensure 1:1 mathematical fidelity with vanilla OGame. The frontend JS rounding now perfectly aligns with the backend's decimal truncation, allowing the exact total of available deuterium to be loaded into the cargo hold while correctly reserving the necessary fuel.

**Final Test Logs confirming exact calculation:**
```text
[#1102] getDeuteriumOnPlanetWithoutConsumption | RawDeut=6360141 | RawConsumption=287096 | Deut to leave=287096 | Max loadable=6073045
fleet:1923 [#1102] selectMaxDeuterium → deutMax=6073045 freeSpace=4000000000 amount=6073045 cargoDeuterium: 0 → 6073045
[...]

PR #1499, opened 2026-07-19
Fix cannot send all resources at max cargo (#1131)

### LINKED ISSUE #1131: [BUG] Cannot send all resources
**Describe the bug**
#1102 attempted to fix the issue but it persists. Players currently cannot send large amounts of resources at once, because the game attempts to subtract more resources than there are on the planet, blocking fleet dispatch. We probably do need to add a buffer so that the operation does not fail.

To add to that: I suspect there is an issue with ressource generation in play. When I just sent all resources from a moon to a planet, it worked without a hitch, but when sending the resources back from the planet to the moon, it fails with aforementioned error message.

It also appears to be a Deuterium constraint only. When leaving roughly 1000 Deuterium out of the cargo, it works. Maybe the Deuterium consumption isn't properly being accounted for or perhaps the game is trying to deduct it multiple times.

tiweb442:
## Summary
- Use floor/ceil in `getDeuteriumOnPlanetWithoutConsumption`
- Remove duplicate blade override

Fixes #1131

## Test plan
- [ ] Max-all resources with fractional deuterium on planet succeeds


tiweb442:
Tested on production at https://space.tiweb-host.com — deployed locally for validation. Cross-ref tiweb442/OGameX branches.

Geda173 (changes requested):
Thanks for digging into this one! The diagnosis is right (it's a rounding issue on deuterium), but there are a few problems with how the fix lands:

1. The fix doesn't actually reach players. The change was made to the source JavaScript file, but the game serves the pre-built bundle in public/build/, and that bundle wasn't rebuilt, so it still has the old code. Editing the source without running `npm run build` and committing the rebuilt bundle means nothing changes in-game.

2. It removes the fix that's currently working. Right now the rounding fix is applied by a small override in `fleet/index.blade.php` (added back in #1290). That override is the reason "send all resources" works today. This PR deletes it, so with the bundle not rebuilt, we'd be left with no fix at all and the bug would come back.

3. The new test doesn't test the fix (and currently fails). The bug is in the browser-side calculation, but the test re-does that math in PHP and just checks the backend accepts it , so it can't catch the real problem. It also fails as written, because it tries to load ~9,990 deuterium into a single small cargo that only holds 5,000.

Suggested direction:
- Either keep the blade override, or bake the fix into the source and rebuild + commit the `public/build/ `bundle, but not one without the other.
- The PHP test can't verify a browser-side calculation, so it's better to drop it (or replace it with a browser test) rather than leave a test that passes without proving anything.

Happy to help verify once it's updated.

tiweb442:
Thanks for the review — addressed in 975d497:

1. **Bundle rebuilt & committed** — source floor/ceil change is now in `public/build/assets/ingame-7716e02a.js` + updated `manifest.json`, so the served asset includes the fix.
2. **Blade override stays removed** — the working fix is baked into the served bundle rather than re-applied in `fleet/index.blade.php`.
3. **Dropped the PHP test** — it couldn't verify the browser-side calculation (and was incorrectly sized for small cargo capacity).

Ready for another look when you have a moment.

Geda173:
Thanks for the changes, but the bug seems to still be existing. You can login to the preview environment with [test1@ogamex.dev](mailto:test1@ogamex.dev) and try to send the 300k LC I added with all of the resources on the planet. Trying to dispatch the fleet fails with the error: "Not enough resources or units on the planet to send fleet."

The bundle is correctly rebuilt and served, so that part of the previous round is fine. The problem is that the fix can't work where it currently sits.

`fleet/index.blade.php` passes the deuterium to the JS already rounded:

```blade
var deuteriumOnPlanet = {{ $planet->deuterium()->getRounded() }};
```

`getRounded()` rounds to nearest, so it rounds **up**, and the deuterium column stores fractional values. With 9990.7 on the planet, the JS receives 9991, so `Math.floor()` in the bundle is a no-op and the browser sends more deuterium than actually exists.

The backend then needs `cargo + consumption <= planet deuterium`. When consumption is a whole number the request comes to exactly 9991 against 9990.7 available, and the atomic deduction fails. When consumption happens to have a large fractional part it slips through, which is probably why it appeared fixed in testing.

The fix should be at the source:

```blade
var deuteriumOnPlanet = {{ (int) floor($planet->deuterium()->get()) }};
```

Then the JS is never told about deuterium that isn't there, and the existing `Math.floor`/`Math.ceil` in the bundle stay as a harmless safety net. Note you'll need to rebuild the bundle again only if you also change the source JS, a blade-only change doesn't require it.

Worth a quick look at whether metal and crystal have the same rounding issue; they'd only show up when the cargo is exactly full.

tiweb442:
Thanks for the sharper diagnosis — that matches what we were seeing.

**Fix (blade source):** planet resources passed into fleet JS are now floored, not `getRounded()`:

```php
var metalOnPlanet = {{ (int) floor($planet->metal()->get()) }};
var crystalOnPlanet = {{ (int) floor($planet->crystal()->get()) }};
var deuteriumOnPlanet = {{ (int) floor($planet->deuterium()->get()) }};
```

Metal and crystal got the same treatment — those columns are also fractional, and the round-up would only show when cargo is exactly full. The existing `Math.floor` / `Math.ceil` in the bundle stay as a safety net; no rebuild needed for this blade-only change.

**Verified on the preview** (`af8a063b`, [pr-1499.preview.ogamex.dev](https://pr-1499.preview.ogamex.dev), `test1@ogamex.dev`):

- Recreated the 300k LC path (DB reset on redeploy) and dispatched Deployment with **all resources** — succeeded, no “Not enough resources or units…” error.
- Fractional round-up cases: metal `4999.58` and crystal `4999.59` → JS got `4999`, POSTed `4999`, both sends succeeded. Under `getRounded()` those would have been `5000` and failed.

Ready for another look when you have a moment.

tiweb442:
Added `FleetDispatchPageResourcesTest::testFleetPageEmbedsFlooredPlanetResources`. It sets fractional metal/crystal/deuterium on the current planet, GETs `/fleet`, and asserts the Blade JS vars are the **floored** integers (`4999` / `4999` / `9990`), not the `getRounded()` values (`5000` / `5000` / `9991`). Mutation-check: reverting Blade to `getRounded()` fails this test; restoring `floor()` passes.

tiweb442:
The 2026-08-15 diagnosis is addressed on the current head:

- Blade now embeds **floored** planet resources (`(int) floor(...)`) for metal/crystal/deuterium
- `FleetDispatchPageResourcesTest` asserts the JS vars are the floored integers, not `getRounded()`

CI is green. Please re-review — GitHub still shows Changes Requested from the July pass.

Geda173:
Thanks for the fix!

PR #1509, opened 2026-07-20
feat: soft-delete abandoned planets as Destroyed Planets (#146)

### LINKED ISSUE #146: Add "Destroyed Planet" logic when abandoning a planet

tiweb442:
## Summary
- Soft-flag abandoned planets/moons via `planets.destroyed` (Unix timestamp) instead of hard-deleting immediately, matching official Destroyed Planet behavior from #146.
- Keep zero production (including base income), show Destroyed Planet / `moon_c` in galaxy with no click missions, allow attack/espionage/transport/loot on destroyed planets, and block targeting destroyed moons.
- Add daily `03:00` purge (`ogamex:scheduler:cleanup-destroyed-planets`) for bodies flagged ≥24h, with fleet teleport for missions still traveling to/from the body.

## Test plan
- [ ] Abandon a colony with ≥2 planets → row remains with `destroyed > 0`, drops from planet list, production is 0
- [ ] Galaxy view shows “Destroyed Planet”, Deep space owner, empty actions; moon (if any) uses `moon_c` and is untargetable
- [ ] Manual fleet dispatch: attack/espionage/transport succeed against destroyed planet; deploy fails; destroyed moon rejected
- [ ] Run `php artisan ogamex:scheduler:cleanup-destroyed-planets` with a planet flagged <24h (kept) and ≥24h (permanently deleted + fleets teleported, including in-flight return missions)
- [ ] `php vendor/bin/phpunit tests/Feature/DestroyedPlanetTest.php tests/Feature/PlanetAbandonTest.php`

Closes #146


app/Services/PlanetService.php
```diff
@@ -238,44 +248,147 @@ public function abandonPlanet(): void
...
+                $fleetMissionService->cancelMission($mission);
+            } else {
+                // Return trip still in flight: force immediate arrival at the home planet.
+                $mission->time_arrival = $nowTs;
+                $mission->time_arrival_ms = (int) $now->valueOf();
+                if ($mission->time_holding !== null) {
+                    $mission->time_holding = 0;
+                }
+                $mission->save();
+                $fleetMissionService->processDueMissionEventsForMission($mission);
```
Geda173: This method doesn't exist. A repo-wide search finds only this call site; no definition, no trait, no `__call`, and a reflection check on `FleetMissionService` confirms it (`updateMission()` exists; this one doesn't). As written it throws a fatal "Call to undefined method" every time this branch runs (return fleet still in flight to/from the body). In the purge command the `try/catch` swallows it, so the body silently never gets deleted and retries failing nightly; in `MoonDestructionMission::processArrival` it's uncaught inside a transaction and rolls the whole destruction back. Looks like you meant `updateMission($mission)`. Please also add a test that puts an in-flight return mission through the purge. None of the current tests hit this branch, which is why CI is green.

app/GameMissions/TransportMission.php
```diff
@@ -44,6 +44,11 @@ public function isMissionPossible(PlanetService $planet, Coordinate $targetCoord
             return new MissionPossibleStatus(false);
         }
 
+        // Transport is not possible to destroyed planets/moons.
+        if ($destroyedCheck = $this->checkDestroyedTarget($targetPlanet, $targetType, false)) {
```
Geda173: Per the OGame wiki, a destroyed planet can be "ganz normal angegriffen, ausspioniert und mit Rohstofftransport unterstützt" (-> attacked normall, spied on and supported with resources). Transport is explicitly allowed (it even calls it "logically questionable" but permitted). This should be `checkDestroyedTarget($targetPlanet, $targetType, true),` matching attack/espionage/missile. Destroyed moons still stay blocked, which is correct.

tests/Feature/DestroyedPlanetTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $probes = new UnitCollection();
+        $probes->addUnit(ObjectService::getUnitObjectByMachineName('espionage_probe'), 1);
+
+        $this->assertTrue(
+            $attackMission->isMissionPossible($this->planetService, $coords, PlanetType::Planet, $fighters)->possible
+        );
+        $this->assertTrue(
+            $espionageMission->isMissionPossible($this->planetService, $coords, PlanetType::Planet, $probes)->possible
+        );
+        $this->assertFalse(
```
Geda173, line 273: This asserts transport is not possible to a destroyed planet, which locks in the deviation above. Once the guard is fixed, flip this to assert transport is possible (and maybe rename the test to reflect that attack/espionage/transport are all allowed).

app/Services/PlanetService.php
```diff
@@ -238,44 +248,147 @@ public function abandonPlanet(): void
             }
         }
 
-        // If this is a planet and has a moon, delete the moon first
+        $destroyedAt = (int) Date::now()->timestamp;
+
+        // Flag moon first with the same timestamp (destroyed alongside parent).
         if ($this->isPlanet() && $this->hasMoon()) {
-            $this->moon()->abandonPlanet();
+            $this->moon()->applyDestroyedFlag($destroyedAt);
```
Geda173, line 267: The wiki's moon section says the moon is deleted/released at 1:10, separately from the planet's 3:00 run, but the moon here is flagged with the parent's timestamp and purged in the same 3:00 job, so the two always vanish together. The wiki is internally inconsistent on this (3:00 in the description, 1:10 for moons), so this may be fine, just flagging it so it's a deliberate choice rather than an accident.

Geda173 (changes requested):
Really solid implementation.

Three things to address before merge, one of them a guaranteed crash. I also checked against the German OGame wiki's "Zerstörter Planet" article and left inline notes.

See the comments attached.

app/Services/PlanetService.php
tiweb442: Fixed — now calls `updateMission($mission)` and dropped the nonexistent `time_arrival_ms` assignment. Added `testCleanupTeleportsInFlightReturnMission` to cover this path.

app/GameMissions/TransportMission.php
tiweb442: Agreed — switched to `checkDestroyedTarget(..., true)` so transport to destroyed planets is allowed (destroyed moons still blocked).

tests/Feature/DestroyedPlanetTest.php
tiweb442: Updated — renamed to `testAttackEspionageAndTransportAllowedOnDestroyedPlanet` and asserts transport is possible.

app/Services/PlanetService.php
tiweb442: Deliberate: moons keep the parent timestamp and purge in the same 03:00 job. Added a code comment documenting that choice vs the wiki 01:10 note.

tiweb442:
Fixed the CI failure in `FleetDispatchRecycleTest`:

- `teleportFleetsAwayFromPlanet()` was recalling moon-origin outbound fleets, then nulling `planet_id_to` on delete, which left return trips with a null destination.
- Moon-origin outbound fleets are now left in flight and rebound to the parent planet (same as main’s moon-delete behavior). Return trips headed back to the dying moon are also rebound to the parent instead of being force-arrived.

Also merged latest `main` (dependency bumps). Locally verified:
- `testDispatchFleetFromMoonReturnsToPlanetWhenMoonIsDestroyed`
- `DestroyedPlanetTest` + `PlanetAbandonTest`

@Geda173 ready for another look when you have a moment — prior review notes were already addressed in `b12ea9c`.

tiweb442:
Recycle moon-return regression is fixed (passed on all 3 CI attempts).

The remaining red check was a flaky unrelated failure: `FleetDispatchAttackTest::large scale attack moon creation within attempts` (`140020` not enough resources). Attempt 1–2 passed AttackTest; attempt 3 flaked. Pushing an empty commit to retrigger CI.

tiweb442 (commented):
## PR Review: Destroyed Planet soft-delete (#146)

**Scope:** +911 / −75 across 24 files

This is a strong implementation of a long-standing gap. CI is fully green (Pint, Rector, PHPStan, tests, preview deploy). Geda173's Jul 24 feedback appears addressed in later commits; GitHub still shows `CHANGES_REQUESTED` pending a re-review from Geda173.

---

### What's done well

- **Correct two-phase lifecycle:** `markAsDestroyed()` / `applyDestroyedFlag()` for the grace period, `permanentlyDeletePlanet()` + daily `03:00` purge for hard removal — matches OGame's "visible for a while, then gone" model.
- **Clean separation of concerns:** `abandonPlanet()` → soft flag; dev shortcuts / moon-destruction / purge → `permanentlyDeletePlanet()`. That keeps player abandon, admin tools, and the scheduler on distinct paths.
- **Fleet handling is thorough:** transactional purge, outbound recall, in-flight return force-arrival, moon-origin fleet rebind to parent — the recycle regression fix (`8bc5c57`) is the kind of edge case that would bite in production.
- **Gameplay rules match the wiki:** attack / espionage / transport / missiles allowed on destroyed **planets**; deployment, ACS defend, colonisation, and all destroyed-**moon** targeting blocked.
- **Production safety:** `updateResourcesUntil()` and `updateResourceProductionStats()` short-circuit on destroyed bodies — good defense against stale non-zero rates.
- **Galaxy UX:** "Destroyed Planet", `moon_c`, Deep space owner, empty click-actions — aligned with the issue screenshot.
- **Test suite:** `DestroyedPlanetTest` (488 lines) covers abandon, galaxy UI, mission guards, combat on destroyed planets, resource accrual, purge timing, and fleet teleport/recall. Solid coverage.
- **Bonus fix:** `MissionPossibleStatus` uses `error`, not `reason` — fixes a latent bug in `startMissionSanityChecks()`.

---

### Must fix / merge blocker

**None for #1509 alone** — CI passes and Geda173's crash (`processDueMissionEventsForMission` → `updateMission`) is fixed with a regression test.

---

### Should fix / discuss before merge

**1. Cross-PR conflict with #1490 (inactive player deletion)**

These two PRs interact badly if merged naively:

| | #1509 behavior | #1490 expectation |
|---|---|---|
| Planet removal | Soft-flag (`destroyed > 0`), row kept | `assertDatabaseMissing('planets', …)` — expects immediate hard delete |
| Account deletion | `abandonPlanet()` after #1509 | Then `user->delete()` while planet rows still reference `user_id` via FK |

`planets.user_id` is a **non-nullable FK** to `users` with no `ON DELETE CASCADE`. After soft-abandon, deleting the user will fail unless ownership is cleared or planets are hard-deleted.

**Recommendation:** Merge **#1509 first**, then update #1490 to either:
- use `permanentlyDeletePlanet()` for full account wipe (if immediate removal is desired), or
- clear `user_id` on destroy + adjust tests to expect soft-flagged planets until purge (if #922 wants visible destroyed planets for inactive accounts).

Worth coordinating explicitly — this is the main integration risk.

**2. `user_id` not cleared on `applyDestroyedFlag()`**

Galaxy UI shows "Deep space", but the DB row still owns `user_id` to the abandoning player. That works for voluntary abandon (player still exists, colony slot freed via `PlanetListService` `destroyed = 0` filter), but it blocks deleting the owning user while destroyed rows exist — relevant for #922/#1490.

Consider whether abandon should null out ownership (may need a migration to make `user_id` nullable, or a sentinel "system" user).

**3. Geda173 re-review still outstanding**

All three Jul 24 items (crash fix, transport allowed, moon purge timing documented) look addressed. A formal re-approval from Geda173 would clear the stale `CHANGES_REQUESTED` state.

---

### Nits

- **24h vs 48h (#146 text):** Purge uses `destroyed <= now()->subDay()` at daily `03:00`, so retention is **24–48h** depending on abandon time relative to the job — reasonable interpretation of the issue's "~48 hours," worth one sentence in the PR description.
- **Moon 1:10 vs 3:00 wiki note:** Deliberately paired with parent — documented in code. Fine.
- **Cleanup command logging:** Errors go to `Log::error`; success count is console-only. Matching #1490's `Log::info` summary pattern would help ops.
- **No `checkDestroyedTarget` on colonisation/recycle:** Colonisation is blocked because the row still exists (tested). Recycle targets debris fields — correct as-is.

---

### Issue alignment (#146)

| Requirement | Status |
|---|---|
| Soft "Destroyed Planet" after abandon | ✅ |
| Visible in galaxy for grace period | ✅ (~24–48h) |
| Espionage / attack / loot defense | ✅ (combat test included) |
| Permanent delete after grace period | ✅ |
| Fleet behavior when body purged | ✅ (return + outbound tests) |
| Destroyed moon untargetable, `moon_c` sprite | ✅ |

---

### Verdict

**Approve for #146** — well-architected, thoroughly tested, and CI-clean. The main follow-up is **coordination with #1490** so inactive-account deletion doesn't assume immediate hard-delete or break on the `planets.user_id` FK.

Suggested merge order: **#1509 → update #1490 → merge #1490**.

tiweb442:
## Follow-up fixes (`ed6a969d`)

Pushed additional edge-case fixes on `tiweb442/feature/146-destroyed-planet` (`ed6a969d`). Summary for re-review:

### Moon fleets on parent destroy / cascade delete
When a parent planet is destroyed (or cascading-deleted), **moon-origin fleets now rebind to a living colony** instead of being lost. Fleet loss still applies on a **joint moon + planet purge** (as expected when there is no surviving body to rebind to).

### Vacation mode vs Destroyed Planet (Deep space)
Vacation mode **no longer protects destroyed (Deep space) targets** for **attack / espionage / transport**. Those missions remain allowed against destroyed planets regardless of the former owner's vacation status.

### Abandon blocking rules
Abandon is now blocked only by **own outbound fleets** from the planet **or its moon**. **Incoming foreign fleets no longer block** abandon.

### Ownership / `user_id`
Leaving `user_id` on soft-destroyed rows is **intentional for now**, pending coordination with **#1490** / **#922** (nullable ownership / inactive-account deletion).

### Tests
`DestroyedPlanetTest` updated/extended for the above — **14 tests passed locally**.

---

@Geda173 (and anyone else reviewing): please re-check these paths when you have a chance — especially moon fleet rebind, vacation + Deep space targeting, and abandon vs foreign incoming fleets. Prior `CHANGES_REQUESTED` items should still be covered; this commit is the follow-up on the remaining edge cases.

tiweb442:
## Preview smoke test (`59807dc`)

**Preview:** https://pr-1509.preview.ogamex.dev  
**Commit tested:** `59807dc139797bd9a871e4f89e7b9f680714d35e` (Pint-only follow-up to `ed6a969`; same feature behavior)  
**Deploy:** Preview bot reported Live on `59807dc` at 2026-08-15 23:43:47 UTC — deploy check passed before testing.

### Passed
| Area | Result |
|------|--------|
| HTTP / load | `/` → `/overview` → `/login` (unauth); app responds 200 after login |
| Login | `test2@ogamex.dev` / `test` OK |
| Register | Throwaway account registered successfully |
| Core pages | overview, resources, facilities, research, shipyard, defense, fleet, fleet/movement, galaxy, messages, highscore, alliance, premium, rewards — all 200 |
| Abandon guard | Sole homeworld correctly blocked: *“Cannot abandon only remaining planet.”* |
| Abandon → Destroyed Planet | Abandoned test7 colony `[1:50:8]` (cp=130) → success |
| Galaxy UI | System `1:50` position 8 shows `planetName: "Destroyed Planet"`, `playerName: "Deep space"`, `playerId: 99999`, `isDestroyed: true` |
| Fleet `check-target` to destroyed | `targetOk: true`; name `Destroyed Planet` / player `Deep space`; missions enabled: **attack (1), transport (3), espionage (6)** |
| Vacation account | `test6@ogamex.dev` logs in; galaxy shows vacation restriction messaging |

### Not covered / needs manual QA
- **03:00 purge** — permanent delete + fleet teleport after soft-destroy window
- **Moon fleet rebind** when parent is destroyed / cascading delete (needs moon + outbound/return fleets)
- **Vacation + Destroyed** — attack/espionage/transport vs Deep space while former owner is in vacation (logic is in code; not end-to-end’d here)
- **Abandon vs foreign incoming fleets** — should no longer block
- **Joint moon+planet purge** fleet loss when no surviving body to rebind to
- Full browser UX (JS galaxy icons/tooltips, mini-fleet from galaxy row)

### Notes
- No code changes made during this smoke pass.
- Shared preview seed: test7 lost colony `[1:50:8]` (now Destroyed Planet). Preview reseeds on redeploy / destroy-on-close.
- Unit-test docker check was still pending when this comment was posted; Pint / PHPStan / Rector / prod docker / preview deploy were green.

tiweb442:
The Jul 24 crash/transport comments are fixed on the current head (including `updateMission()` and transport-allowed tests). The remaining thread is the documented product choice: moons keep the parent timestamp and purge in the same 03:00 job.

CI is green. Please re-review — GitHub still shows Changes Requested from the earlier pass.

Geda173 (approved):
Thanks for the changes. Approving.

One coordination note before merge: #1490 expects hard-deleted planet rows, while this PR soft-flags them with user_id still set on a non-nullable FK. Merge this first, then adapt #1490 (either use permanentlyDeletePlanet() for account wipes or clear ownership on destroy), otherwise user deletion will trip over the FK.

lanedirt (approved):

PR #1511, opened 2026-07-20
Show max fleet slots message in galaxy tooltips (#666)

### LINKED ISSUE #666: Add max fleet slots message to galaxy planet and moon tooltip hovers
When maximum amount of fleet slots are in use, the tooltips should instead of showing the possible mission types, look like this:

Image
Image

Debris field tooltip too:

Image

Also add it to colonisation tooltip (and add colonisation astrophysics warning too if it doesn't exist yet):

Image

tiweb442:
## Summary
- Wire real `usedFleetSlots`, `maximumFleetSlots`, `canFly`, and `hasAdmiral` into the galaxy AJAX response so existing planet/moon/debris tooltips show "No fleet slots available" when slots are full
- Enrich empty-slot colonisation tooltips with astrophysics and max-fleet-slots warnings, and disable the colonise link when colonisation is not allowed
- Add GalaxyTest coverage for slot counts, `canFly: false`, and the astrophysics colonisation warning

Closes #666

## Test plan
- [ ] Fill all fleet slots, open galaxy, hover a foreign planet/moon — tooltip shows "No fleet slots available" (+ Hire admiral) instead of mission links
- [ ] Same check for a debris field tooltip
- [ ] With astrophysics at 0, hover an empty-slot colonise icon — tooltip includes astrophysics warning and icon is inactive
- [ ] With slots full and astrophysics researched, colonise tooltip shows no-fleet-slots message
- [ ] Confirm used/total fleet slot counters in the galaxy header match the fleet page
- [ ] `php artisan test --filter=GalaxyTest`

### DIFF
--- a/app/Http/Controllers/GalaxyController.php
+++ b/app/Http/Controllers/GalaxyController.php
@@ -75,8 +75,8 @@ class GalaxyController extends OGameController
             'espionage_probe_count' => $planet->getObjectAmount('espionage_probe'),
             'recycler_count' => $planet->getObjectAmount('recycler'),
             'interplanetary_missiles_count' => $planet->getObjectAmount('interplanetary_missile'),
-            'used_slots' => 0,
-            'max_slots' => 1,
+            'used_slots' => $player->getFleetSlotsInUse(),
+            'max_slots' => $player->getFleetSlotsMax(),
             'max_galaxies' => $settingsService->numberOfGalaxies(),
             'is_in_vacation_mode' => $player->isInVacationMode(),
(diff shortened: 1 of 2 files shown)


Geda173 (approved):
Looks good, clean approach.

One optional follow-up (non-blocking): `canColonize` doesn't account for the player's colony-slot limit, so with astrophysics researched and a free fleet slot it can read true even at the planet cap. It's strictly better than the previous behavior (which allowed colonise whenever a ship was present), so this is fine to merge as-is, maybe a separate issue if we want to tighten it later.

tiweb442:
Thanks @Geda173 — opened a follow-up for the colony-slot limit check: #1516

PR #1515, opened 2026-07-23
Upgrade PHPStan analysis to level 8

### LINKED ISSUE #121: Upgrade PHPstan code analysis level to 8

Geda173:
## Description
This PR raises the PHPStan level from 7 to 8, enabling full null-safety enforcement across the codebase.

### Type of Change:
- [ ] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update
- [x] Other (please describe):

## Related Issues
Fixes #121 

## Checklist

- [x] **Testing:**

## Additional Information

### DIFF
--- a/app/Events/ChatMessageSent.php
+++ b/app/Events/ChatMessageSent.php
@@ -50,12 +50,14 @@ class ChatMessageSent implements ShouldBroadcast
      */
     public function broadcastWith(): array
     {
+        $createdAt = $this->message->created_at;
+
         $data = [
             'id' => $this->message->id,
             'senderId' => $this->message->sender_id,
             'senderName' => $this->message->sender->username,
             'text' => e($this->message->message),
-            'date' => $this->message->created_at->timestamp,
+            'date' => $createdAt !== null ? $createdAt->timestamp : 0,
         ];
 
         if ($this->message->alliance_id) {
--- a/app/Console/Commands/Dev/PreviewSeedUsers.php
+++ b/app/Console/Commands/Dev/PreviewSeedUsers.php
@@ -2,9 +2,9 @@
 
 namespace OGame\Console\Commands\Dev;
 
+use Exception;
 use Illuminate\Console\Attributes\Description;
 use Illuminate\Console\Attributes\Signature;
-use Exception;
 use Illuminate\Console\Command;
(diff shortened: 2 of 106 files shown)


lanedirt (approved):
LGTM, nice to get more test and compile time certainty!

app/Events/ChatMessageSent.php
```diff
@@ -50,12 +50,14 @@ public function broadcastOn(): array
...
     {
+        $createdAt = $this->message->created_at;
+
         $data = [
             'id' => $this->message->id,
             'senderId' => $this->message->sender_id,
             'senderName' => $this->message->sender->username,
             'text' => e($this->message->message),
-            'date' => $this->message->created_at->timestamp,
+            'date' => $createdAt !== null ? $createdAt->timestamp : 0,
```
jackbayliss, line 60:
```suggestion
            'date' => $createdAt?->timestamp ?? 0,
```
Would this work? may read nicer is my thought- could be the same elsewhere. (probably no need for the $createdAt var then..?)

Geda173: Good call, I tried this and it reads much nicer. Unfortunately PHPStan at level 8 won't accept it here. The two forms actually trigger contradictory errors on this line:

- `$this->message->created_at->timestamp` → `property.nonObject`: "Cannot access property $timestamp on Carbon|null" -> treats it as nullable.
- `$createdAt?->timestamp ?? 0` (your suggestion) → `nullsafe.neverNull`: "Using nullsafe... is unnecessary. Use -> instead." -> treats it as never-null.

So `->` fails (possibly null) and `?->` fails (unnecessary). You can't satisfy both. It comes from `treatPhpDocTypesAsCertain: false` in our config: the plain-access check honors the phpdoc `Carbon|null`, but the "unnecessary nullsafe" rule ignores that setting and sees the native non-null type. The explicit `!== null` ternary is the only form that passes both, which is why I went with it. Same reasoning applies to the other Carbon spots you mentioned.

We can revisit should we ever decide to flip `treatPhpDocTypesAsCertain` back on.

jackbayliss: Cool! Thanks for checking :)

jackbayliss (approved):

PR #1502, opened 2026-07-20
feat(152): implement Technology tab in "?" info popup

tiweb442:
## Summary
- Implements the Technology tab in the building/unit/research "?" info popup (#152)
- Shows all game objects grouped by Construction, Research, Ships, Defense, and Rockets, with accordion expand/collapse
- Renders sprite icons and direct prerequisites with fulfilled/unfulfilled status based on the current player's levels

## Test plan
- [ ] Open any building/unit/research "?" popup and switch to the **Technology** tab
- [ ] Confirm all five category sections are listed and **collapsed by default**
- [ ] Confirm clicking a category heading toggles its list
- [ ] Confirm technology links open Techinfo and prerequisite links open Techtree
- [ ] Confirm prerequisite rows show green (fulfilled) / red (unfulfilled) based on current levels
- [ ] Run `php artisan test --filter=TechtreeTest`

PR #1536, opened 2026-08-04
Implement technology tab in object popup

### LINKED ISSUE #152: Implement "?" popup technology tab
Implement the technology tab which appears in the popup when you click on the "?" button for buildings/units/research objects.  It should look like this:

Screenshot 2024-05-14 at 08 14 09

Kevinlucs:
## Summary

Implements the Technology tab displayed in the object information popup.

The tab now provides a categorized overview of buildings, research, ships, and defenses, including their prerequisite status and required levels.

## Changes

- Add categorized technology overview for:
  - Construction
  - Research
  - Shipyard
  - Defense
- Display object images and names
- Display fulfilled and unfulfilled prerequisites
- Show current and required levels for unmet prerequisites
- Add accordion behavior for technology categories
- Add translations for all currently supported languages
- Add feature tests for popup availability and prerequisite status

## Testing

- `php artisan test --filter TechtreeTest`
  - 7 tests passed
  - 305 assertions
- `composer run cs -- --test`
  - Passed
- `composer run stan`
  - Passed
- Manually tested the popup and category accordion

## Note

`npm run build` is currently blocked by pre-existing legacy CSS syntax in the main branch, including invalid single-line CSS comments and IE-specific selectors. This pull request does not modify CSS files.

Closes #152

### DIFF
--- a/resources/lang/en/t_ingame.php
+++ b/resources/lang/en/t_ingame.php
@@ -930,7 +930,6 @@ return [
         'is_requirement_for'                    => 'is a requirement for',
         'level'                                 => 'Level',
 
-        // Categories
         'category_construction' => 'Construction',
         'category_research' => 'Research',
         'category_shipyard' => 'Shipyard',
--- a/resources/lang/en/t_ingame.php
+++ b/resources/lang/en/t_ingame.php
@@ -930,6 +930,12 @@ return [
         'is_requirement_for'                    => 'is a requirement for',
         'level'                                 => 'Level',
 
+        // Categories
+        'category_construction' => 'Construction',
+        'category_research' => 'Research',
+        'category_shipyard' => 'Shipyard',
+        'category_defense' => 'Defense',
+
         // Shared table columns
         'col_level'                             => 'Level',
         'col_difference'                        => 'Difference',
--- a/app/Http/Controllers/TechtreeController.php
+++ b/app/Http/Controllers/TechtreeController.php
@@ -80,6 +80,9 @@ class TechtreeController extends OGameController
         } elseif ($tab === 3) {
             return view('ingame.techtree.technology')->with([
                 'object' => $object,
+                'technology_categories' => $this->getTechnologyCategories(
+                    $player->planets->current()
+                ),
             ]);
         } elseif ($tab === 4) {
             return view('ingame.techtree.applications')->with([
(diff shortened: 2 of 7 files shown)


resources/lang/en/t_ingame.php
```diff
@@ -930,6 +930,12 @@
         'is_requirement_for'                    => 'is a requirement for',
         'level'                                 => 'Level',
 
+        // Categories
```
jackbayliss: We can drop these comments, it's clear its categories from the prefix :)

Kevinlucs: Should I remove it and resend it ?

jackbayliss: You can just remove the comments and commit, it should alter here

Kevinlucs (commented):
removing comments from the technology tab

tiweb442:
i had one already made https://github.com/lanedirt/OGameX/pull/1502 you can review my code if you estimate yours to be better let me know i close my pr

tiweb442:
Looking closer, ours is closer to classic: separate Rockets category, tech name opens Techinfo (tab 2), prereqs open Techtree. Happy to keep #1502 open unless maintainers prefer consolidating — open to cherry-picking a11y bits from yours into ours.

Geda173 (commented):
This looks good, nice work.

I like that you reused the original class names. The existing stylesheet already targets that exact structure, so it renders correctly without any new CSS.

Two small things, neither blocking:

1. The popup title now drops the object name (`data-title` is just "Technology tree" instead of "Technology tree - Metal Mine"). This is the only one of the four tabs that does, so the header changes as you switch between them. Was that deliberate? I can see the argument since this tab lists everything, but I'd rather it be a conscious choice.

2. `data-current-level`, `data-required-level` and `data-requirement-met` aren't used by any CSS or JS, they only exist for the test, and `data-requirement-met` says the same thing as the `fulfilled`/`unfulfilled` class. Fine to leave, just noting it.

Could you rebase on `main`? The branch is around 33 commits behind. It merges cleanly, so this is just housekeeping. I'll squash the translation commits on merge.

Happy to approve once it's rebased.

tiweb442:
Thanks for the review — useful notes on `data-title` and the test-only `data-*` attrs.

Quick cross-check with #1502 (already approved): classic OGame has a separate **Rockets** category, tech names open **Techinfo (tab 2)**, and prereqs open **Techtree (tab 1)**. #1536 currently uses 4 categories (missiles under Defense), sends both links to tab 1, and drops the object name from `data-title`.

Happy to keep #1502 open and cherry-pick the a11y accordion bits from here if maintainers prefer classic fidelity — or close #1502 if you’d rather land this one after rebase and fold Rockets/nav into a follow-up. Which direction do you want?

PR #1541, opened 2026-08-09
Fix recycle mission auto-selection on position 16

### LINKED ISSUE #1514: [BUG] Recycle mission on pos 16 should be auto selected
**Describe the bug**
Currently and with #1497 merged, the game does not auto-select the recycle debris mission when it should. It currently defaults to no mission selected at all, even though this is the only possible one.

Kevinlucs:
## Summary

Automatically selects the Recycle mission when it is the only available mission for an expedition debris field on position 16.

## Changes

- Add automatic Recycle mission selection when it is the only available mission
- Keep the existing Expedition auto-selection behavior unchanged
- Avoid changing mission selection when multiple missions are available

## Testing

- `git diff --check`
- `node --check resources/js/ingame/e7c74974620fa35b197315ebdbb8c2.js`
- Manually tested with:
  - Pathfinder selected
  - Expedition debris field at position 16
  - Recycle confirmed as the only available mission
  - Recycle automatically selected on Fleet Dispatch II

## Note

`npm run build` was not executed due to the pre-existing legacy CSS syntax issues in the main branch.

Closes #1514

### DIFF
--- a/resources/js/ingame/e7c74974620fa35b197315ebdbb8c2.js
+++ b/resources/js/ingame/e7c74974620fa35b197315ebdbb8c2.js
@@ -33937,7 +33937,7 @@ FleetDispatcher.prototype.validateMissions = function () {
 };
 
 FleetDispatcher.prototype.refreshMissions = function () {
-  $('#missions>li>a.selected').removeClass('selected'); //select expedition if no mission is selected and if it is the only one available
+  $('#missions>li>a.selected').removeClass('selected');
 
   if (this.isOnlyMissionAvailable(this.fleetHelper.MISSION_EXPEDITION)) {
     if (this.hasMission() === false) {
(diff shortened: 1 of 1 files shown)


Geda173 (changes requested):
Thanks for the fix. The logic is right, and it matches how the expedition auto-select already works.

One thing needs doing before this can be merged: the built bundle wasn't regenerated.

The change is only in the source file. The game actually loads `public/build/assets/ingame-9fae412d.js`, and that file doesn't contain the fix, so merging this as-is wouldn't change anything for players. Nothing rebuilds assets on deploy, the committed bundle is what gets served.

Worth knowing this also means the preview environment doesn't show the fix. If you tested there and saw no change, that's why. It'll only work under `npm run dev`, which serves the source files directly.

Running `npm run build` and committing the result should sort it out. That'll rename the bundle and update `manifest.json` too.

Rebase + rebuild and I'll take another look.

Issue #742, opened 2025-09-21
Add player honor system


lanedirt:
Related to:
- #703 

Where it shows in the galaxy whether a player is a honourable target. The honor system consists of multiple game mechanics:
- A battle against a honorouble vs non-honourable target results in honor point gain or loss
- The amount of honor points (both positive and negative) will determine how much loot a player can gain from a successful battle against another player
- Honor points gives the user a custom status in the galaxy overview

More information: https://ogame.fandom.com/wiki/Honor_Points

PR #1551, opened 2026-08-11
Implement battle engine tactical retreat (#294)

### LINKED ISSUE #294: Implement battle engine tactical retreat logic
The "tactical retreat" functionality is not implemented yet in the battle engine. The value currently displayed in the battle reports is static. This feature will need to be added.

tiweb442:
## Summary

* Implements OGame tactical retreat in the shared battle engine: supremacy ratio, flee gates, deuterium cost, and optional attacker return-on-retreat
* Persists defender preference (`users.tactical_retreat_ratio`) and attack option (`fleet_missions.retreat_after_defender_retreat`), and wires the fleet UI
* Replaces the static battle-report `1:100` / “did not flee” with real ratio and flee status
* Aligned to authentic OGame behavior (Gameforge / simulators): Deathstars count for points but cannot flee; flee deuterium uses neighbouring **planet slot**; threshold uses `>=` (e.g. exact 5:1 flees)

Closes #294

### Behavior notes

* **Moon destruction** uses the same `simulateBattle()`, so moon-stationed fleets currently flee. Flee fuel is still 1.5× a neighbouring **planet slot** computed from the moon’s coordinates. This matches the shared engine; there is no separate moon-destruction exception. Covered by `testMoonStationedFleetFleesViaSharedBattleEngine`. Moon destruction battle reports now persist the same `tactical_retreat` payload as attacks.
* **`ogamex:test:battle-engine-performance`** sets the fixture player’s retreat preference to Never so defending ships cannot flee mid-benchmark and invalidate timing.

### Follow-ups (out of scope here)

* Honourable-combat / bandit no-flee gates → #1560 (blocked on honour system #742)
* Stale battle-report `data-raw-*` mock client metadata → #1561 (visible PHP ratio/flee text is already correct)

## Test plan

* [x] `php artisan test --filter='TacticalRetreat|PhpBattleEngineTest|HamillManoeuvreTest|FleetDispatchAttackTest::testDispatchFleetCombatReport'` (61 passed; Pint + PHPStan clean)
* [x] Preview https://pr-1551.preview.ogamex.dev @ `35f63e4`: preference persist; flee at 5:1; Never = no flee; inactive no flee; mutual retreat (0 rounds / 0 loot); defenses-only combat
* [ ] Manual UI glance: Fleet Never / 5:1 radios, “Return upon retreat by defenders” checkbox, Admiral 3:1 when available
* Attack a much weaker fleet and confirm the report shows a real `1:N` ratio, flee text, and the deuterium cost when the defender flees
* Confirm fleed ships remain on the defender planet while defenses can still fight
* Toggle Never / 5:1 on the fleet page and confirm preference persists
* Send an attack with “Return upon retreat by defenders” and confirm both sides withdraw without combat rounds

### DIFF
--- a/app/Console/Commands/Test/TestBattleEnginePerformance.php
+++ b/app/Console/Commands/Test/TestBattleEnginePerformance.php
@@ -88,6 +88,12 @@ class TestBattleEnginePerformance extends TestCommand
         $this->playerService->setResearchLevel('shielding_technology', 10);
         $this->playerService->setResearchLevel('armor_technology', 10);
 
+        // simulateBattle() evaluates tactical retreat and may deduct deuterium / strip
+        // defending ships. Disable flee so the benchmark measures combat, not retreat.
+        $benchmarkUser = $this->playerService->getUser();
+        $benchmarkUser->tactical_retreat_ratio = 0;
+        $benchmarkUser->save();
+
         // Set up defender planet with provided units
         foreach ($fleets['defender']->units as $unit) {
             $this->currentPlanetService->addUnit($unit->unitObject->machine_name, $unit->amount);
--- a/app/GameMessages/BattleReport.php
+++ b/app/GameMessages/BattleReport.php
@@ -305,6 +305,8 @@ class BattleReport extends GameMessage
         $moonChance = 0;
         $moonCreated = false;
         $hamillManoeuvreTriggered = false;
+        $tacticalRetreatRatio = 1;
+        $tacticalRetreatDefenderFled = false;
 
         if (isset($battleReportModel->general['moon_existed'])) {
             $moonExisted = $battleReportModel->general['moon_existed'];
@@ -317,6 +319,12 @@ class BattleReport extends GameMessage
         }
         if (isset($battleReportModel->general['hamill_manoeuvre_triggered'])) {
             $hamillManoeuvreTriggered = $battleReportModel->general['hamill_manoeuvre_triggered'];
+        }
+        if (isset($battleReportModel->general['tactical_retreat']['ratio'])) {
+            $tacticalRetreatRatio = (int)$battleReportModel->general['tactical_retreat']['ratio'];
+        }
+        if (isset($battleReportModel->general['tactical_retreat']['defender_fled'])) {
+            $tacticalRetreatDefenderFled = (bool)$battleReportModel->general['tactical_retreat']['defender_fled'];
         }
 
         // Load attacker player
@@ -444,6 +452,8 @@ class BattleReport extends GameMessage
             'moon_chance' => $moonChance,
             'moon_created' => $moonCreated,
             'hamill_manoeuvre_triggered' => $hamillManoeuvreTriggered,
+            'tactical_retreat_ratio' => $tacticalRetreatRatio,
+            'tactical_retreat_defender_fled' => $tacticalRetreatDefenderFled,
             'attacker_weapons' => $attacker_weapons,
             'attacker_shields' => $attacker_shields,
             'attacker_armor' => $attacker_armor,
--- a/app/GameMessages/BattleReport.php
+++ b/app/GameMessages/BattleReport.php
@@ -307,6 +307,7 @@ class BattleReport extends GameMessage
         $hamillManoeuvreTriggered = false;
         $tacticalRetreatRatio = 1;
         $tacticalRetreatDefenderFled = false;
+        $tacticalRetreatDeuteriumCost = 0;
 
         if (isset($battleReportModel->general['moon_existed'])) {
             $moonExisted = $battleReportModel->general['moon_existed'];
@@ -325,6 +326,9 @@ class BattleReport extends GameMessage
         }
         if (isset($battleReportModel->general['tactical_retreat']['defender_fled'])) {
             $tacticalRetreatDefenderFled = (bool)$battleReportModel->general['tactical_retreat']['defender_fled'];
+        }
+        if (isset($battleReportModel->general['tactical_retreat']['deuterium_cost'])) {
+            $tacticalRetreatDeuteriumCost = (int)$battleReportModel->general['tactical_retreat']['deuterium_cost'];
         }
 
         // Load attacker player
@@ -454,6 +458,8 @@ class BattleReport extends GameMessage
             'hamill_manoeuvre_triggered' => $hamillManoeuvreTriggered,
             'tactical_retreat_ratio' => $tacticalRetreatRatio,
             'tactical_retreat_defender_fled' => $tacticalRetreatDefenderFled,
+            'tactical_retreat_deuterium_cost' => $tacticalRetreatDeuteriumCost,
+            'tactical_retreat_deuterium_cost_formatted' => AppUtil::formatNumberLong($tacticalRetreatDeuteriumCost),
             'attacker_weapons' => $attacker_weapons,
             'attacker_shields' => $attacker_shields,
             'attacker_armor' => $attacker_armor,
--- a/app/GameMissions/Abstracts/GameMission.php
+++ b/app/GameMissions/Abstracts/GameMission.php
@@ -269,7 +269,7 @@ abstract class GameMission
      * @return FleetMission The created fleet mission.
      * @throws Exception
      */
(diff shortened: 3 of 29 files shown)


tiweb442:
## Update — review / fidelity / preview

Follow-up work on this PR after open:

### Fixes
* Mutual retreat no longer awards loot when only ships fled (empty combat side was treated as attacker win)
* Fleed ships no longer padded into combat rounds as destroyed (`sanitizeRoundArray`)
* Explicit NPC flee skip (no longer relying on `isInactive()` alone)

### OGame 1:1 alignment
* Deathstars: count 100% for retreat points, still cannot flee
* Flee deuterium: neighbouring **planet slot** (not system)
* Exact 5:1 encoded as `>=`, with unit tests at / just below the threshold

### Verification
* Filtered suite: **40 passed**; Pint PASS; PHPStan OK
* Preview redeployed to `35f63e4` — functional smoke tests passed (preference, flee, Never, inactive, mutual retreat, defenses-only)

### Deferred (tracked)
* #1560 — honourable / bandit retreat gates (needs #742)
* #1561 — hardcoded `data-raw-*` mock metadata in battle-report templates

Geda173 (changes requested):
Reviewed this locally. This is solid work, and I want to say that clearly before the nitpicks: the rules are correct, the architecture is right, and the tests are real ones rather than decoration.

The game rules match the official sources. 5:1, 3:1 with Admiral, civil ships at 25%, defenses/probes/satellites excluded, Deathstars unable to flee, and deuterium at 1.5x the fuel to the neighbouring position. All correct, including the 1.5x multiplier.

Four things I would like addressed.

**1. The NPC and inactive gates have no test at all.** Replacing either with `if (false)` leaves all 15 unit tests green and all 200 expedition/espionage tests green. The NPC gate especially, since expedition battles reuse this engine and your own commit message says it was a fix. Note that `TacticalRetreatDecision::$blockedReason` is set in eight places and read nowhere in the codebase, so asserting it in the tests would cover both gates cheaply and give that field a purpose.

**2. The tests only exercise `PhpBattleEngine`, but `SettingsService::battleEngine()`
defaults to `rust`.** We already have `BattleEngineTestAbstract` with `PhpBattleEngineTest` / `RustBattleEngineTest` subclasses for exactly this reason. Could the retreat tests move onto that pattern? The Rust path works today, but nothing currently protects it.

**3. `FleetDispatchTacticalRetreatTest::testTacticalRetreatPersistsRatioAndFleeInBattleReport` is order-dependent.** It failed for me with "Failed asserting that 10 matches expected 5" on a database carrying earlier test state, then passed in isolation and on later runs. The manual cleanup loop only clears ships and assumes the planet is otherwise consistent. Given how much trouble the shared foreign planet has caused in CI before, this needs proper isolation rather than a targeted cleanup.

**4. Scope.** `MoonDestructionMission` uses the same `simulateBattle()`, so moon
fleets now flee as well, with the destination computed as a neighbouring planet slot from the moon's coordinates. That is not mentioned in the description and has no test. Relatedly, `simulateBattle()` now writes to the database via `deductResources()`, which means `ogamex:test:battle-engine-performance` will have its defending fleet flee mid-benchmark.

Smaller points:

- The flee deuterium cost is stored in the report data but never displayed, so the defender loses deuterium with nothing in the report explaining it.
- `calculateSupremacyRatio` returns raw attacker points when the defender has no fleet points, so a report can read "1:400" where 400 is a point total, not a ratio. It is tested, so I assume it is intentional, but it looks wrong in the UI.
- The slot test hardcodes `15` rather than `UniverseConstants::MAX_PLANET_POSITION`.
- The comment in `testFleeDoesNotTriggerJustBelowFiveToOne` describes numbers the test does not use.

Two questions for @lanedirt rather than for you:

- The migration defaults `tactical_retreat_ratio` to `5`, so this turns on retroactively for every existing player. That matches OGame, but it does change live raid outcomes (three `ReaperDebrisCollectionTest` cases had to opt out). Worth a deliberate decision.
- The threshold uses `>=`, so an exact 5:1 flees. The wiki wording is "more than five times", which implies `>`. Ambiguous, so worth a ruling either way.

tiweb442:
Thanks for the thorough review — addressed the required items in 2dc2dc88.

**1. NPC and inactive gates.** `testNpcDefenderBlockedReasonPreventsFlee` and `testInactiveDefenderBlockedReasonPreventsFlee` call `TacticalRetreatService::evaluate()` and assert `$blockedReason` (`npc` / `inactive`), then also run `simulateBattle()` so replacing either gate with `if (false)` fails. Expedition NPC battles reuse this path via `NPCPlanetService`.

**2. PHP vs Rust.** Retreat tests now live on `TacticalRetreatTestAbstract` with `PhpTacticalRetreatTest` / `RustTacticalRetreatTest` subclasses (same pattern as `BattleEngineTestAbstract`). They stay on `AccountTestCase` rather than `BattleEngineTestAbstract` itself because that abstract uses `player_id = 0`, which the NPC gate already treats as non-fleeing — so real flee coverage needs a real player.

**3. Order-dependent fleet-dispatch test.** `testTacticalRetreatPersistsRatioAndFleeInBattleReport` now uses `getNearbyForeignCleanPlanet()` (fresh hostile planet) and looks up the battle report by that planet’s coords/owner instead of “latest report + ship-only cleanup” on the shared foreign planet.

**4. Scope.**
- Moon fleets currently **do** flee: `MoonDestructionMission` shares `simulateBattle()`, and flee fuel still uses a neighbouring planet slot from the moon coords. Documented with `testMoonStationedFleetFleesViaSharedBattleEngine` (current behavior, not a rules change). MD battle reports now persist the same `tactical_retreat` payload as attacks. If moon destruction should *not* flee, that is a product call — left as-is pending authentic OGame evidence.
- `ogamex:test:battle-engine-performance` now sets the fixture player’s preference to Never so defending ships cannot flee mid-benchmark.

**Smaller points.**
- Flee deuterium cost is shown on the full battle report when the defender fled.
- `calculateSupremacyRatio` still returns raw attacker points when the defender has 0 fleet points (`1:400` style). Existing `testSupremacyRatioRounding` locks that in, and I do not have a clear authentic source for `1:-` / infinite, so left unchanged.
- Slot test now uses `UniverseConstants::MAX_PLANET_POSITION`.
- Comment on `testFleeDoesNotTriggerJustBelowFiveToOne` now matches the 19 LF vs 4 LF numbers.

**Still for @lanedirt (not decided here):**
- Migration default `tactical_retreat_ratio = 5` is retroactive for existing players (matches OGame, changes live raids).
- Threshold is `>=` (exact 5:1 flees); wiki “more than five times” could be read as `>`.

Filtered suite: 61 passed (`TacticalRetreat|PhpBattleEngineTest|HamillManoeuvreTest|FleetDispatchAttackTest::testDispatchFleetCombatReport`). Pint + PHPStan clean.

tiweb442:
Review items from 2026-08-17 look addressed on `2dc2dc88` (NPC/inactive `blockedReason` tests, PHP+Rust abstract, isolated feature test, moon-destruction coverage, benchmark opt-out, flee deuterium in the report UI, `UniverseConstants::MAX_PLANET_POSITION`, comment/numbers aligned).

Still waiting on @lanedirt for the two product questions (migration default of `5`, and `>=` vs wiki “more than five times”). Please re-review when you have a moment — GitHub still shows Changes Requested from the earlier pass.

Geda173 (approved):
Awesome, thanks for the changes. Now we'll have to wait to get those two questions resolved.

lanedirt:
> Two questions for @lanedirt rather than for you:
> - The migration defaults tactical_retreat_ratio to 5, so this turns on retroactively for every existing player. That matches OGame, but it does change live raid outcomes (three ReaperDebrisCollectionTest cases had to opt out). Worth a deliberate decision.
> - The threshold uses >=, so an exact 5:1 flees. The wiki wording is "more than five times", which implies >. Ambiguous, so worth a ruling either way.

Hi @Geda173, both sound good as-is to me. The "more than five times" is indeed a bit ambigious, but keeping it at `>=` seems fine by me as it can be read as "5 times or more" just as well (maybe a translation wording issue)?

So +1 from me. Feel free to merge when ready.

Geda173:
Great, that's a wrap then. Thanks to both of you for your contributions. Merging.

Issue #96, opened 2024-04-22
Add fleet mission deployment logic


lanedirt:

lanedirt:
Added in PR #105

PR #1573, opened 2026-08-17
refactor(i18n): move hardcoded English strings in Galaxy and blades to translation keys

piciolo:
Split out of #1385 as requested in [this comment](https://github.com/lanedirt/OGameX/pull/1385#issuecomment-5312978273) — one concern per PR, and this one is i18n only.

### What this does

Replaces the remaining hardcoded English strings and inline `@lang('English text')` calls in `GalaxyController` and a set of blades with translation keys, and adds the **257 missing keys** to `resources/lang/en`.

Pages covered: galaxy, the resources / facilities / research / shipyard / defense detail panels, fleet dispatch, jump gate dialog, notes, scrap merchant, server settings (both the admin page and the in-game overlay), developer shortcuts, and the shared build queue partial.

### How the keys were derived

Each new key was given the exact literal it replaces. Where a string carried interpolated values, it became a `:placeholder` parameter (`Cancel expansion of :name to level :level?`, `Costs: :amount DM`, and so on). Where an equivalent key already existed it was reused rather than duplicated — for example the scrap merchant reuses `t_merchant.error.scrap.storage_insufficient` and `t_merchant.unknown_item`.

**Two strings do change wording**, both in `GalaxyController`, because reusing an existing key was the right call there:

| Was | Now | Why |
|---|---|---|
| `Moon destruction` | `Moon Destruction` | Capitalisation only; `t_ingame.fleet.mission_destroy_moon` already labels the same mission this way on the fleet page. |
| `Not enough deuterium to use phalanx` | `Not enough deuterium to deploy phalanx.` | `t_ingame.galaxy.phalanx_no_deut` already carried this wording for the same message in `galaxy/index.blade.php`, and it matches OGame 13.0.0-r12. |

Two other spots looked like candidates for the same treatment but are **not** consolidated, because the galaxy view deliberately labels them differently from the fleet page: the debris menu keeps `Harvest` (its own `t_ingame.galaxy.mission_harvest`, not `mission_recycle` = "Recycle Debris Field") and the relocation mission entry keeps `Relocate` (its own `t_ingame.galaxy.relocate_action`, not `relocate_title` = "Resettle Planet", which is still the confirmation dialog title).

### Verification

Two checks, both repeatable:

1. **Key audit** — walks every key the refactor introduced, reused and new alike, and compares the literal it replaced against the value it now resolves to. It reports exactly the two consolidations in the table above and nothing else.
2. **Render diff** — every affected page and AJAX endpoint rendered on `main` and on this branch with the locale forced to English, then diffed. This includes `POST /ajax/galaxy`, where the mission names and phalanx strings actually live.

Apart from the two table entries, what is left is only equivalent escaping:

- `"Abandon\/Rename …"` → `"Abandon/Rename …"` — same JS string.
- `@json(__('…'))` emits a double-quoted JSON string where the old code had a single-quoted JS literal.
- Inside `title=` / `data-tooltip-title=` attributes, `<br />` is now written as `&lt;br /&gt;`, which the HTML parser decodes back to the same attribute value.

The one place where that last point would *not* have been equivalent is the tactical retreat tooltip, which contains `&amp;#96;` — `{{ }}` would have escaped it one level too far, so that one uses `{!! !!}`, matching how `hire_admiral_tooltip` is already rendered in `main.blade.php`.

Locally against the project's Docker image (PHP 8.5): `pint --test` clean, `phpstan analyse` at level 8 reports **No errors**, and `Http200Test`, `GalaxyTest`, `JumpGateTest`, `MerchantTest` and `AdminTest` pass — **70 tests, 975 assertions**.

### Deliberately not included

The original branch also touched `fleetevents/eventrow.blade.php` and `FleetEventRowViewModel`, which is a behaviour change (showing hostile/friendly fleet labels instead of always "Own fleet"), and added a masquerade form to the developer shortcuts page. Neither is i18n, so both are left out; happy to bring the fleet-event one up separately if it is wanted.


Geda173 (changes requested):
Thanks for splitting this out of #1385.

1. `rank_founder_default` and `rank_newcomer_default` in `t_ingame.php` include the Blade `}}` closing tag and surrounding quotes in their values (`'\'Founder\' }}'`). This renders `'Founder' }}` in the input field instead of `Founder`. Change the values to plain `'Founder'` and `'Newcomer'`.

2. Four strings were changed, not just extracted. The PR description says "nothing was reworded or invented" but:
- `'Harvest'` became `'Recycle Debris Field'`
- `'Relocate'` became `'Resettle Planet'`
- `'Not enough deuterium to use phalanx'` became `'Not enough deuterium to deploy phalanx.'`
- `'Moon destruction'` became `'Moon Destruction'`

These changes need to be reverted unless you know they are warranted. Harvest and Relocate I am pretty sure should stay the way they were prior to this PR. 13.0.0-r12 of OGame uses the string `'Not enough deuterium to deploy phalanx.'`, so that may be okay to stay.  The capitalization of Destruction in Moon Destruction seems to also be okay. However, your claim in the PR body goes directly against these changes.

piciolo:
Both correct, thanks — and the second one is a fair hit on the PR description, which overstated what I had verified.

**1. `rank_founder_default` / `rank_newcomer_default`**

Fixed — they are plain `'Founder'` and `'Newcomer'` now.

The cause is worth stating because it explains the shape of the bug. I extracted the values mechanically by aligning each removed line of the diff against the line that replaced it. That works when the whole expression is replaced, but here only part of it was:

```blade
value="{{ $alliance->founder_rank_name ?? 'Founder' }}"
value="{{ $alliance->founder_rank_name ?? __('t_ingame.alliance.rank_founder_default') }}"
```

The extractor matched the trailing ` }}` as part of the replacement rather than as shared context, so the captured value ran on past the string literal. I have since swept every added lang value for that signature; these two were the only ones affected. The same pattern appears in `fleet/index.blade.php` for `t_ingame.fleet.planet` / `.moon`, but those keys already existed with the right values, so nothing was written.

**2. The four changed strings**

You are right that these were not extractions, and the PR body should not have claimed otherwise. What happened is that in `GalaxyController` these four spots were consolidated onto keys that already existed for other screens, and I only checked that *newly added* keys matched their originals — not the reused ones. That is the gap that let it through, and the render diff I ran did not catch it either, because I only fetched `GET /galaxy` and never the `POST /ajax/galaxy` payload where these strings actually live.

Reverted:

- **Harvest** — `t_ingame.fleet.mission_recycle` is `'Recycle Debris Field'` and is used on the fleet page, so I did not want to change its value. The galaxy debris menu now uses its own `t_ingame.galaxy.mission_harvest` => `'Harvest'`.
- **Relocate** — same reasoning: `t_ingame.galaxy.relocate_title` is `'Resettle Planet'` and is still used as the confirmation dialog title in `galaxy/index.blade.php`. The mission entry now uses `t_ingame.galaxy.relocate_action` => `'Relocate'`.

Kept, on your reading:

- **`'Not enough deuterium to deploy phalanx.'`** — matches 13.0.0-r12, and the key was already worded that way for the same message in `galaxy/index.blade.php`.
- **`'Moon Destruction'`** — capitalisation only, and it matches how the mission is already labelled on the fleet page.

I have written both of these into the PR description as deliberate consolidations rather than leaving the blanket "nothing was reworded" claim.

**What I added to stop this recurring**

- An audit that walks every key the refactor introduced — reused *and* new — and compares the literal it replaced against the value it now resolves to. It reports exactly the two intentional consolidations above and nothing else.
- The render comparison now covers `POST /ajax/galaxy` as well as the GET pages. Re-running it across `main` and this branch with the locale forced to English, the galaxy payloads are byte-identical apart from `Moon destruction` → `Moon Destruction`; `Harvest` appears 4 times and `Relocate` 26 times in both, and `Recycle Debris Field` / `Resettle Planet` appear in neither.

Pint clean, PHPStan level 8 reports no errors, and the feature tests still pass.

Geda173 (approved):
Thanks for the changes, approved.

Optional, non-blocking: GalaxyTest asserts structure on `ajax/galaxy` but never the mission labels, which is why the rewording went unnoticed originally. A one-line `assertSeeText('Harvest')` in the existing debris test would pin the exact string that regressed. Fine to leave for a follow-up.

nagi1:
Hello @piciolo
Awesome work, please resolve the conflict so maintainers can merge this PR

Thanks for your contributions

jackbayliss (approved):

piciolo:
Rebased on current `main` to clear the conflict. One file conflicted, and it is worth a second pair of eyes because `main` changed underneath it.

`fleet/index.blade.php` — `main` has since turned the tactical retreat block into a real feature: a `fleet.tacticalretreat` route with `@csrf`, a persisted `$tacticalRetreatRatio` driving the `checked` state, and the 3:1 option gated behind `@if ($hasAdmiral ?? false)`. This branch only knew the old markup, where the form posted to `#TODO_tacticalRetreat`.

I kept `main`'s version in full and applied only the four keys this branch had already introduced for that block: the long tooltip (`{!! !!}`, since it contains `&amp;#96;`), the "Tactical retreat:" label, "Never", and the disabled 3:1 tooltip. The route, the CSRF token, the ratio state and the Admiral gate are `main`'s, untouched.

Nothing else changed: still 18 files, +638 −379, and the audit that walks every key the refactor introduces and compares it against the literal it replaced still reports only the two documented consolidations.

Verified after the rebase: `pint --test` clean, and `Http200Test`, `GalaxyTest`, `JumpGateTest`, `MerchantTest` and `AdminTest` pass — 70 tests, 975 assertions.

Unrelated to this PR, but I noticed it while resolving and it seems worth passing on: `TacticalRetreatService::resolveRetreatThreshold()` and `FleetController` both gate the 3:1 ratio on `PlayerService::hasAdmiral()`, and on `main` that method is still `// TODO: add logic; return false;`. So the escape-rate half of the feature is built and wired but can never trigger today. That is exactly the method #1385 implements — which also means my earlier claim there, that the Admiral's improved escape rate had nothing to hook into, is now out of date.

Geda173:
Thanks, let's get this finally merged.

PR #1578, opened 2026-08-19
Fix login pages confirmed issues

### LINKED ISSUE #1574: [UI/UX] Login page: minor layout duplication, empty sections, tab video playback, and error display issues
### Summary
On the OGameX login page (`/login`), navigating via the **About OGame** or **Media** tabs triggers a minor double layout rendering issue. Also, youtube video is played along when tab switches. Additionally, most of links are currently empty. Form buttons Register and Login are at different places and error messages are rendered in two places.

---

### Key Layout Issues

1. **Layout Duplication:** 
   - Clicking on **About OGame** or **Media** causes a minor double layout rendering on the page.
2. **Persistent Video Playback:** 
   - The embedded YouTube trailer plays continuously and keeps on playing even when switching to different tabs or sections.
3. **Empty Informational Sections:** 
   - The following pages/tabs currently contain no content:
     - **T&Cs** (Terms & Conditions)
     - **Privacy Policy**
     - **About OGame**
     - **Media**
     - **Social Media Links** 
     - **Wiki Links** 
4. **Duplicated Error Messages:** 
   - Form/validation errors are displayed simultaneously in two separate locations on the interface.
5. **Form Layout Note:** 
   - Registration correctly sits on the left side and login is on the right side.

---

### Proposed Improvements

* **App Store Badges:** Replace the legacy placeholder store icon with official, modern **Apple App Store** and **Google Play Store** badges and working download links.

baigny:
Fixes from #1574:

- Double layout render when switching to About OGame / Media tabs
- YouTube trailer keeps playing after switching away from Home tab
- Error messages showing in two places for the same submission (login vs register sharing the default error bag)

## Test plan
- [x] Switch About/Media tabs, confirm no duplicate header/footer render
- [x] Play trailer on Home, switch tabs, confirm playback stops; switch back, confirm it resumes
- [x] Submit invalid register form, confirm error shows only in register box, not under Home


Geda173 (changes requested):
Thanks @baigny,

Two requests before merge:

1. Rebuild the assets instead of hand-editing `public/build/assets/outgame-b0a7e11e.js`. Please run `npm run build` and commit the renamed asset plus the updated `manifest.json`.
2. Add a regression test for the error-bag change; nothing currently fails if `validateWithBag('register')` is reverted to `validate()`. For example: POST `/register` with a too-short password, then assert `$response->assertSessionHasErrorsIn('register', ['password'])` and that the default bag is empty, so the duplicate display cannot silently come back.

One observation, no change needed if intentional: About and Media now point at the same `#tabContentContainer` panel, so they display the Home content until real content exists (tracked in #1574).

baigny:
> Thanks @baigny,
> 
> Two requests before merge:
> 
> 1. Rebuild the assets instead of hand-editing `public/build/assets/outgame-b0a7e11e.js`. Please run `npm run build` and commit the renamed asset plus the updated `manifest.json`.
> 2. Add a regression test for the error-bag change; nothing currently fails if `validateWithBag('register')` is reverted to `validate()`. For example: POST `/register` with a too-short password, then assert `$response->assertSessionHasErrorsIn('register', ['password'])` and that the default bag is empty, so the duplicate display cannot silently come back.
> 
> One observation, no change needed if intentional: About and Media now point at the same `#tabContentContainer` panel, so they display the Home content until real content exists (tracked in #1574).

@Geda173  
fixes are completed, pls check
1 About Media doesn't have realistic content and is similar to Home content. So its better to disable if we are showing the tabs
2 Also, wiki link is broken can we keep just like that and content it's from server?
3 Through which account, should I test as a player ?

Geda173 (commented):
There are still merge conflicts that need resolving before this can proceed further.

baigny:
> There are still merge conflicts that need resolving before this can proceed further.

resolved pls check. @Geda173

Geda173 (approved):
Thanks for the fix. The failing test is unrelated to this branch and there is an open PR to resolve it. Approving.

lanedirt:
Thanks!

PR #1584, opened 2026-08-23
fix: prevent cache lock from hanging tests (frozen-time-safe lock wait)

nagi1:
## Description

Fixes a test hang that occurred in `PlanetServiceFactory::createInitialPlanetForPlayer()` (and therefore in any test that creates a user/planet). When a stale `planet_create_lock` was present, `Cache::lock('planet_create_lock', 10)->block(10)` never returned, causing the test runner to hang forever with no output.

**Root cause:** `Lock::block()` derives its timeout from `Carbon::now()`. Tests freeze time via `travelTo()` in `AccountTestCase::setUp()`, so `now()` never advances — the 10-second deadline can never arrive, and the loop spins in `usleep()` indefinitely. A leftover lock (e.g. from an abnormally terminated run) triggered this path.

**Changes:**
- `app/Factories/PlanetServiceFactory.php`: replaced `$lock->block(10)` with an equivalent wait that uses a real wall-clock deadline (`microtime(true)`) instead of `Carbon::now()`. Behavior is unchanged (waits up to 10 seconds, then throws `RuntimeException`), but it can no longer hang under frozen test time.
- `tests/TestCase.php`: `setUp()` now clears stale `cache_locks` before every test, so leftover locks from interrupted runs self-heal instead of blocking or failing subsequent tests.

### Type of Change:
- [x] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update

## Related Issues
N/A — no related issue (fixes an intermittent test hang discovered while adopting Pest).

## Checklist

- [x] **Automated Refactoring:** Rector run, no outstanding issues.
- [x] **Code Standards:** Code adheres to PSR-12, verified with Laravel Pint.
- [x] **Static Analysis:** Code passes PHPStan.
- [x] **Testing:**
    - Relevant unit and feature tests pass locally, including `tests/Feature/FleetDispatch/FleetDispatchLargeResourcesTest.php` (previously the hang case).
    - Verified the fix by re-creating the stale `planet_create_lock` and confirming the test self-heals instead of hanging.
- [x] **CSS & JS Build:** N/A — no JS/CSS files changed.
- [x] **Documentation:** N/A — behavior is unchanged in production.

## Additional Information

- No production behavior change: the lock is still held for up to 10s and the same `RuntimeException` is thrown if it can't be acquired. The only difference is the deadline source (real clock vs. Carbon), which makes the timeout robust to frozen test time.
- The `cache_locks` cleanup uses `DELETE` (not `RefreshDatabase`) and only touches transient lock state.


nagi1:
@jackbayliss this is a quick one please approve and merge when you can.

nagi1:
@Geda173 If you interrupt the test mid processing it leaves locked rows in db that make next run hangs

nagi1:
@jackbayliss This PR is ready for review the flaky test is fixed in another Pr by @Geda173

tests/TestCase.php
```diff
@@ -25,6 +25,21 @@ public function createApplication(): Application
         return $app;
     }
 
+    /**
+     * Set up the test environment.
+     *
+     * @return void
+     */
+    protected function setUp(): void
```
jackbayliss, line 33: Can't we just use CACHE_STORE in the phpunit.xml to prevent this from ever happening? 

Maybe it would break other tests tho, so not sure.

nagi1: I did this in another pr and nothing hasn't broke

jackbayliss: Can you link the PR? If we do that we don't need this logic

nagi1: please review this pr
And consider merge it it will open up a new possibilities for faster test feedback. 

https://github.com/lanedirt/OGameX/pull/1585

nagi1:
@jackbayliss PR now is ready for merge

Geda173 (commented):
I think this one is superseded. #1585 merged `CACHE_STORE=array` into `phpunit.xml`, which is exactly what @jackbayliss suggested above. With it `Cache::lock()` resolves to `ArrayLock` instead of `DatabaseLock` (I checked both ways), so tests never write to `cache_locks` and a killed run cannot strand a lock. The hang this fixes is no longer reachable.

The production half does not look needed either. The hang requires frozen time *and* a database-backed lock. There is exactly one `Cache::lock()` call site in the app (`PlanetServiceFactory.php:368`), reached via `CreateNewUser` in `TestCommand::handle`, which runs before any `setTestNow()`. The two Artisan commands that do freeze time (`TestBattleEnginePerformance.php:83` and `TestRaceConditionUnitQueue.php:78`) never create a planet after the freeze. Main's CI also runs all four `ogamex:test:race-condition-*` commands green today without this fix.

Worth closing unless I am missing a case. If any part is kept, it should not be the `tests/TestCase.php` cleanup: an unqualified `DELETE FROM cache_locks` in every test setUp would release the running scheduler's `withoutOverlapping` mutexes, which live in that table under the database store.

jackbayliss:
Not sure if we need this, if it's the array driver it shouldn't cause collisions anymore? Feel free to close.

nagi1:
Hi, @jackbayliss and @Geda173 It's a small change, and I'd suggest keeping it.

If we ever switch back to the database cache driver, it can be changed easily in `.env.testing`. This also acts as a safeguard against this kind of issue, which can be very hard to debug. It took me quite a while to track this down because the test produced no output.

But it's your call.

Thanks!

jackbayliss:
@nagi1 i'd rather close it, I wouldnt expect we'd switch back as the array driver fixes all of it :D  

If we did, we can do it afterwards...  

Less to maintain imo

PR #1580, opened 2026-08-20
Fix flaky ReaperDebrisCollectionTest: stale PlayerService cache and test pollution

Geda173:
## Description
The flaky `ReaperDebrisCollectionTest` failures had two independent causes, both fixed here:

**1. Stale PlayerService cache (primary cause of the CI flakiness)**
The tests disable tactical retreat on the defender (`tactical_retreat_ratio = 0`) after dispatching the fleet, but battle processing reused a `PlayerService` cached during the dispatch's fleet event requests, carrying the pre-write ratio of 5. The defender then tactically retreated, no ships were destroyed, and no debris was created (`Failed asserting that 0 is greater than 0`). Whether the retreat actually triggered depended on the randomly selected foreign planet (deuterium stock, activity), which is why failures appeared mostly on freshly provisioned CI databases and vanished on retry.

Fix: `$this->reloadApplication()` after writing the ratio in all three test methods, so battle processing starts with cold factory caches. 

The underlying factory behavior (reloadCache not propagating to the player) is tracked separately in #1581.

**2. Test pollution via shared settings**
- `ReaperDebrisCollectionTest` never called `basicSetup()`, so `debris_field_from_ships` was inherited from whatever test ran before; now called in all three test methods.
- `DefenseRepairBattleEngineTest` sets debris/repair settings but had no `tearDown()`; now resets `debris_field_from_ships`, `debris_field_from_defense`, `debris_field_deuterium_on`, and `defense_repair_rate` to defaults.

### Type of Change:
- [X] Bug fix
- [ ] Feature enhancement
- [ ] Documentation update
- [ ] Other (please describe):

## Related Issues
Fixes #1579 

## Checklist
Before submitting this pull request, ensure all following requirements as outlined in [CONTRIBUTING.md](https://github.com/lanedirt/OGameX/blob/main/CONTRIBUTING.md) are met:

- [X] **Automated Refactoring:** Rector has been run and no outstanding issues remain.
- [X] **Code Standards:** Code adheres to PSR-12 coding standards. Verified with Laravel Pint.
- [X] **Static Analysis:** Code passes PHPStan static code analysis.
- [X] **Testing:**
    - Relevant unit and feature tests are included or updated.
    - Tests successfully run locally.
- [X] **CSS & JS Build:** CSS and JS assets are compiled using Vite (`npm run build`) if any changes are made to JS/CSS files.
- [X] **Documentation:** Documentation has been updated to reflect any changes made.

## Additional Information
[...]

Issue #1579, opened 2026-08-20
[BUG] ReaperDebrisCollectionTest fails depending on test execution order


Geda173:
**Describe the bug**
`ReaperDebrisCollectionTest` has two failing tests that depend on test execution order:

- `testReaperCollectsDebris`
- `testReaperCollectsDebrisWithNonGeneralClass`

Both fail with:
Battle should create debris from destroyed units
Failed asserting that 0 is greater than 0.

**Cause**
`DefenseRepairBattleEngineTest` sets `debris_field_from_ships` to 0 but has no `tearDown()` method to reset it. When it runs before `ReaperDebrisCollectionTest`, the 0% setting leaks through and no debris is generated regardless of how many defender ships are destroyed.

Issue #1021, opened 2026-01-04
Change tests to use factories with RefreshDatabase / DatabaseTransactions


jackbayliss:
Factories should be used re https://laravel.com/docs/12.x/eloquent-factories

This means, everything we need can be setup per test if needed - that way we won't have flaky tests - we should use it with either the RefreshDatabase trait or DatabaseTransactions so any DB changes made for the test are rolled back.

Geda173:
I've started looking into this because at this point testing features with a more elaborate setup (ACS) is simply impossible when having to run the test suite, as about 8 or so tests completely refresh and reseed the database. As I've already said in a PR, I would like to completely avoid RefreshDatabase, as it makes testing impossible. Instead, we should be relying on DatabaseTransactions. Implementing DatabaseTransactions in the test suite faces a key challenge: the codebase uses database locks (cache_locks table) in `PlanetServiceFactory` for planet creation. These locks are incompatible with transactional tests because locks aren't properly released when transactions roll back, causing lock wait timeouts.

What we could do is modify the locking mechanism to be test-aware:
1. Use a transaction-aware lock strategy (Redis/file-based locks for production, skip locks in tests)
2. Make PlanetServiceFactory detect test environment and bypass locks
3. Refactor to use in-memory locks instead of database locks

These three have varying degrees of invasiveness, with 2 probably being the least invasive, but even then I have encountered problems:

- Cache Configuration Issue: Laravel uses CACHE_STORE environment variable (not CACHE_DRIVER), and tests were configured with the wrong variable. Need to add CACHE_STORE=array to phpunit.xml.
-  Even after addressing cache locks and configuration, the settings table uses updateOrCreate() which acquires row locks. When multiple tests run or when locks aren't immediately released after transaction rollback, subsequent tests timeout waiting for these orphaned locks.

These orphaned locks cause loads of issues and need to be dealth with, otherweise the tests simply hang.

EDIT: I've identified a few paths forward to resolve these issues:

For the database locks in PlanetServiceFactory:
Add an environment check in `createInitialPlanetForPlayer()` to bypass locking when running tests. This would be a minimal change - just wrapping the lock acquisition in `!app()->environment('testing')`. Production behavior remains unchanged.

For the cache configuration:
Add `CACHE_STORE `to `phpunit.xml` (alongside the existing `CACHE_DRIVER` setting). This ensures tests use the array cache driver instead of database cache, preventing cache-related transaction issues.

[...]

Issue #1577, opened 2026-08-19
[Dev] Test suite is very slow; evaluate Pest 5 for faster development feedback


nagi1:
**Is your feature request related to a problem? Please describe.**

The test suite has grown large enough that running it locally is becoming expensive during development.

A full run currently takes around **12 minutes** on my machine:

```text
Tests:    1080 passed (11968 assertions)
Duration: 738.14s
```

This makes running the full suite frequently while working on a PR impractical and slows down the feedback loop.

**Describe the solution you'd like**

I think we should evaluate migrating the test runner to **Pest 5**, mainly to take advantage of its new **Test Impact Analysis (TIA)** support.

TIA keeps track of which tests depend on which parts of the application and, after the initial baseline run, can rerun only the tests affected by the current changes.

This could make the normal development workflow much faster without reducing the coverage of the full test suite in CI.

Since OGameX is already on PHP 8.5 and PHPUnit 13, the project should already meet Pest 5's main runtime requirements.

I would suggest:

1. Evaluate running the existing suite through Pest 5 without immediately rewriting all tests.
2. Benchmark the current full suite against Pest.
3. Test `--tia` during normal local development.
4. Keep a complete test-suite run in CI to ensure nothing is missed.
5. If the results are useful, gradually migrate tests to Pest syntax rather than doing a large rewrite.

**Describe alternatives you've considered**

We could optimize individual slow tests or increase parallelization with the existing PHPUnit/ParaTest setup.

Those are still worth doing, especially for improving the actual full-suite runtime, but they do not solve the common development case where a small change causes us to repeatedly run a large suite.

Pest 5/TIA could complement those optimizations rather than replace them.

**Additional context**

Current local result:

```text
Tests:    1080 passed (11968 assertions)
Duration: 738.14s
```

Screenshot attached.

The goal here is not necessarily to migrate the whole test suite to Pest syntax. The first step would be to benchmark whether Pest 5/TIA gives us a meaningful improvement in developer feedback time.

Locally:
<img width="888" height="124" alt="Image" src="https://github.com/user-attachments/assets/7a52abef-9939-4fad-b488-6233581174af" />

nagi1:
In `tests/Feature/FleetDispatch/FleetDispatchAcsDefendTest.php`

> @todo Refactor test architecture to support DatabaseTransactions/RefreshDatabase by removing dependency on reloadApplication() expecting persisted users.

I think we totally need to work on this because the tests are painfully slow!

[...]

PR #1585, opened 2026-08-23
test: add isolated account test base (DatabaseTransactions + factory user setup)

nagi1:
## Description

First step of the test suite isolation migration.

Introduces `Tests\IsolatedAccountTestCase`, a drop-in base class that makes tests order-independent and removes the real HTTP registration round-trip.

This PR also converts two pilot files to prove the pattern:

- `PlanetFieldRestrictionTest`
- `UnitQueueTest`

**Why:** the suite currently has no DB isolation. Every test writes permanently to a shared test DB and can rely on data left by previous tests. This is the root cause of the order-dependent flakiness tracked in #1579 and #1580.

See `docs/test-isolation-migration-plan.md` for the full migration plan.

### What `IsolatedAccountTestCase` changes

Assertions in the converted tests are unchanged.

- Wraps each test in a database transaction using `DatabaseTransactions`, so no rows leak between tests.
- Creates the user through the production `CreateNewUser` action and authenticates with `actingAs()` instead of doing `/register`, `/login`, and `/overview` HTTP requests.
- Starts the session directly so `csrf_token()` and the controllers' `hash_equals(session()->token(), _token)` checks continue to work.
- Neutralizes the "first user becomes admin" side effect, since each isolated transaction starts with `User::count() === 1`.
- Resets stateful singletons in `tearDown()`:
  - `SettingsService`
  - `PlayerServiceFactory`
  - `PlanetServiceFactory`

These resets are needed because their in-memory caches are not rolled back by `DatabaseTransactions`.

### Cache isolation

Also adds `CACHE_STORE=array` to `phpunit.xml`.

Laravel reads `CACHE_STORE`, not the existing `CACHE_DRIVER`, so tests were silently using the database cache and its `cache_locks` table.

Using the array cache keeps `Cache::lock()` in memory and isolated between tests.

This global config change was verified against the full suite.

### Type of Change

- [ ] Bug fix
- [x] Feature enhancement
- [ ] Documentation update
- [x] Other (test infrastructure)

## Related Issues

Progress toward #1021 for using factories and proper DB isolation.

Complements #1584 for the frozen-time-safe lock wait fix.

## Checklist

- [x] **Automated Refactoring:** Rector run on changed files, no outstanding issues.
- [x] **Code Standards:** Laravel Pint passes on all changed files.
- [x] **Static Analysis:** PHPStan passes on changed files with no errors.
- [x] **Testing:**
  - `PlanetFieldRestrictionTest` + `UnitQueueTest`: 21 tests passing.
  - Both pass in isolation and in different execution orders.
  - Full suite locally: **1132 passed, 1 failed**.
  - The single failure, `VacationModeTest > research queue pauses during vacation mode`, is the existing time-travel flake and is unrelated to this PR.
- [x] **CSS & JS Build:** N/A, no JS/CSS changes.
- [x] **Documentation:** Added `docs/test-isolation-migration-plan.md`.

## Additional Information

No production code changed. This PR only touches `phpunit.xml` and tests.

Assertions in the two converted tests are untouched. Only the setup and isolation mechanism changed.

The scope is intentionally small. Follow-up PRs can migrate the remaining test files one at a time using the same base class.

### DIFF
--- a/tests/IsolatedAccountTestCase.php
+++ b/tests/IsolatedAccountTestCase.php
@@ -1,0 +1,123 @@
+<?php
+
+namespace Tests;
+
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+use Illuminate\Support\Str;
+use OGame\Actions\Fortify\CreateNewUser;
+use OGame\Factories\PlanetServiceFactory;
+use OGame\Factories\PlayerServiceFactory;
+use OGame\Models\User;
+use OGame\Services\SettingsService;
+
+/**
+ * Isolated variant of AccountTestCase.
+ *
+ * Differences from AccountTestCase:
+ *  - Wraps each test in a database transaction (DatabaseTransactions) so no rows leak
+ *    between tests and tests no longer depend on execution order.
+ *  - Creates the user via the production CreateNewUser action + actingAs(), instead of a
+ *    real HTTP /register + /login round-trip.
+ *  - Starts the session up-front so csrf_token() returns a real token, avoiding a
+ *    session-initializing GET request; CSRF stays fully enforced.
+ *  - Neutralizes the "first user becomes admin" side effect (see createUser()).
+ *  - Resets the stateful singleton services (SettingsService + service factories) on
+ *    teardown, because their in-memory caches are NOT rolled back by DatabaseTransactions.
+ *
+ * Test intent (assertions) is identical to AccountTestCase-based tests; only the setup
+ * mechanism changes. See docs/test-isolation-migration-plan.md (Phase 1).
+ */
+abstract class IsolatedAccountTestCase extends AccountTestCase
+{
+    use DatabaseTransactions;
+
+    /**
+     * @return void
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        // Start the session so csrf_token() returns a real token. Controllers verify the
+        // token themselves (e.g. AbstractUnitsController::addBuildRequest compares
+        // session()->token() to the posted _token), so this avoids needing a
+        // session-initializing GET request while keeping CSRF fully enforced.
+        $this->app['session']->driver()->start();
+    }
+
+    /**
+     * Reset stateful singletons between tests. Their in-memory caches (settings values,
+     * cached PlanetService/PlayerService instances) survive across tests because the app
+     * container is not rebuilt, so we clear them explicitly here.
+     *
+     * @return void
+     */
+    protected function tearDown(): void
+    {
+        $this->app->forgetInstance(SettingsService::class);
+        $this->app->forgetInstance(PlayerServiceFactory::class);
+        $this->app->forgetInstance(PlanetServiceFactory::class);
+
+        parent::tearDown();
+    }
+
+    /**
+     * Create a user and authenticate without HTTP round-trips.
+     *
+     * Reuses the production CreateNewUser action (creates User + UserTech + initial planets
+     * + welcome message), then authenticates via actingAs() and wires up the planet services
+     * from the real factories — mirroring retrieveMetaFields() without the HTML parsing.
+     *
+     * @return void
+     */
+    protected function createAndLoginUser(): void
+    {
+        $user = $this->createUser();
+
+        $this->actingAs($user);
+
+        $this->currentUserId = $user->id;
+        $this->currentUsername = $user->username;
+
+        // Wire up planet services from the real factory.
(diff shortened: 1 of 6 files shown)


nagi1:
Adding to #1577

## Test isolation migration: plan & direction

Starting an incremental migration to make the test suite isolated and faster, one small PR at a time.

Full plan: `test-isolation-migration-plan.md`.

**Problem:** the suite currently has no DB isolation. Tests share state and can depend on execution order, which is the root cause of the flakiness in #1579 and #1580.

**Direction:**

1. **Phase 0: foundations**
   - Test-safe locks, cache, and settings.
   - #1584 + `CACHE_STORE=array`.

2. **Phase 1: isolated base**
   - Introduce `IsolatedAccountTestCase`.
   - Wrap each test in a transaction.
   - Replace the HTTP registration setup with direct user creation.
   - Convert tests one file at a time while keeping assertions unchanged.

3. **Later**
   - Remove `reloadApplication()`.
   - Remove HTTP registration from test setup completely.
   - Reduce heavy HTTP render loops inside test bodies where possible.

**Guardrails:**

- No mocks.
- No assertion changes.
- No duplicated coverage.
- Keep every PR small and green on its own.

This PR is the Phase 1 pilot and converts:

- `PlanetFieldRestrictionTest`
- `UnitQueueTest`

Feedback is welcome before converting the rest of the suite.

[test-isolation-migration-plan.md](https://github.com/user-attachments/files/31354879/test-isolation-migration-plan.md)

nagi1:
## About the `VacationModeTest` CI failure

The CI failure in `VacationModeTest > research queue pauses during vacation mode` looked related to this PR, so I traced it down.

It is not a regression from this PR. The isolation change exposed an existing settings leak between tests.

### Root cause

The sequence is:

1. `ResearchQueueTest` sets `research_speed=2` and leaves it in the DB because the test is not transactional.
2. Before this PR, `UnitQueueTest` later set `research_speed=1` and also left it in the DB.
3. That accidentally masked the leak, so `VacationModeTest` received `research_speed=1` and passed.
4. After migrating `UnitQueueTest` to `DatabaseTransactions`, its `research_speed=1` is rolled back.
5. The leaked `research_speed=2` from `ResearchQueueTest` now reaches `VacationModeTest`, causing the research to finish during the first 100 seconds and fail the assertion.

I also confirmed this is unrelated to `CACHE_STORE=array`.

Removing that change still reproduces the failure, while running against a fresh migrated DB with no leaked settings makes the test pass.

### Fix

Two small fixes address the actual problem:

- `AccountTestCase.php`
  - Add `research_speed=1` to the default settings baseline alongside `economy_speed=8`.
  - This prevents tests from inheriting a leaked research speed from earlier tests.

- `ResearchQueueTest.php`
  - `testResearchLabRequirement` now explicitly sets `research_speed=2`.
  - The test was previously relying on leaked state to finish the research within 2 minutes.

### Verified

- `VacationModeTest` passes even when `research_speed=2` is deliberately left in the DB beforehand.
- `ResearchQueueTest` + `VacationModeTest` + `UnitQueueTest` + `PlanetFieldRestrictionTest`: **47 passed**.
- Pint passes.
- PHPStan passes on all changed files.

This is exactly the kind of cross-test settings pollution tracked in #1021. The isolation work did not introduce the bug, it exposed it.

nagi1:
@jackbayliss This PR is ready for review

nagi1:
Kind reminder

nagi1:
@jackbayliss @Geda173 This PR is now ready for merge

tests/IsolatedAccountTestCase.php
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * Reuses the production CreateNewUser action (creates User + UserTech + initial planets
+     * + welcome message), then authenticates via actingAs() and wires up the planet services
+     * from the real factories — mirroring retrieveMetaFields() without the HTML parsing.
+     *
+     * @return void
+     */
+    protected function createAndLoginUser(): void
+    {
+        $user = $this->createUser();
```
jackbayliss, line 75: We could use a User factory here no? Seems odd to need a custom method?

nagi1: just wanted to make things as close to original code as possible and refactor from stable base for total isolation later on .. I still have 14 PRs to go.

jackbayliss (approved):
I'll approve for now as tests seem to pass, I would prefer using proper factories etc though.

Issue #1581, opened 2026-08-20
[BUG] PlanetServiceFactory::make with reloadCache does not reload the planet's player, causing battles to use stale user data


Geda173:
**Describe the bug**

`PlanetServiceFactory::make($planetId, reloadCache: true) `reloads the planet from the database, but the `PlanetService` constructor fetches the owning player via `PlayerServiceFactory::make($userId) `without the reload flag. If a PlayerService for that user is already cached in the factory, the returned planet carries a stale `User` model, defeating the documented purpose of the `reloadCache` flag.

Battle processing reads gameplay-relevant user preferences through this path. `AttackMission::processArrival() `loads the defender planet with `reloadCache: true` and then evaluates tactical retreat via `TacticalRetreatService::resolveRetreatThreshold()`, which reads `tactical_retreat_ratio` from the (potentially stale) user model.

**How this was discovered**

While investigating the flaky `ReaperDebrisCollectionTest` failures (#1579, fixed in #1580), the reproduced failure showed a battle report with d`efender_fled: true `while the users table simultaneously contained `tactical_retreat_ratio = 0.` The defender fled based on a `PlayerService` cached earlier in the same application instance  (before the test had written the ratio) even though the battle explicitly requested a reloaded planet. #1580 works around this at the test level with `reloadApplication()`; this issue covers the underlying factory behavior.

**Impact**

- Any code path that serves multiple operations from one application instance can battle with outdated user data. This is what made the tests in #1579 flaky: within a single test, requests made during fleet dispatch cached the defender's player before the test modified it, and battle processing in a later request of the same application instance reused the stale instance.
- The same within-instance staleness applies to a single queue job that processes multiple missions for one destination (#363): a mission processed early in the batch caches the defender's player, and a later battle in the same batch reuses it.
- Cross-job staleness in long-lived queue workers was also a concern, but #1413  now includes a per-job singleton reset (a `JobProcessing` listener that forgets the factory and settings singletons), so workers start each job with fresh state. That reset does not fix the within-request/within-job cases above. Those need the factory-level fix described here.

**Suggested fix**

In `PlanetServiceFactory::make()`, when `reloadCache` is true, resolve the owner with `PlayerServiceFactory::make($userId, true)` and construct the `PlanetService` with that fresh player, instead of letting the `PlanetService` constructor fall back to the cached instance. The same consideration applies to `makeForPlayer(..., useCache: false)`.

nagi1:
[...]

PR #1593, opened 2026-08-24
test: isolate 5 self-contained feature tests + factory-based account setup

nagi1:
## Description

Related: #1585

Continues the test suite isolation migration from `test-isolation-migration-plan.md`.

This is the Phase 2 batch. It converts 5 more self-contained feature tests to `IsolatedAccountTestCase` and updates the base account setup to use standard Eloquent factories based on review feedback from #1585.

### Part 1: Factory-based account setup

`Tests\IsolatedAccountTestCase::createUser()` no longer uses the `CreateNewUser` action.

It now creates a production-shaped account using:

- `User::factory()` for the user
- `UserTech::factory()` for the tech record
- `PlanetServiceFactory::createInitialPlanetForPlayer()` for the initial planets

Using the factory also removes the "first user becomes admin" side effect, so the previous demote workaround is no longer needed.

### Part 2: Convert 5 tests

The following tests now extend `IsolatedAccountTestCase` instead of `AccountTestCase`:

- `ResearchQueueCancelTest`
- `BuildQueueCancelTest`
- `FacilitiesHeaderTest`
- `MineEnergyAtLevelZeroTest`
- `TechtreeTest`

They now use factory-based account setup with `DatabaseTransactions` instead of real HTTP registration and a shared DB.

Assertions and coverage are unchanged.

The batch was verified:

- Each file passes independently.
- All 5 pass together.
- All 5 pass in reverse order.

### `MerchantTest` not converted

I also attempted to migrate `MerchantTest`, but reverted it.

`testTradeCappedWhenStorageNearlyFull` fails under `IsolatedAccountTestCase` while passing under `AccountTestCase`.

The test depends on fresh cross-request planet storage state and currently hits the stale planet cache issue from #1581.

It will stay on `AccountTestCase` until that issue is fixed, tracked as PR-01 in the migration plan.

### Type of Change

- [ ] Bug fix
- [x] Feature enhancement
- [ ] Documentation update
- [x] Other (test infrastructure)

## Related Issues

Progress toward #1021 for factories and proper DB isolation.

## Checklist

- [x] **Automated Refactoring:** Rector passes on changed files.
- [x] **Code Standards:** Laravel Pint passes on all 6 changed files.
- [x] **Static Analysis:** PHPStan passes on changed files.
- [x] **Testing:**
  - 5 converted files, 19 tests, pass independently and in reverse order.
  - `MerchantTest` migration reverted due to #1581.
- [x] **CSS & JS Build:** N/A, no frontend changes.
- [x] **Documentation:** Updated `test-isolation-migration-plan.md`.

## Additional Information

No production code changed.

Assertions in the 5 converted tests are untouched. Only the setup and isolation mechanism changed.

The `MerchantTest` result also gives us a clear migration boundary: tests that depend on fresh cross-request planet or service state need the #1581 cache fix before they can be safely isolated.


Geda173 (commented):
Thanks for PR.

Two things worth resolving before this becomes the foundation for the remaining PRs:

1. `createUser()` now duplicates `CreateNewUser::createInitialGameDataForUser()`, the planet-name array and the `registrationPlanetAmount()` loop, and it has already drifted: the factory path does not send the welcome message. That divergence appeared in the very first refactor of this base class, which is the argument for not hand-maintaining a second copy. Could that routine be extracted so registration and the test base call the same code? Otherwise every migrated test is asserting against an account shape that quietly diverges from a real signup, and with 14 PRs planned on top, that is expensive to unwind later.

2. `reloadApplication()` is incompatible with `DatabaseTransactions`. `refreshApplication()` opens a new connection outside the transaction, so the setUp user becomes invisible; I confirmed it with a scratch test, which fails on "Current user not found." at `AccountTestCase.php:115`. Counting direct callers plus everything extending `FleetDispatchTestCase`, that reaches 30 of 73 feature test files. This is inherited from #1585 rather than introduced here, but it caps the migration at roughly 60% of the suite until it is dealt with, so it would help to know the plan for those before more batches land.

Smaller points:

- The `createAndLoginUser()` docblock (lines 62-70) is now false. It still says "Reuses the production CreateNewUser action (creates User + UserTech + initial planets + welcome message)". You updated the class docblock and the `createUser()` docblock in the same commit, so this one just got missed.
- The docs checklist item is checked but there is no docs change in the diff.
- Optional nit: `User::factory()` uses a bare `fake()->name()`, where `CreateNewUser` had `generateUniqueName()` plus a retry loop, and `users.username` is UNIQUE. It cannot bite CI, which starts from an empty database, and it is around 1 in 90 runs against a long-lived local one. `User::factory()->create(['username' => 'test_' . Str::random(16)])` closes it if you are touching the method anyway.

nagi1:
Thanks, addressed the two foundation concerns.

**Shared account bootstrap**

`IsolatedAccountTestCase::createUser()` no longer maintains a separate copy of registration setup. I extracted the shared account-initialization routine into `InitialUserDataService`, which is now called by both `CreateNewUser` and the isolated test base.

The shared path creates:

- the `UserTech` record
- the settings-driven initial planets with the production `Homeworld` / `Colony` naming policy
- the welcome message

The isolated base still creates the user with `User::factory()` so test transactions do not trigger the real-signup “first registered user becomes admin” behavior. It now uses a collision-safe `test_` username as well. I also corrected the stale `createAndLoginUser()` docblock.

**`reloadApplication()` / transaction boundary**

I confirmed that `reloadApplication()` is incompatible with `DatabaseTransactions`: it rebuilds the application with a connection outside the test transaction, making fixtures created in `setUp()` unavailable.

The underlying stale-cache defect is tracked in [#1581](https://github.com/lanedirt/OGameX/issues/1581), and is now fixed here. `PlanetServiceFactory::make($planetId, reloadCache: true)` reloads the planet’s owning `PlayerService` and injects it into the reloaded `PlanetService`. Previously, the planet was refreshed but its owner could still come from the factory cache.

Added a regression test that:

1. primes the cached planet/player,
2. persists a changed `tactical_retreat_ratio`, then
3. verifies `make(..., true)` observes the new value through the reloaded owner.

`IsolatedAccountTestCase` still rejects `reloadApplication()` immediately, so future conversions cannot accidentally escape their transaction. The migration plan now links #1581 and records the conversion sequence: remove the reload callers after this cache fix, then migrate the blocked files individually.

Plan file for reference: [test-isolation-migration-plan.md](https://github.com/user-attachments/files/31410719/test-isolation-migration-plan.md)

Geda173 (changes requested):
Thanks for the changes. The `reloadApplication()` guard is a better answer than I asked for. The new regression test is real too, it fails if the factory hunk is reverted. Thanks for the sequencing answer on the blocked files.

1. **Blocker, and it quietly undoes the username fix from last round.** The stated reason for `User::factory()` is wrong. The first-user promotion lives in two places: `CreateNewUser.php:161-163`, and the `User` model's `created` hook at `app/Models/User.php:113-128`. The factory escapes the first, not the second, so deleting the demotion leaves nothing neutralizing it. Probing inside real isolated runs against a database migrated from empty:

       [PROBE test-1] username=Admin admin=YES
       [PROBE test-2] username=Admin admin=YES

It recurs on every isolated test, since `DatabaseTransactions` rolls the previous one back, and the hook overwrites the collision-safe `test_` username with `Admin`. Nothing fails yet, but the base now asserts against an admin rather than a normal player. Please restore the demotion, or assert non-admin so it cannot regress silently.

2. `make($id, reloadCache: true)` went from 1 query to 5 (measured 100 vs 20 over 20 calls, against the same code with only that hunk reverted), because it rebuilds the owner's entire `PlayerService` and `PlanetListService`. 28 call sites across 13 files, including two per iteration in `AllianceDepotController`'s fleet loop. Could you refresh just the user row instead? Related question: the reload now replaces the cached `PlanetService` for every sibling planet of the same owner, so `make(from, true); make(to, true)` leaves the caller holding a detached first object. I found no call site where that breaks, but the invariant changed.

Geda173 (approved):

nagi1:
Will stack the next PR-phase on-top of this one and rebase later after this PR gets merged.

nagi1:
This pr is approved and ready for merge.

lanedirt (approved):

PR #1594, opened 2026-08-26
fix: use correct singular translation keys for missile types in destroy rockets modal

### LINKED ISSUE #1286: [BUG] Raw translation keys in missile silo missile destruction menu
**Describe the bug**
/facilities -> missile silo -> destroy rockets:
Image

baigny:
anti_ballistic_missiles and interplanetary_missiles were plural but t_resources.php only defines the singular keys, so the modal rendered raw translation strings instead of the missile names.

Fixes #1286


Geda173:
Thanks!

PR #1599, opened 2026-08-29
test: isolate all feature tests and enable full parallel runs

nagi1:
## Description

Moves the last 11 `AccountTestCase` feature tests to `IsolatedAccountTestCase`. Every feature test now runs inside a database transaction with factory-based setup instead of HTTP registration.

This finishes the test isolation migration and makes the full suite safe to run with `php artisan test --parallel`.

Converted: `AdminTest`, `AdminAttackBlockTest`, `BanTest`, `AdminStuckFleetMissionTest`, `BuddyTest`, `FactoryTest`, `CsrfTest`, `Http200Test`, `MerchantTest`, `DestroyedPlanetTest`, `AllianceDepotSupplyRocketTest`.

`AllianceDepotSupplyRocketTest` also drops its manual `tearDown()`, which ended in a `TRUNCATE` that would have committed the open transaction.

### Type of Change:
- [x] Other (test isolation, no production behavior change)

## Related Issues

Relates to the test isolation migration plan (`docs/test-isolation-migration-plan.md`).

## Checklist

- [x] **Testing:**
- [ ] **Manual Testing:** No production code changed.
- [x] **Documentation:** Migration status log updated.

## Additional Information

Local runtimes on the dev machine, 4 workers for parallel:

- `php artisan test`: 1149 passed, 655s (10m 55s)
- `php artisan test --parallel`: 1107 passed, 42 notices, 0 failed, 456s (7m 36s)

The 42 notices are pre-existing tests that print debug output. They still pass.


tests/IsolatedAccountTestCase.php
```diff
@@ -106,23 +93,20 @@ protected function createAndLoginUser(): void
...
-        }
+        // Create without model events to skip the `created` hook that promotes the first
+        // user to admin. That hook issues a cross-row `update users`, which contends with
+        // the scheduler/queue worker and deadlocks (1205 lock wait) in tests.
+        $user = User::withoutEvents(fn (): User => User::factory()->create([
+            'username' => 'test_' . Str::random(16),
+        ]));
+
+        // Guard against a regression that silently promotes factory users to admin.
+        $this->assertFalse($user->hasRole('admin'), 'Factory user should not be an admin.');
```
jackbayliss: This feels like it should be in an actual test rather than the Testcase itself imo/

nagi1: Done

nagi1:
@jackbayliss @Geda173 untill you review this PR and merge it will work on phase 4 which will be to convert the rest of the hard files and after that we will only have 1 phase left which is to use --parallel flag and use pest sharding feature to split tests across 3-4 workers will make the tests pass in 1 minute instead of 5 minutes.

then we can use tia engine later on ... 

This is so exciting 🔥

.github/actions/docker-compose-setup/action.yml
```diff
@@ -24,6 +24,12 @@ runs:
       run: |
         timeout 60 bash -c 'until docker compose exec -T ogamex-app php artisan migrate:status > /dev/null 2>&1; do echo "Waiting for application to be ready..." && sleep 2; done'
 
+    - name: Provision isolated PHPUnit database
+      shell: bash
+      run: |
+        docker compose exec -T ogamex-db mariadb -uroot -ptoor -e "CREATE DATABASE IF NOT EXISTS \`ogamex-test\` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci"
```
jackbayliss, line 30: How come we need to do this? I'm sure Laravel does this automatically in the RefreshDatabase trait

nagi1: @jackbayliss 

`RefreshDatabase` doesn't create the database, it only runs migrations against one that already exists. Per the [Laravel docs](https://laravel.com/docs/11.x/database-testing), the trait wraps each test in a transaction and runs `migrate:fresh`, but it never issues a `CREATE DATABASE`. If `ogamex-test` didn't exist, `migrate:fresh` would fail with `SQLSTATE[HY000] [1049] Unknown database`.

Two things also make this step necessary rather than redundant:
1. These tests use `DatabaseTransactions` (`IsolatedAccountTestCase`), not `RefreshDatabase`  and `DatabaseTransactions` likewise assumes the schema already exists.

2. The step provisions a *dedicated* `ogamex-test` DB, separate from the main `laravel` DB (`MARIADB_DATABASE: laravel` in `docker-compose.yml`). No Laravel trait creates a second database for you  that's exactly what this manual `CREATE DATABASE` does.

jackbayliss: Hm, it feels odd to me that we run `artisan migrate`, is the idea we're gonna pull it out eventually and let laravel do it?

nagi1: Ya sure my end goal is full isolation that will allow us to use --parallel and sharding and eventually using tia engine as well to allow for faster development feedback loops. 

Let him cook 😉

jackbayliss (approved):
Seems ok if the end goal is to use pest and clean a lot of this stuff up etc

nagi1:
@jackbayliss BTW the new rector update doen by deps bot needs a PR to run on main branch because it caues this PR fail on unrelated files

Would you do it or should I do it?

jackbayliss:
@nagi1 you can merge main in now, sorted

nagi1:
Thanks @jackbayliss, I'm happy to say that we reached level where can run --parallel and I'm going to explore this in the next PR with multiple github actions workers as well.

so since this PR is approved please merge it as soon as you can.

Cheers 🥂

jackbayliss:
Seeing as this is all test related, I'm going to merge.. under the assumption a lot of clean up and improvements are going to happen so we can drop stuff like the migration calling in the docker image etc.

